PR #415, opened 2015-01-04
added translations

schnautzr:
added translations for en.twitter, en.facebook

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -957,6 +957,7 @@ en:
   extinct_in_the_wild: extinct in the wild
   extra_properties: Extra Properties
   extras: Extras
+  facebook:"iNaturalist.org on Facebook"
   facebook_account_information: Your Facebook Account Information
   facebook_account_not_linked: "%{user} doesn't have a linked facebook account anymore, so %{site_name} doesn't have permission to access this photo."
   facebook_access_token_expired: "Access token has expired. %{user} needs to re-link their Facebook account."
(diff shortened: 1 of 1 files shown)


kueda:
Hey @schnautzr, thanks for that, and thanks for using a pull request. Couple issues, though:

1. Are these translations being used anywhere? If not, is there a reason you're adding them? Maybe you forgot to add some of the changes to the files that use them?
1. We try to have translation keys correspond to their translations, so a better key for `Follow or tweet us!` would be `follow_or_tweet_us`.
1. There are numerous sites using our code, so we try to keep references to iNaturalist out of the translations. Things like `iNaturalist.org on Facebook` should be replaced with `%{site_name} on Facebook`

schnautzr:
Hi...I'm new to this pull request stuff. I am attempting to address the issue where if you hover over the links at the bottom of the site that say "Facebook" and "Twitter" and see tooltips that complain about there not being translations for en.facebook and en.twitter.

Rob

----- Original Message -----

kueda:
Got it. Ok, so those are used at https://github.com/inaturalist/inaturalist/blob/master/app/views/shared/_footer.html.haml#L8, but I think a better approach there would be to not translate them at all, since I don't think they use different names in different languages, right? Facebook is still "Facebook" no matter what language you speak? So to fix that you might replace `pieces << link_to(t(:facebook), site.facebook_url) unless site.facebook_url.blank?` with `pieces << link_to("facebook", site.facebook_url) unless site.facebook_url.blank?`. 

If I'm wrong and they *do* have different names in different languages, then you'd want to translate them as they are used in those languages, not as some other string like "Follow us on Facebook." So `facebook: Facebook` might go in `en.yml`, `facebook: フェイスブック` might go in `ja.yml`, etc.

kueda:
I'm closing this since you haven't responded in 6 months. If you still want to do something like this, I'd recommend updating your fork, b/c there have been many translation changes since you opened this PR.

PR #432, opened 2015-01-21
translations on map overlays

calonso-conabio:


kueda:
Thanks @calonso-conabio, we'll get these merged and deployed later today.

PR #465, opened 2015-02-03
Fixed issue with i18n arguments getting translated

pleary:
This is what has been causing the email updates task to fail


kueda:
I suspect the key that was messing with the updates mailer was `observations_from`. Can you add a spec to test for that?

pleary:
ok I added a spec for the failing mailer case in 62efd6a3283d9ede0c73707806e88ce10206d13c

PR #476, opened 2015-02-18
fixed typo in romanian key noticed by translatewiki volunteers

loarie:
its 'romanian' not 'rumanian' right?


kueda:
Can you also update any taxon names in the db that were using "Rumanian"?

loarie:
Will do - looks like only 3 - will zap them later today

loarie:
Done

PR #492, opened 2015-03-02
fixing broken URL in translation

calonso-conabio:


kueda:
Can you remove the span? We should try to avoid hard-code styling in the translation files. Also, we're about to merge in the rails4 branch, so you might want to make this change there.

PR #497, opened 2015-03-04
Rails 4 Upgrade

kueda:
Many, many changes to get us on Rails 4... only a year late! The most
significant changes are probably the finder methods and the different behavior
of `includes`, which seems to be even lazier about loading associates and
doesn't seem to trigger joins early enough for conditions to use them.


md5:
Put on your helmets everyone :construction_worker:

kueda:
Hah, helmets have been on since last week. Just been holding off on a merge so we could use the old code as a reference, but most things seem to be working.

md5:
:fireworks:

PR #507, opened 2015-03-20
Performance improvements with preloading and indices

pleary:


config/initializers/active_record.rb
```diff
@@ -40,36 +40,28 @@ def to_json(options = {})
...
+      if year || month || day
         begin
-          Date.parse(datestring)
-          ["#{column}::DATE = ?", datestring]
+          if day
+            date = Date.parse("#{ year }-#{ month }-#{ day }")
+            [ "#{column} >= ? AND #{column} < ?", date, date + 1.day ]
+          elsif month
+            date = Date.parse("#{ year }-#{ month }-1")
+            [ "#{column} >= ? AND #{column} < ?", date, date + 1.month ]
```
kueda: Does this replicates the former functionality, where requesting `/observations?month=6` would show observations observed in June of *any* year?

pleary: No. How was that working before? Looks like it takes a date string and splits it into year-month-day. What do you send to get it to search on month? Something like "-3-" to search for March? Can they do the same on day (e.g. "--5" to get things records on the 5th of the month)? Presumably we are using and need to keep that functionality? Can you think of a place we take advantage of this feature that I can test with?

kueda: Yup, that's how it works. We do not really use that internally to my knowledge, but it is something we've published as a part of our [API](http://www.inaturalist.org/pages/api+reference#get-observations), and I don't want to break that contract. http://www.inaturalist.org/observations?on=-6- will work, but http://www.inaturalist.org/observations?month=6 basically sends the same info down to `conditions_for_date`

PR #533, opened 2015-04-23
Added the Updates ES index logic

pleary:
This has a migration that will delete all updates older than 2014-11-01, which can take a few minutes. There may be some query locking that happens during the delete, so we should release this at a slow time, or put up the maintenance page for a few minutes while it runs. The Updates ES indexing will need to be kicked off manually after the deploy


kueda:
This all looks good to me. Would you merge and deploy it when you're around to keep an eye on it?

PR #539, opened 2015-04-29
Removed Brisbane "Daylight Savings"

scottwarren:
This is because Brisbane doesn't have daylight savings. This might be better changed to somewhere that does (for example, Sydney or Melbourne).

### DIFF
--- a/config/initializers/time_zone_codes.rb
+++ b/config/initializers/time_zone_codes.rb
@@ -6,6 +6,7 @@ class ActiveSupport::TimeZone
     'AWST' => ActiveSupport::TimeZone["Perth"],
     'ACST' => ActiveSupport::TimeZone["Adelaide"],
     'AEST' => ActiveSupport::TimeZone["Brisbane"],
+    'AEDT' => ActiveSupport::Timezone["Sydney"],
     'AWDT' => ActiveSupport::TimeZone["Perth"],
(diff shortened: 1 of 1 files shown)


kueda:
Those mappings are intended to link time zone abbreviations like AEDT to time zones supported by the tzinfo gem, so I think mapping AEDT to one of the zones listed in https://github.com/tzinfo/tzinfo/blob/master/test/zoneinfo/zone.tab would be more appropriate than deleting it. Could you take a look at that, and if Australia/Sydney looks like the best choice, update your feature branch to use that instead of just deleting this mapping?

scottwarren:
I've added Sydney as the AEDT timezone - let me know if this is all good.

config/initializers/time_zone_codes.rb
```diff
@@ -6,9 +6,9 @@ class ActiveSupport::TimeZone
     'AWST' => ActiveSupport::TimeZone["Perth"],
     'ACST' => ActiveSupport::TimeZone["Adelaide"],
     'AEST' => ActiveSupport::TimeZone["Brisbane"],
+    'AEDT' => ActiveSupport::Timezone["Sydney"],
```
kueda: This is needs to be capitalized correctly: `TimeZone`, not `Timezone`.

scottwarren: fixed - sorry about that

PR #547, opened 2015-05-04
Can set session locale; Custom homepage wikis per locale

pleary:


spec/controllers/welcome_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+describe WelcomeController do
+
+  describe "set_homepage_wiki" do
+    before(:all) do
+      @home = WikiPage.make!(path: "home")
+      @homeES = WikiPage.make!(path: "eshome")
+      @homeFR = WikiPage.make!(path: "frhome")
+    end
+
+    it "does't set @page if there is no home_page_wiki_path" do
```
md5: Typo here. Should be "doesn't set @page"

PR #564, opened 2015-05-21
more translations

calonso-conabio:


kueda:
These should be released now.

PR #567, opened 2015-05-25
Using elasticsearch to generate all but ProjectList listed_taxa.cache…

pleary:
…_columns


pleary:
I updated this pull request to remove the ProjectList exception for cache_columns. ProjectLists will no longer do anything for cache_columns

PR #587, opened 2015-06-08
French translations [WIP]

jbourassa:
Adds french translations as provided by the CWF and improves some other areas of the translation:

- Introduce `I18n.has_t?` to bring new translation constructs without breaking other translations (probably a necessary evil to improve some translations without introducing breaking changes)
- Add `i18n-inflections` to help genderize some i18n constructs in french
- Add `devise-i18n` to translate Devise messages
- Improve french copy
- Wrap some plain english strings in `I18n.t`

This is still a huge WIP but having this deployed on staging (or wherever a proper database live) would really help me identify areas that need the most work under normal use.

/cc @loarie


kueda:
Can you issue this pull request against https://github.com/inaturalist/inaturalist/tree/i18n? We'd be happy to get these changes on a staging machine, but it's easier for us to do so if they're not in master.

Regarding `has_t?`, were the translation fallbacks not working? In theory if a translation doesn't exist in the current locale, it should fallback to English, which means you might get some English showing up for other locales, but I've always figured that was better then having no content, or poorly-translated content. That said, use of `has_t?` would be vastly superior to 500 errors.

jbourassa:
>Can you issue this pull request against https://github.com/inaturalist/inaturalist/tree/i18n?

Sure!

>Regarding has_t?, were the translation fallbacks not working? In theory if a translation doesn't exist in the current locale, it should fallback to English, which means you might get some English showing up for other locales

The fallback is indeed working and falls back to english. I'm assuming that the previous translations were working decently well (especially spanish) — in which case replacing the current translation key for a new one (for instance, [this change](https://github.com/inaturalist/inaturalist/pull/587/files#diff-92444b9cfa8fd1c47fee508e5d1c08a6R65)) would introduce an english fallback where other languages had a proper translation before.

If retranslating the new key is not a big deal for iNaturalist, I can issue a new PR without `has_t?`. Just let me know.

PR #589, opened 2015-06-08
French translations [WIP] 

jbourassa:
(Clone of #587 for i18n branch)

Adds french translations as provided by the CWF and improves some other areas of the translation:

- ~~Introduce `I18n.has_t?` to bring new translation constructs without breaking other translations (probably a necessary evil to improve some translations without introducing breaking changes)~~
- Add `i18n-inflections` to help genderize some i18n constructs in french
- Add `devise-i18n` to translate Devise messages
- Improve french copy
- Wrap some plain english strings in `I18n.t`

This is still a huge WIP but having this deployed on staging (or wherever a proper database live) would really help me identify areas that need the most work under normal use.


kueda:
@loarie, do you think it would be better to have English show up in non-English locales for keys that are not translated in those locales, or to show nothing or a potentially poor translation in those cases? IMO, it's better to have English, b/c a) our templates will be simpler and easier to maintain, and b) it incentivizes translation teams to translate keys.

If we can make a decision on that @jbourassa can clean some of this up and we can get this onto a staging machine for everyone to test.

loarie:
@kueda I agree with you

kueda:
Cool. @jbourassa, please remove those `has_t?` calls then and well get this running somewhere.

jbourassa:
Pushed a new version without `has_t?`.

I had to handle the `something.match(/^[aeiou]/i) ? t(:an) : t(:a)`, which took more time than I would have liked. Generally speaking, I'd recommend avoiding such constructs for I18n'd projects. I found a way to keep these in for now, but it's not very pretty.

------

Please hold before deploying, I would like to finish one more round of changes. I'll finish tomorrow morning — since I'm in EST I'm guessing I should before regular business hours in PST.

Thanks!

jbourassa:
Ready to go on staging whenever you get the chance. Thanks!

kueda:
@jbourassa, you can now check this out at http://gorilla.inaturalist.org/set_locale?locale=fr. Username: preview, password: 313phant

PR #590, opened 2015-06-09
Improve french translations - pt. 2

jbourassa:
Hi @kueda,

Thanks for the previous merge. Here's more commits that would be great to merge before the official launch.

Thanks!

### DIFF
--- a/app/helpers/taxa_helper.rb
+++ b/app/helpers/taxa_helper.rb
@@ -210,7 +210,7 @@ module TaxaHelper
     if controller.fragment_exist?(key)
       return content_tag(:div, controller.read_fragment(key), :id => element_id)
     end
-    html = content_tag(:div, "Loading images...", :id => element_id, :class => "loading status")
+    html = content_tag(:div, t(:loading_images), :id => element_id, :class => "loading status")
     js = <<-JS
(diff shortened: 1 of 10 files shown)


kueda:
Alrighty, merged and up at http://gorilla.inaturalist.org/set_locale?locale=fr, same username/pw. CC @loarie

loarie:
Looking good! I took a quick look a popular pages gorilla looking just at formatting (I don't speak French) - in my opinion, the biggest bang for one's buck fixes due to prominence on the site would be the following 3:

1) Overflow on menu, is there a shorter choice of word for Dashboard and Sign Out?
screen shot 2015-06-09 at 4 22 47 pm

2) Overflow on observations component in observations/index, is there a shorter choice of word for Observer? Maybe Person or user?
screen shot 2015-06-09 at 4 22 03 pm

3) I don't think 'Member of the iNaturalist Network' is internationalized yet but this would be good to do and translate as its in the footer on every page
screen shot 2015-06-09 at 4 21 59 pm

jbourassa:
Good catch. I was aware of 2) but somehow missed 1) and 3). I gotta sign off for ~1.5 hours but I'll get to it once I get back, hopefully it won't be too late.

Thanks!

PR #592, opened 2015-06-10
Improve french translations - pt. 3

jbourassa:
Hi @kueda,

Some more translations that can hopefully make it before the deployment. Special thanks to @loarie for bringing up a few of them!

Thanks!


jbourassa:
Argh, proposed against master — sorry.

kueda:
I think this is in here: you've translated `research_grade_short_html: "R<span class='full'>esearch </span>G<span class='full'>rade</span>"` as `research_grade_short_html: R<span class='full'>calibre </span> N<span class='full'>recherche </span>`. Maybe translation files aren't the best way to do this, but the intent was to have text that could be shorted to an acronym with CSS, so "Research Grade" becomes "RG" when spans of class "full" are hidden. Your French translation will show "RN" appropriately in that case, but when expanded it shows "Rcalibre Nrecherche," which doesn't seem right. Was that intentional, and if not can you recommend a better translation?

jbourassa:
It was _not_ intentional — it was from the translator and I missed it.

Here's a better translation: `C<span class='full'>alibre </span> R<span class='full'>echerche</span>`

kueda:
Thanks, I'll fix it then.

PR #594, opened 2015-06-10
Improve french translations - pt. 4

jbourassa:
One more! Should be good to go for the launch.

Thanks!

### DIFF
--- a/app/views/lists/by_login.html.erb
+++ b/app/views/lists/by_login.html.erb
@@ -1,10 +1,14 @@
+<%
+  observation_by_login_title = if is_me?(@selected_user)
+    t('lists_by_login.title.current_user',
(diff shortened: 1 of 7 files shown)


app/views/lists/by_login.html.erb
```diff
@@ -60,7 +64,7 @@
     </div>
   <% end %>
 <% end %>
-<%= render :partial => "shared/by_login_header", :locals => { :after => tools } %>
+<%= render :partial => "shared/by_login_header", :locals => { :after => tools, title: observation_by_login_title } %>
```
jbourassa: James asked for a custom title in one specific page, so I introduced a new translation string along with a default. Not incredibly pretty, but I think it's not too bad.

PR #652, opened 2015-08-04
added FB js to facebook canvas endpoint

loarie:
I think this is what is preventing our FB app from being approved


kueda:
We generally don't hardcode instance-specific attributes like the Facebook app ID. Try using `<%= CONFIG.facebook.app_id %>` instead. We also have `@site.facebook_app_id` but we haven't really converted to using the `Site` records for this, so might as well stay consistent

loarie:
sorry about that - how's this

PR #695, opened 2015-09-10
ja-phonetic.yml

zabas:
Created a reference file for sorting phonetically in Japanese. (Currently only lists some lexicons).


kueda:
Should already have been integrated in #705

PR #701, opened 2015-09-17
Bootstrap3

kueda:
Replaces old bootstrap2 stuff with bootstrap 3, updates jquery and jquery-ui.


app/assets/javascripts/jquery/plugins/inat/observation_selectors.js
```diff
@@ -69,11 +69,11 @@
       return false;
     });
     var selectAll = $('<a href="#" class="selectall button">' + I18n.t('select_all') + '</a>').click(function(e) {
-      $(wrapper).find('input:checkbox').attr('checked', true);
+      $(wrapper).find('input:checkbox'). prop('checked', true);
```
pleary: only questionable thing I've seen so far. Is the extra space OK here and below?

app/models/guide_range.rb
```diff
@@ -23,6 +23,9 @@ class GuideRange < ActiveRecord::Base
   validates_attachment_content_type :file, :content_type => [/jpe?g/i, /png/i, /gif/i, /octet-stream/], 
     :message => "must be JPG, PNG, or GIF"
 
+  validates_attachment_content_type :file, :content_type => [/jpe?g/i, /png/i, /gif/i, /octet-stream/],
```
pleary: duplicate validation

app/assets/javascripts/jquery/plugins/inat/observation_selectors.js
kueda: I think it parses, but not good form. Will fix.

PR #709, opened 2015-09-26
BugGuide name provider

kueda:
Adds BugGuide as a potential name provider. @pleary, I made a slight change to the way `ApiEndpointCache` works (doesn't cache empty or failed responses). I think it's fine, but I wanted you to take a look before I merged it in.


pleary:
this looks fine to me. ApiEndpointCache is really only useful when it successful caches stuff, and since you added the success? check, it should work OK

pleary:
on second thought, maybe we could do something like:
```
if api_endpoint_cache
  api_endpoint_cache.update_attributes(
    request_completed_at: Time.now,
    success: !response.body.blank?,
    response: response.body
  )
end
```

which will set the request_completed_at, so we don't think there's an endless request going on. That will match what we're doing around https://github.com/inaturalist/inaturalist/blob/master/lib/meta_service.rb#L96

PR #725, opened 2015-10-06
New homepage

kueda:
Redesigned the homepage to look better and more clearly communicate what iNat is about.


loarie:
@siebrand would it be possible to do a pull from inaturalist to translatewiki? We added a new homepage and have gotten some requests from folks who want to get to work translating it. Thanks!

PR #734, opened 2015-10-08
Added min_id obs API param, ability to page by min_id and limit.

kueda:
@pleary, can you take a look at this? I had to change some parts of elastic_query and related methods. Specs seem to pass, but wanted to run this by you all the same.

### DIFF
--- a/app/es_indices/observation_index.rb
+++ b/app/es_indices/observation_index.rb
@@ -377,19 +377,13 @@ class Observation < ActiveRecord::Base
 
     if p[:min_id]
       search_filters << {
-        range: { id: {gte: p[:min_id] } }
+        range: { id: { gte: p[:min_id] } }
       }
     end
-
-    if p[:limit]
-      search_filters << {
-        limit: { value: p[:limit] }
-      }
-    end
-
+    
     { where: search_wheres,
       filters: search_filters,
-      per_page: p[:per_page].blank? && p[:limit].blank? ? 30 : (p[:per_page] || 30),
+      per_page: p[:per_page] || 30,
       page: p[:page],
       sort: sort,
       extra_preloads: extra_preloads }
--- a/app/es_indices/observation_index.rb
+++ b/app/es_indices/observation_index.rb
@@ -375,9 +375,21 @@ class Observation < ActiveRecord::Base
       end
     end
 
+    if p[:min_id]
+      search_filters << {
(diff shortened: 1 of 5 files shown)


lib/observation_search.rb
```diff
@@ -29,20 +29,14 @@ def site_search_params(site, params = {})
 
     def search_in_batches(raw_params, options={}, &block)
       search_params = Observation.get_search_params(raw_params, options)
-      search_params.merge!({ page: 1, per_page: 500, preload: [ ],
+      search_params.merge!({ min_id: 1, limit: 500, preload: [ ],
         order_by: "id", order: "asc" })
-      scope = Observation.get_search_scope_or_elastic_results(search_params)
-      if scope.is_a?(ActiveRecord::Relation)
```
pleary: I should have added some comments here. Some queries cannot be performed with ES and we need to fall back to querying the DB. `get_search_scope_or_elastic_results` will return either an AR:Relation (DB query) or a WP:Collection (ES query). I'm thinking we might need this "if scope is AR:Relation" conditional back, and then add the min_id/limit override in the WP:Collection bit of it

app/es_indices/observation_index.rb
```diff
@@ -375,9 +375,21 @@ def self.params_to_elastic_query(params, options = {})
...
+    if p[:limit]
+      search_filters << {
+        limit: { value: p[:limit] }
+      }
+    end
+
     { where: search_wheres,
       filters: search_filters,
-      per_page: p[:per_page] || 30,
+      per_page: p[:per_page].blank? && p[:limit].blank? ? 30 : (p[:per_page] || 30),
```
pleary: doesn't this work out to be equivalent to `p[:per_page] || 30`?

lib/observation_search.rb
pleary: Also, the AR:Relation find_in_batches will already to proper min_id/limit kinds of queries

kueda: Since I pushed the `min_id` and `limit` handling down into both the db-based and es-based query implementations, `Observation.get_search_scope_or_elastic_results(search_params)` should handle deciding which kind of query to use and both implementations should use `min_id` and `limit` correctly. Both return objects that can be used to get the ID of the last result, so it seemed simpler to remove the conditional.

pleary: yep good point. I retract my comment

PR #822, opened 2016-01-18
Precompile

pleary:
faster asset precompiling on deploy - tmp/cache isn't getting flushed


kueda:
So the main thing here was switching to a different version of angular-rails4-templates? Or changing the order in which we include all our angular deps?

pleary:
There were two separate issues: assets taking a long time to precompile and not being able to deploy to hechtia. I was mainly focused on the former. Turns out angular-rails-templates clears the tmp files, which are needed for fast asset precompiling. Moving from that to angular-rails**4** seemed to fix that.

Unrelated, I split the bundles so we don't have to recompile the bower stuff any time we change our app logic. The benefit of that remains to be seen.

For the deploy to hechtia, the main bug was different, and fixed in 35c733b . That server was configured to load Spanish by default, which created an i18n key that did not exist. It expected the English key. After that was fixed, I needed to remove multiple public/assets/manifest-*.yml* files to get assets to precompile again. Right now I am able deploy, and asset precompiling is down from 220 seconds to 20 seconds or less.

kueda:
Oh jeez, the hechtia issue was mine. Forgot that would be precompiled in different languages. Apologies. Will try to be on the lookout for that in the future.

PR #852, opened 2016-01-29
Suspend subscriptions of inactive users

pleary:


pleary:
Everything is in place, but the actual suspensions and subsequent emailing is commented out for now until we get translations

PR #855, opened 2016-01-30
Pagination in obs search

pleary:


kueda:
Maybe I'm not deploying this right, or assets are not getting rebuilt or something, but when I deploy this to staging I don't get pagination.

pleary:
The problem was that the master branch had newer assets than this branch, so these assets were not getting precompiled. I merged master into search_pagination and I'm now seeing pagination on Staging with `BRANCH_NAME=search_pagination cap staging deploy`

PR #856, opened 2016-01-30
change hash # to search ? to allow for crawling

jhpoelen:
I used https://developer.mozilla.org/en-US/docs/Web/API/History/pushState to allow for search ? in url instead of hash / anchor. Apparently, search engines crawl search ? but not hashes.

Thanks!


kueda:
Thanks, Jorrit. Will release soon.

PR #888, opened 2016-02-17
semantics change for still_needs_id

loarie:
trivial change - mostly wanted to test making commits from this new VB setup


kueda:
Is this ready to go? Really just a change in text?

loarie:
yep -

Issue #851, opened 2016-01-29
change semantics of 'Still Needs ID?'


loarie:
People are misunderstanding the 'Still Needs ID?' vote in the Data Quality Assessment to mean
'Do I think my own ID on this obs is incorrect?'

Here are some examples:
http://www.inaturalist.org/observations?place_id=any&quality_grade=casual

Suggest switching the label from "Still Needs ID?" to "Community can confirm/improve ID?"
<img width="292" alt="screen shot 2016-01-29 at 1 01 39 pm" src="https://cloud.githubusercontent.com/assets/598026/12688052/780284a2-c688-11e5-9f9b-9c6d99f121b4.png">

kueda:
Fine with me.

loarie:
I can probably take this one on

pleary:
Scott resolved this with b2e8e2d

loarie:
Still some confusion here https://groups.google.com/forum/#!topic/inaturalist/0ZDP9YKEaO8
In this pull request: https://github.com/inaturalist/inaturalist/pull/910 I made these additional changes:
<img width="926" alt="screen shot 2016-03-02 at 12 06 39 pm" src="https://cloud.githubusercontent.com/assets/598026/13473696/54b60ac2-e06f-11e5-8b63-818b1d41d667.png">

naturalistcharlie:
Wouldn't you want it visible if there's a community ID of species or lower?
I thought one of the main points of this was to get a 'third opinion' if
you barely had research grade but wanted more feedback. I may be
misunderstanding but I thought that was the main point of having this
field. If it isn't, you could probably solve the problem by rewording it to
say 'observation can not be identified from photo' because that would be
the only other use for it, right?

Apologies if I'm not understanding.

On Wed, Mar 2, 2016 at 3:08 PM, Scott Loarie <notifications@github.com>
wrote:

> Still some confusion here
> https://groups.google.com/forum/#!topic/inaturalist/0ZDP9YKEaO8
> In this pull request: #910
> <https://github.com/inaturalist/inaturalist/pull/910> I made these
> additional changes:
> [image: screen shot 2016-03-02 at 12 06 39 pm]
> <https://cloud.githubusercontent.com/assets/598026/13473696/54b60ac2-e06f-11e5-8b63-818b1d41d667.png>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/inaturalist/inaturalist/issues/851#issuecomment-191404832>
> .
>



[...]

PR #910, opened 2016-03-02
Made former 'Still needs ID?' conditionally display and deduped Appropriate UI

loarie:
In another attempt to reduce the confusion causing people to vote yes on 'Community can improve/confirm ID?' AKA 'Still needs ID', I moved it up with the rest of the community ID pieces and have it not display if the obs is not a research grade candidate or the obs already has a community ID of species or lower. If the obs is needs ID but there are no votes yet, 'Yes' is hidden (to prevent people proactively voting yes)

I also consolidated the 'appropriate?' logic which appeared below licensing and was mostly repeated again in the data quality assessment into the data quality assessment

### DIFF
--- a/app/views/observations/show.html.erb
+++ b/app/views/observations/show.html.erb
@@ -802,24 +802,6 @@
       </div>
     <% end -%>
   </div>
-  
-  <% if @observation.flagged? %>
-    <div id="flaggings_heads_up" class="description">
-      <%=t :heads_up_this_observation_has_been_flagged %>
-      <%= link_to t(:view_flags), observation_flags_path(@observation) %>
-      |
-      <%= link_to t(:add_another_flag), new_observation_flag_path(@observation),
-                  :id => "flag_this", :rel => "nofollow", :class => "flaglink" %>
-    </div>
-  <% else %>
-    <p class="description ui">
-      <%= t(:is_this_observation_offensive_html) %>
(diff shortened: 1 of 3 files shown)


loarie:
ok this is ready to go unless any one wants a semantics change to the labeling (see https://github.com/inaturalist/inaturalist/issues/851#issuecomment-190950625)

kueda:
Can you capitalize and not link the "inappropriate?" text? By linking just one word you make it seem like an action. If you want to link to the help, maybe do so in the flag popup. E.g. change "Misleading or illegal content, racial or ethnic slurs, etc." to "Misleading or illegal content, racial or ethnic slurs, etc. For more on our definition of "appropriate," see the [FAQ](http://www.inaturalist.org/pages/help#inappropriate)."

loarie:
I can remove the link but since the flag popup is reused I don't think I'll be able to add the link to it only in the Obs context

kueda:
http://www.inaturalist.org/pages/help#inappropriate is not specific to observations. It applies to all content on the site.

loarie:
by it I mean the flag popup, read:
"I can remove the link but since the flag popup is reused I don't think I'll be able to add the link to the flag popup only in the Obs context"

loarie:
sorry about that - total brainmelt

Issue #996, opened 2016-06-07
Observation in same project two times


naturalistcharlie:
this observation apparently is double-added to Vermont Atlas of Life:

http://www.inaturalist.org/observations/3396952

![image](https://cloud.githubusercontent.com/assets/7171475/15860273/52d3908a-2c96-11e6-847f-103c151bd6f5.png)

kueda:
Do you remember how you added this observation to that project? E.g. did you add it manually, or was it added via the aggregator?

kueda:
Well, I removed the duplicates, but I still have no clue how they got there. If you can come up with some conditions under which this seems to be happening, please re-open this issue and let us know.

PR #1008, opened 2016-06-15
Trying to add obs from the aggregator in a transaction (#996)

kueda:
@pleary, do you think this would kill performance for the aggregator? I'm trying to prevent it from creating the same project observation twice. Not entirely sure why it would be doing that, but maybe it's running on multiple machines at the same time?


pleary:
There shouldn't be much overhead from adding transactions here. Seems good to me

PR #1011, opened 2016-06-20
ListedTaxon.observed_in_place? querying ES instead of DB

pleary:
@kueda You should probably OK this one in case there's something I'm missing. ES won't have 0 latency, but I think that's alright here. This is similar to what I just did in 64a66e6 which is making a small but positive impact

### DIFF
--- a/app/controllers/observations_controller.rb
+++ b/app/controllers/observations_controller.rb
@@ -139,7 +139,7 @@ class ObservationsController < ApplicationController
       end
       
       format.atom do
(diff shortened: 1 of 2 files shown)


kueda:
How much latency are we talking about? When you change a place geometry, it does queue a job to refresh the place check list, which iterates over every `ListedTaxon` in the place and runs this method, so if that happens *before* the ES index has the new geometry (or `ObservationPlace`s), I'm guessing this would result in things being on checklists when they shouldn't be.

PR #1026, opened 2016-06-27
cleaned up dashboad

loarie:
multiple major changes to the dashboard views to support a bootstrapified redesign
& minor changes to various controllers/models mainly to:
1) support getting lat/lon from ip in order to localize getting started material
2) move some parts of the dashboard into partials to alter how the dashboard is cached
3) preferences for whether or not to show onboarding material


app/helpers/application_helper.rb
```diff
@@ -276,16 +276,16 @@ def modal_image(photo, options = {})
     )
   end
   
-  def stripped_first_paragraph_of_text(text)
+  def stripped_first_paragraph_of_text(text,split)
```
kueda: Doesn't look like this broke anything, but it's generally preferable to make optional args like this optional, e.g. `stripped_first_paragraph_of_text( text, split = nil )`, then set a default value for `split` in the method.

app/controllers/users/registrations_controller.rb
```diff
@@ -16,6 +16,11 @@ def create
 
     if resource.save
       if resource.active_for_authentication?
+        last_ip = request.env['REMOTE_ADDR']
```
pleary: for logging I check in a different order which I found at the time to yield better results than checking REMOTE_ADDR first: `HTTP_X_FORWARDED_FOR, HTTP_X_CLUSTER_CLIENT_IP, REMOTE_ADDR`

pleary: I see this same logic block repeated in session_controller below, and I have something similar for logs in https://github.com/inaturalist/inaturalist/blob/master/config/initializers/logstasher.rb#L42 . Maybe we could extract this into a shared method. Not a blocker for merge - something for later

app/views/subscriptions/_form.html.haml
```diff
@@ -1,10 +1,10 @@
 = form_for @subscription, :builder => DefaultFormBuilder do |f|
   = f.hidden_field :resource_type, :value => @type.camelcase
-  = f.text_field :resource_id, :placeholder => "Choose a #{@type.humanize.downcase}", :required => true, :label => @type.humanize
-  = f.text_field :taxon_id, :placeholder => t(:choose_a_taxon)
+  = f.text_field :resource_id, :placeholder => t("choose_a_#{@type.humanize.downcase}".to_sym), :required => true, :label => t(@type.to_sym)
```
kueda: Nothing broken here, but when using `t` with dynamic values like this it's usually good practice to set a `default`, otherwise you might get some ugly results if the key hasn't been translated into the user's locale.

app/controllers/users_controller.rb
```diff
@@ -18,20 +18,19 @@ class UsersController < ApplicationController
...
   
-  MOBILIZED = [:show, :dashboard, :new, :create]
+  MOBILIZED = [:show, :new, :create]
   before_filter :unmobilized, :except => MOBILIZED
   before_filter :mobilized, :only => MOBILIZED
 
   protect_from_forgery unless: -> {
     request.parameters[:action] == "search" && request.format.json? }
 
-  caches_action :dashboard,
```
pleary: Any reason to remove the dashboard page cache? I'll be interested to play with this once it's merged to make sure we're caching everything we can since dashboard can be very slow to load

kueda: With Scott's change the hard stuff should be getting loaded in `dashboard_updates` and not in `dashboard`, and the former is still cached.

app/views/shared/_taxon.html.erb
```diff
@@ -1,30 +1,39 @@
...
+  trunclength ||= truncate ? 23 : 500
   if taxon.species_or_lower?
     pieces = taxon.name.split
     sciname = if taxon.variety? && pieces.size == 3
       "<span class=\"sciname #{taxon.rank}\">#{pieces[0..1].join(' ')} <span class=\"rank\">var.</span> #{pieces[2..-1].join(' ')}</span>"
     elsif taxon.subspecies? && pieces.size == 3
       "<span class=\"sciname #{taxon.rank}\">#{pieces[0..1].join(' ')} <span class=\"rank\">ssp.</span> #{pieces[2..-1].join(' ')}</span>"
     else
-      "<span class=\"sciname\">#{taxon.name}</span>"
+      "<span class=\"sciname\">#{truncate(taxon.name,length: trunclength)}</span>"
```
pleary: if this just for display, CSS can do string truncation on the fly based on element size

app/views/users/_followees.html.haml
```diff
@@ -1,9 +1,7 @@
-.col-md-3
+.col-md-3.col-sm-3
   .media
     .media-left
-      = link_to(                                                                     |
-        image_tag(friend.icon.url(:thumb), :class => 'media-object img-circle', :width => 40), |
-        person_by_login_path(friend.login))                                          |
+      = link_to(image_tag(friend.icon.url(:thumb), :class => 'media-object img-circle', :style => "max-width:50px;max-height:50px;", :width => 50), person_by_login_path(friend.login))
```
pleary: another CSS thing. Ideally we don't do any inline styles and leave this for CSS to deal with

app/controllers/users/registrations_controller.rb
loarie: ok - will hold off on this

app/controllers/users_controller.rb
loarie: yes - just so that the onboarding material and the sidebar stuff (e.g. announcements) aren't cached

app/views/shared/_taxon.html.erb
loarie: yes just for display - how do you do that?

app/controllers/users_controller.rb
pleary: Looks like `/dashboard` still queries the updates index for `user.recent_notifications` makes a few queries to the obs index to fetch observations via `get_local_onboarding_content` which will get called with every load. Is it OK if eventually that sidebar stuff is cached for just 15 minutes? Doesn't seem like a blocker for release, but I'd probably still prefer to cache `/dashboard` eventually

loarie: fine with me

Issue #530, opened 2015-04-21
Fiddling with improvements to counting IDs


loarie:
We currently count IDs as #IDs made for others. A problem is that this gives equal weight to confirming agrees or coarse rank IDs as it does for IDs that actually add new information. Here I'm fiddling with a now way to count IDs that better reflects novel contributions.

I've broken down a user's IDs into 5 groups:
1) 'novel' IDs as all active IDs where the ID was the earliest ID to equal the observation's community ID
2) 'confirming/coarse' IDs as are all active IDs where the ID was not the earliest ID to equal the observation's community ID
3) 'no_cid' which includes all active IDs on observations that don't yet have a community ID because they are too young or esoteric
4) 'dissents' which are all active IDs that don't match the observation's community ID
5) 'inactive' which are inactive IDs usually because they change their minds or taxon changes

Here's how I'm fetching my novel IDs:
`SELECT count(*)
FROM identifications i
INNER JOIN observations ON (i.observation_id = observations.id)
WHERE i.current = true
AND i.user_id = 477
AND observations.community_taxon_id = i.taxon_id
AND i.created_at = (select MIN(created_at) 
       from identifications i2 where i2.observation_id = i.observation_id);`

This graph shows the breakdown of a few users categories:
![screen shot 2015-04-21 at 8 38 22 am](https://cloud.githubusercontent.com/assets/598026/7256385/9c66d05a-e803-11e4-8229-76034417971f.png)

The top 10 IDers counting by # of 'novel' IDs are: 
finatic (10285) => 19955
greglasley (9706) => 12933
maractwin  (10787) => 10466
kueda (1) => 10059
loarie (477) => 9921
jmaughn (2873) => 7764
dpom (16101_ => 7156
jon_sullivan (1115) => 7061
muir (1000) => 6667
sea-kangaroo (2991) => 6610 

And in terms of number of taxa they've been able to ID by counting # of 'novel' IDs are:
loarie (477) => 3581
kueda (1) => 3310
maractwin (10787) => 2337
finatic (10285) => 2204
greglasley (9706) => 2203
carmelo_lopez (15211) => 2162
sea-kangaroo (2991) => 2030
jon_sullivan (1115) => 1781
jaykeller (27743) => 1747
robberfly (13979) => 1607

The way I've done it here, 'novel' IDs include a user's own IDs as long as they were the first to get to whatever the community ID is. So this isn't a great metric of altruism, but it is a good metric of whether and what people actually know how to ID. 
[...]

PR #1047, opened 2016-07-08
Identification Categories

kueda:
Categorize IDs in an effort to find which IDs move the ball forward (improving), which are making more specific but so far unsupported claims (leading), and which are totally outside the community consensus (maverick). Still testing, but we'd like to see what the data look like in production.


kueda:
For future reference, this mostly came out of https://github.com/inaturalist/inaturalist/issues/530

PR #1058, opened 2016-07-13
Content type detection

kueda:
Force paperclip to check the content type of the file instead of relying on the file name. This should prevent us from storing photo files without extensions and deal with some issues with Twitter previews. Older photos can still be repaired and rotated, but any future code dealing with the `path` or `url` of past photos attachments will need to re-download the image referenced in `original_url` before doing any work.


pleary:
I tested various things locally with my S3 account and everything seemed to work well. I made this one tiny change to the only place I saw using photo.file incorrectly. Merge is OK by me

PR #1099, opened 2016-08-08
Removing updates table

pleary:
We no longer need the `Updates` table now that we have `UpdateActions` and `UpdateSubscribers`, which seem to be working well


kueda:
Looks ready to go to me.

PR #1115, opened 2016-08-18
Stop using listed_taxa.taxon_ancestor_ids

pleary:
Relying on joining with taxa and using taxa.ancestry. There are 470k taxa and 8.1M listed_taxa so we're using a little extra space by duplicating ancestry in LT. But the wort part is all the delayed jobs we create to update LT.taxon_ancestor_ids when taxa change. This will get rid of all of those jobs.

This will not yet remove LT.taxon_ancestor_ids since that would take a while to compute again if there were a problem. We can create a migration to remove that column if this deploy works well for a while


kueda:
I get an exception when I try to filter a list by a taxon and show leaves, e.g. http://localhost:3000/lists/26710-alameda-stuff?q=&taxon=18&observed=any&rank=leaves&taxonomic_status=active

kueda:
That seems to have done it. Ready to go as far as I'm concerned.

Issue #1139, opened 2016-09-12
issues with users with login's beginning with integers


loarie:
e.g. user_id: 325478 login: '20aaron'
https://www.inaturalist.org/people/20aaron works but

https://www.inaturalist.org/users/20aaron redirects to https://www.inaturalist.org/users/2
and  https://www.inaturalist.org/observations/20aaron returns 404

loarie:
hmmm - apparently the bug here is that you shouldn't be able to make accounts starting with integers. Will try to figure out how these were made and edit this issue appropriately

loarie:
Apparently these accounts were created through Google - I'll try to reproduce now, but I suspect that the step where iNat creates a login based on a Google account is skipping this validation step.

The validation does seem to be in place on normal signup but we should probably try to be a bit more explicit with the error message (e.g. 'Use only letters, numbers and -_ please and make sure your login begins with a letter')
<img width="868" alt="screen shot 2016-09-12 at 2 00 27 pm" src="https://cloud.githubusercontent.com/assets/598026/18453564/bf455d66-78f4-11e6-96b3-56f4bd7c038e.png">

PR #1152, opened 2016-09-23
Do not suggest a login beginning with a number (#1139)

cyberhobo:
I've clarified the error message when signing up with an invalid login name, and prevent suggesting a login beginning with a number for new OAuth accounts to address #1139.

I've used this to get familiar with the code, happy to know if I'm in need of more schoolin'.


kueda:
Thanks!

cyberhobo:
Welcome! Feel free to assign any issues that might be good for learning to me. No guarantees, but so far I'm having fun.

Issue #1067, opened 2016-07-18
Identifications: "Search external name providers" looks like it's failing but not


carrieseltzer:
Lately when you're adding an ID to an observation, can't find a taxon and select "search external name providers", it always seems to fail: "No results found". However, in many cases it *is* finding a name and importing it. Then there's a delay before you can actually add the scientific name in the ID field.

Someone from American Samoa added lots of rainforest trees that aren't already in iNaturalist so [his observations unattached to taxonomy](http://www.inaturalist.org/observations/tavita_togia2016?utf8=%E2%9C%93&q=&search_on=&quality_grade=any&reviewed=&identifications=any&captive=&iconic_taxa%5B%5D=unknown&place_id=&swlat=&swlng=&nelat=&nelng=&taxon_name=&taxon_id=&day=&month=&year=&order_by=observations.id&order=desc&rank=&hrank=&lrank=&taxon_ids%5B%5D=&d1=&d2=&created_on=&site=&tdate=&list_id=&filters_open=true&view=table&changed_fields=&changed_since=&change_project_id=) are perfect for illustrating this. 

To replicate: 
1. Pick an observation from here: http://www.inaturalist.org/observations/tavita_togia2016?utf8=%E2%9C%93&q=&search_on=&quality_grade=any&reviewed=&identifications=any&captive=&iconic_taxa%5B%5D=unknown&place_id=&swlat=&swlng=&nelat=&nelng=&taxon_name=&taxon_id=&day=&month=&year=&order_by=observations.id&order=desc&rank=&hrank=&lrank=&taxon_ids%5B%5D=&d1=&d2=&created_on=&site=&tdate=&list_id=&filters_open=true&view=table&changed_fields=&changed_since=&change_project_id=
2. Search iNat species pages for the scientific name entered in the species guess. DON'T search in the id field yet. 
3. See that iNaturalist doesn't have that species (if it does, go back to step 1). Keep that tab open.
4. Go back to the observation and enter the scientific name in the id field. Select "search external name providers".
5. See "no results found".
6. Go back to the tab where you searched species pages and search again. See that it's there now. Edit the taxon to graft it if necessary.

I went through the steps above about 10 times last night and it worked almost every time, so if it doesn't work (i.e. misspelling or already in iNaturalist), try again with a different observation.

Further quirk: Adding a common name means you can add the ID more quickly on the observation page (there is a longer lag if it only has a scientific name). To replicate (first do steps 1-6 above):
[...]

PR #1163, opened 2016-10-05
Provide a default per page value for external taxon searches (#1067)

cyberhobo:
It looks like at least part of the problem in #1067 is that there is no default for options.perPage on external searches. This is generating a querystring with per_page=undefined, and that never shows results. This PR adds a default value of 10 like the internal request.

### DIFF
--- a/app/assets/javascripts/jquery/plugins/inat/taxon_autocomplete.js.erb
+++ b/app/assets/javascripts/jquery/plugins/inat/taxon_autocomplete.js.erb
@@ -166,7 +166,7 @@ $.fn.taxonAutocomplete = function( options ) {
       var thisSearch = Math.random();
       ac.searchInProgress = thisSearch;
       $.ajax({
-        url: "/taxa/search.json?per_page="+ options.perPage +
+        url: "/taxa/search.json?per_page="+ (options.perPage || 10) +
           "&include_external=1&partial=elastic&q=" + field.val( ),
         dataType: "json",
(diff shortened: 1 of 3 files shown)


cyberhobo:
Note that I reference the wrong issue in the commit, should be #1067.

kueda:
Hm, http://www.inaturalist.org/taxa/search?q=nevada&per_page=undefined doesn't just show no results, it yields an exception. I think your change here looks good, but could you look into the exception, which seems like the ultimate problem?

cyberhobo:
Do you mean maybe
http://www.inaturalist.org/taxa/search.json?q=nevada&per_page=undefined, or
should it default to JSON?

kueda:
Either. Specifying a page number that's not a number shouldn't yield an exception, it should just ignore that parameter.

cyberhobo:
I'm not sure why yet, but I don't get an exception locally on /taxa/search?q=nevada&per_page=undefined. I noticed I do get one for /taxa/search?q=nevada&partial=elastic, so I committed a fix for that. If you have any more info on the per_page=undefined exception in production that might be helpful.

cyberhobo:
I think I found it, I needed a search with results. I committed a fix, but I couldn't get a matching controller test to work. Probably I need to understand the testing environment better.

cyberhobo:
The exception was raised by a view, so my test needed to render views. That's included now, and I'll stop adding commits unless you see more issues.

kueda:
Thanks again!

Issue #1154, opened 2016-09-28
Atlases


loarie:
An atlas describes a set of continuous, non-overlapping places covering the Earth's land surface (atlas places)
<img width="999" alt="screen shot 2016-09-27 at 9 45 52 pm" src="https://cloud.githubusercontent.com/assets/598026/18901031/319f3630-84fc-11e6-8e01-0ae72bd8d9ef.png">
'atlas presences' are the a atlas places that encompass a species range while 'atlas absences' are the atlas places that are outside the species range
<img width="898" alt="screen shot 2016-09-27 at 9 46 10 pm" src="https://cloud.githubusercontent.com/assets/598026/18901049/677e8d32-84fc-11e6-8968-33dcd25978c1.png">
Atlas presences are determined from the existence of default_checklist listed_taxa for the species for the atlas places. Atlas absences are implied from the absence of default_checklist listed_taxa for atlas places
```ruby
# new model
Atlas
  # taxon
  # place_set
  # comments

  # after_create
  #  update manually_added = true for all listed_taxa in atlas place_set
  #  (or some other way of making sure they don't get destroyed
  #  if the observation they are sourced to is destroyed)
```
Here the place_set summarizes the set of admin_level 0,1,& 2 places that comprise the atlas (atlas places). The notation for place_set is [[explode_admin0],[explode_admin1]] where explode_admin0 is the array of admin_level 0 place ids that are exploded (e.g. swap out the country US for the set of descendant states) and explode_admin1 is the set of admin_level 1 place ids that are exploded (e.g. swap out the state CA with the set of descendant counties). 

Thus a place_set of ```[[],[]]``` would just be the set of all admin_level 0 places with nothing exploded. Likewise, ```[[1,6793,6712,6744,6878,7182,7190,7161,6903,6681,6986,6803],[]]``` would be the set of admin_level 0 places except for US,MX,CA,AU,BR,CL,AR,RU,CN,IN,ZA,NZ which would be replaced by (exploded) with their descendant admin_level 1 places

Here is some code which takes a taxon_id and a place_set and returns the set of atlas presences place_ids as determined by the listed_taxa:
```ruby
taxon_id = 36204 #western fence lizard

#place_set = [[],[]] # don't explode anything
place_set = [[1,6793,6712,6744,6878,7182,7190,7161,6903,6681,6986,6803],[]] #explode these countries: US,MX,CA,AU,BR,CL,AR,RU,CN,IN,ZA,NZ
[...]

PR #1176, opened 2016-10-14
first pass at atlases

loarie:
first pass at resolving #1154 by:

1. creating a model and simple interface for Atlases that use associated ExplodedAtlasPlaces to display atlas places and atlas place presences

2. creating AtlasAlterations that log when atlased listed taxa are created/destroyed or when ExplodedAtlasPlaces are created/destroyed

3. minor changes to listed_taxa so that atlaased listed_taxa aren't automatically created/destroyed

screen shot 2016-10-14 at 10 53 28 am


app/models/atlas.rb
```diff
@@ -0,0 +1,32 @@
...
+      where( "lists.type = 'CheckList'").
+      where( "lists.id = places.check_list_id" ).
+      where( "listed_taxa.taxon_id = ?", taxon ).
+      where( "(admin_level = 0 OR places.id IN (?))", exploded_place_ids_to_include ).
+      where( "places.id NOT IN (?)", exploded_place_ids_to_exclude )
+  end
+  
+  def get_exploded_place_ids_to_include_and_exclude
+    exploded_place_ids_to_include = [0]
+    exploded_place_ids_to_exclude = [0]
```
kueda: Why include place ID 0? There should be no such place.

app/models/atlas.rb
```diff
@@ -0,0 +1,32 @@
+class Atlas < ActiveRecord::Base
+  belongs_to :taxon
+  has_many :exploded_atlas_places
+  has_many :atlas_alterations
```
kueda: Both of these `has_many` relations need `dependent: :delete_all` and `inverse_of:`

db/migrate/20161012202458_create_atlases.rb
```diff
@@ -0,0 +1,9 @@
+class CreateAtlases < ActiveRecord::Migration
+  def change
+    create_table :atlases do |t|
+      t.references :user
```
kueda: You need to add indices for all of these relationships, so in this case, `add_index :user_id`, otherwise every query involving these columns is going to be a sequential table scan (i.e. slower than it needs to be). Much more important for the potentially larger tables like `atlas_alterations`

app/models/atlas.rb
```diff
@@ -0,0 +1,32 @@
+class Atlas < ActiveRecord::Base
+  belongs_to :taxon
```
kueda: You should also add `has_one :atlas, inverse_of :taxa` to `Taxon`.

app/models/listed_taxon.rb
```diff
@@ -488,6 +490,41 @@ def trickle_down_establishment_means(options = {})
...
+      )
+    end
+  end
+  
+  def log_destroy_in_atlas
+    if is_atlased?
+      atlas = Atlas.where(taxon_id: taxon_id).first
+      AtlasAlteration.create(
+        atlas_id: atlas.id,
+        user_id: user_id,
```
kueda: This would leave a record stating that the person who *created* the listed taxon then destroyed it, which might not be the case since other people (like curators and admins) can destroy listed taxa, and listed taxa can be automatically destroyed. I think you need a way to pass down the `current_user` from the controller, perhaps using the `updater` relationship, though you should confirm that `updater` gets set on destruction, b/c it probably doesn't (why would you store that if you're going to delete the record).

app/models/listed_taxon.rb
```diff
@@ -488,6 +490,41 @@ def trickle_down_establishment_means(options = {})
     ActiveRecord::Base.connection.execute(sql)
   end
   
+  def is_atlased?
+    if list.is_a?(CheckList) && list.is_default?
```
kueda: So you're intentionally excluding listed taxa on non-default-but-comprehensive lists like http://www.inaturalist.org/check_lists/69089-Aizoaceae-of-Marin-County--CA--US, at least to start?

kueda:
Why are there empty files like `app/views/exploded_atlas_places/create.js.erb`?

.gitignore
```diff
@@ -1,3 +1,5 @@
+config/initializers/secret_token.rb
```
kueda: What is this file for?

app/models/atlas.rb
loarie: The Place query you proposed here: https://github.com/inaturalist/inaturalist/issues/1154#issuecomment-250304586 doesn't work if exploded_place_ids_to_include = [] & exploded_place_ids_to_exclude = [] (e.g. an atlas of only countries with nothing exploded) so adding a dummy place id of 0 (ie exploded_place_ids_to_include = [0] and exploded_place_ids_to_exclude = [0]) was a hack which fixes that case. But is a bit klutzy and things would break if there actually was a place with ID 0. Can you bear with the klutzyness? If not, can you alter your query in some other way that would work in the exploded_place_ids_to_include = [] & exploded_place_ids_to_exclude = []  case?

kueda: There should never be a place with an ID of 0, so your code will work. Another approach would be to just check whether those arrays are blank and alter the query accordingly, e.g.

```ruby
places = Place.
  joins( listed_taxa: :list ).
  where( "lists.type = 'CheckList'").
  where( "lists.id = places.check_list_id" )
  where( "listed_taxa.taxon_id = ?", taxon )
places = if exploded_place_ids_to_include.blank?
  places.where( admin_level: 0 )
else
  places.where( "(admin_level = 0 OR places.id IN (?))", exploded_place_ids_to_include ).
end
unless exploded_place_ids_to_exclude.blank?
  places = places.where( "places.id NOT IN (?)", exploded_place_ids_to_exclude )
end
places
```

.gitignore
loarie: hmm I guess rails 4.2 is now using config/secrets.yml I'll do that locally to and fix that

app/controllers/atlases_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+    @atlas_places = @atlas.places
+    @atlas_presence_places = @atlas.presence_places
+    @atlas_alterations = @atlas.atlas_alterations
+    @atlas_place_json = {
+      type: "FeatureCollection", 
+      features: @atlas_places.map{|p| 
+        {presence: (@atlas_presence_places.map(&:id).include? p.id) ? true : false, id: p.id, type: "Feature", geometry: RGeo::GeoJSON.encode(p.place_geometry.geom)}
+      }
+    }
+    @listed_taxon = ListedTaxon.where(place_id: @atlas_presence_places.first.id).first
```
loarie: ugh thanks - that variable isn't even being used. Must have forgotten to delete

db/migrate/20161012202458_create_atlases.rb
loarie: fixed

app/models/atlas.rb
loarie: fixed

app/models/atlas.rb
loarie: perfect

app/models/atlas.rb
loarie: fixed

loarie:
@kueda the above should address all your change requests

kueda:
Hm, apparently I can't add another review. This is technically working for me, but I think the UI isn't really ready for prime time, so my only request is that you either make this admin-only for now, or try to keep it on the QT. I find the map pretty confusing, for example, and things like exploding and collapsing places work but the effects are only visible after refreshing the page. Other than that I think this is ready to test out.

PR #1192, opened 2016-11-10
stop updating project lists with taxa from project observations

### LINKED ISSUE #1190: Project Lists should not create listings from observations
currently project lists create listings from observations (like default place checklists and user life lists)

However this is breaking use-cases that use the project list as a manually curated reference, e.g.  the must-be-on-list rule as follows:
1. user adds an observation to a project with a must-be-on-list rule of something on the list
2. community changes the ID to something not on the list
3. listing is added to the list for that species based on that obs
4. now the curated project list has something on it that wasn't indented and observations of that are aggregated into the project

Proposal is to make it so that Project Lists do not create listings from observation (like non-default place checklists)

This might bother some people who use project lists to track the species represented by observations in their project, but they can (1) use obs/search for that or (2) manually update their project list as needed with taxa sourced to observations

The only complaints I've personally gotten about project lists relate to the must-be-on-list use-case, so I suspect that if we make this change the benefit experienced by those users will outweigh the cost to anyone still using project lists to track the species represented by project obs

loarie:
closes #1190 - I also pulled out project.show_from_place which hasn't been used in the UI since we stopped displaying project_lists by default but was still hanging around

### DIFF
--- a/app/controllers/shared/lists_module.rb
+++ b/app/controllers/shared/lists_module.rb
@@ -16,7 +16,7 @@ module Shared::ListsModule
   def show
     @find_options = set_find_options
     @view = params[:view] || params[:view_type]
-    @observable_list = observable_list(@list)
+    @observable_list = @list
     @q = params[:q] unless params[:q].blank?
     @search_taxon_ids = set_search_taxon_ids(@q)
     unless @search_taxon_ids.blank?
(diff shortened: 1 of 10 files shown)


loarie:
@kueda - you might want to review this - but should be ready to go

Issue #1217, opened 2016-12-07
Complete Sets


loarie:
A set of listed taxa is more useful if it can be interpreted as a complete set (e.g. ALL amphibians in Sonoma County as opposed to SOME). Currently, complete sets are indicated on iNat by adding a separate place checklist flagged as 'comprehensive'. This system isn't ideal because:
1. If I make a comprehensive list for Apiaceae of California but there are other listed-taxa for California Apiaceae on other lists, these will display in the identotron. To remedy this I'd have to delete listings on lists owned by other people which is usually not permitted.
2. The system allows multiple listings for the same taxon/place which is complicating curation of listings (e.g. if I delete a 'bad listing' from the default checklist it might still display on the place page/identotron because there are other listings for that place/taxon) and the display of listings (various listings might have different establishment means - which one do we show?). Primary-listings were built to solve the latter issue but its broken at the moment.
3. The philosophy behind comprehensive lists is that there's a static external source for a comprehensive list and a single owner of that list. In my experience this isn't ideal because taxonomy issues etc. mean that the external list usually requires curation, and sometimes the list owner isn't responsive which can create problems (e.g. if there's a comprehensive list for Apiaceae of California that is 'wrong' it prevents me from making listings in that taxon/place not in that list). I'm interested in exploring a more community-curated notion of completeness where numerous people work together to curate the set of listings that can be interpreted as a complete set via discussion around creating/destroying listings.

This proposal is for a different approach which, similar to 'atlases' #1154, uses an additional piece of information (in this case a 'complete set' for a taxon/place) to (1) make the existing set of relevant default-checklist listings interpretable as comprehensive and (2) track changes to this set via a log of Alterations.

``` ruby
# new model
CompleteSet
  # taxon
  # place
  # description
  # source
```
<img width="377" alt="screen shot 2016-12-06 at 4 57 57 pm" src="https://cloud.githubusercontent.com/assets/598026/20950565/5400d20a-bbd5-11e6-8fef-c7adc738c70f.png">

The complete_set show view would display the set of relevant listed-taxa from the default-checklist, the alteration log, comments, and any warnings (e.g. there is already a complete set for a parent of this taxon) 
<img width="958" alt="screen shot 2016-12-06 at 4 58 20 pm" src="https://cloud.githubusercontent.com/assets/598026/20950566/55507dfe-bbd5-11e6-8460-e2b918277fac.png">

## modifying Alterations
[...]

PR #1237, opened 2016-12-23
Complete sets

loarie:
this pull requests adds some improvements to the version of atlases thats in production and adds complete_sets https://github.com/inaturalist/inaturalist/issues/1217. Still needs work but usable and admin_only for now


app/controllers/complete_sets_controller.rb
```diff
@@ -0,0 +1,90 @@
...
+      place_id: @complete_set.place_id, without_taxon_id: @taxa.pluck(:id).join(",")
+    }
+    @num_obs = INatAPIService.observations(@observation_search_url_params.merge(per_page: 0)).total_results
+  end
+
+  def create
+    @complete_set = CompleteSet.new(params[:complete_set])
+    respond_to do |format|
+      if @complete_set.save
+        format.html { redirect_to(@complete_set, notice: 'Complete Set was successfully created.') }
```
kueda: Double quotes, please.

app/models/listed_taxon.rb
```diff
@@ -631,6 +636,20 @@ def self.species_for_infraspecies(lt)
...
   end
 
+  def self.get_defaults_for_taxon_place(place,taxon, options = {})
+    options[:limit] ||= 100
+    options[:limit] = 200 if options[:limit] > 200
+    taxon = Taxon.find_by_id(taxon) unless taxon.is_a?(Taxon)
+    return unless taxon
+    place = Place.find_by_id(place) unless place.is_a?(Place)
+    return unless place
+    place_descendant_ids = place.descendants.where('admin_level IN (?)',[0,1,2]).pluck(:id)
```
kueda: This would be more readable as `.where( "admin_level IN (?)", [Place::COUNTRY_LEVEL, Place::STATE_LEVEL, Place::COUNTY_LEVEL])`

app/models/check_list.rb
```diff
@@ -300,9 +298,9 @@ def self.refresh_with_observation(observation, options = {})
     end
     if observation && observation.research_grade? && observation.taxon.species_or_lower?
       Rails.logger.info "[INFO #{Time.now}] refresh_with_observation #{observation_id}, adding new listed taxa"
-      if atlas = Atlas.where(taxon_id: observation.taxon_id).first
-        new_place_ids = new_place_ids - atlas.places.map(&:id)
-      end
+      if Atlas.where("taxon_id IN (?)", taxon_ids).count > 0 || CompleteSet.where("taxon_id IN (?) AND place_id IN (?)",taxon_ids, current_place_ids).any?
+        new_place_ids = Place.where(id: new_place_ids).where("admin_level NOT IN (?)",[0,1,2]).pluck(:id) #if under atlas or complete set,
```
pleary: I nearly missed this comment. We generally put comments on their own lines above the code their commenting on:

```
  # here is a 
  # multiline comment
  do_something_non_obvious()
```

app/models/check_list.rb
```diff
@@ -3,6 +3,7 @@ class CheckList < List
   belongs_to :place
   belongs_to :taxon
   belongs_to :source
+  has_one :check_list_place, class_name: "Place", foreign_key: :check_list_id
```
pleary: This isn't needed since the `List` model has the association, and `CheckList` extends `List`

app/models/list.rb
```diff
@@ -8,9 +8,11 @@ class List < ActiveRecord::Base
                     comment_type: "item-description",
                     automated: false
   belongs_to :user
+  has_one :check_list_place, class_name: "Place", foreign_key: :check_list_id
   has_many :rules, :class_name => 'ListRule', :dependent => :destroy
   has_many :listed_taxa, :dependent => :destroy
   has_many :taxa, :through => :listed_taxa
+  has_one :check_list_place, class_name: "Place", foreign_key: :check_list_id
```
pleary: duplicate of line 11

app/controllers/atlases_controller.rb
```diff
@@ -1,98 +1,143 @@
...
-    @atlas = Atlas.find(params[:id])
     @atlas.destroy
-    
     respond_to do |format|
       format.html { redirect_to(@atlas.taxon) }
     end
   end
-  
+
+  def map
```
pleary: This is the action I made, but I think you've rolled it all over to atlases::show . If so, then this can go, as can the atlases/map.html.haml view, and the route

app/views/atlases/show.html.haml
```diff
@@ -1,47 +1,86 @@
...
+
+- content_for :extracss do
+  = stylesheet_link_tag "atlases/show"
+  = stylesheet_link_tag "atlases/map"
 
 .container
   .row
     .col-xs-12
+      - if @num_obs > 0
+        .alert.alert-warning{:role => "alert"}
```
pleary: We are preferring in new code to use the newer Ruby hash syntax `{ role: "alert" }` in ruby and HAML files (or alternatively with a symbol as the value `{ role: :alert }` if applicable). It's easier to type, and is easier when moving back and forth between Ruby and JS. There are several cases of this in these views. Not a huge deal nor a blocker, but another for the "style guideline"

PR #1268, opened 2017-01-19
Taxon Page Controlled Terms

kueda:
Use controlled terms instead of observation fields on the taxon page.

### DIFF
--- a/app/webpack/taxa/photos/components/photo_browser.jsx
+++ b/app/webpack/taxa/photos/components/photo_browser.jsx
@@ -151,7 +151,8 @@ const PhotoBrowser = ( {
         eventKey={values[0].controlled_attribute}
         active={grouping.param === `field:${values[0].controlled_attribute.label}`}
       >
-        { values[0].controlled_attribute.label }
+        { I18n.t( _.snakeCase( values[0].controlled_attribute.label ),
+          { defaultValue: values[0].controlled_attribute.label } ) }
       </MenuItem>
     ) )
   );
--- a/app/webpack/taxa/photos/components/photo_browser.jsx
+++ b/app/webpack/taxa/photos/components/photo_browser.jsx
@@ -27,6 +27,8 @@ const PhotoBrowser = ( {
   setParam,
   setTerm,
   showTaxonPhotoModal,
+  selectedTerm,
+  selectedTermValue,
   terms,
   showTaxonGrouping,
   place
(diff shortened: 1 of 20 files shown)


app/webpack/taxa/show/containers/charts_container.js
```diff
@@ -7,6 +7,8 @@ import {
   openObservationsSearch
 } from "../ducks/observations";
 
+const TermsToChart = ["Life Stage"];
```
kueda: This is anal, but can we reserve capitalized camelcase for classes and modules and the like? I'd have this as `TERMS TO_CHART`

app/webpack/taxa/show/components/charts.jsx
```diff
@@ -253,61 +236,39 @@ class Charts extends React.Component {
...
+      _.each( this.props.chartedFieldValues, ( values, termID ) => {
+        fieldValueTabs.push( (
+          <li role="presentation" key={ `charts-field-values-${termID}` }>
+            <a
+              href={ `#charts-field-values-${termID}` }
+              aria-controls={ `charts-field-values-${termID}` }
+              role="tab"
+              data-toggle="tab"
+            >
+              { values[0].controlled_attribute.label }
```
kueda: This should be translated if we have a translation for this term, so maybe `I18n.t( _.snakeCase( values[0].controlled_attribute.label ), { defaultValue: values[0].controlled_attribute.label } )`

app/webpack/taxa/photos/components/photo_browser.jsx
```diff
@@ -143,13 +145,13 @@ const PhotoBrowser = ( {
...
       <MenuItem
-        key={`grouping-chooser-item-${term.name}`}
-        eventKey={term.name}
-        active={grouping.param === `field:${term.name}`}
+        key={`grouping-chooser-item-${values[0].controlled_attribute.label}`}
+        eventKey={values[0].controlled_attribute}
+        active={grouping.param === `field:${values[0].controlled_attribute.label}`}
       >
-        { term.name }
+        { values[0].controlled_attribute.label }
```
kueda: Needs translation too.

pleary:
I think I've addressed all these comments in my latest commit. I also made it so we scroll to the top of the page when changing taxa, with the exception of the links in the taxonomy tab tree. I like it, but others can see what they think when this is on staging shortly

PR #1290, opened 2017-02-16
ES5 compatible syntax

pleary:
Updated Elasticsearch syntax to be (mostly) compatible with ES 5.x


kueda:
Oops, forgot to comment. It all looks good to me, from what I can tell.

PR #1329, opened 2017-04-10
Prevent submenus from closing if the user is using the search inputs

kevinhikaruevans:
This very minor thing was bugging me.

When I'm searching something up on inaturalist using the header search box (see below), the header submenus disappear if you're typing something in but accidentally move the cursor outside of the dropdown box.

This change prevents the box from closing if the user is typing something in. 

image


kueda:
Thanks! Will get this deployed soon.

PR #1341, opened 2017-04-22
Delaying refresh_listed_taxon

pleary:


pleary:
I haven't really described this. The main idea is to delay the processing of `CheckList.refresh_listed_taxon`, and possible reduce the number of times it is called. I suspect there are multiple calls to refresh the same listed taxon many times over, and this will make sure we don't refresh the same one more than once a half hour. As a result, `CheckList.refresh_with_observation` (which is by far our slowest DJ) should finish much faster, and we should get a better sense for how much `CheckList.refresh_listed_taxon` is costing us.

This does introduce a delay where there wasn't one before, but during CNC we were several days behind on the `CheckList.refresh_with_observation` jobs and no one ever noticed, so I think it's OK. Any thoughts on this @kueda?

kueda:
Looks good to me. I think in the past we've generally tried to avoid spawning more delayed jobs from other delayed jobs, but the time delay probably keeps that from getting out of control.

pleary:
Thanks for checking it out. That's a good point. My hope is that with the delay, it'll ultimately reduce the work we need to do. There will always be more obs than taxa going forward, so if we can get closer to having the cardinality of DJs be on the order of the cardinality of taxa, and not that of obs, that'll help us as we scale up. Anyway - this is an experiment so we'll see. I'll merge this in now

PR #1359, opened 2017-05-06
Translatewiki updates for 2017-05-05

kueda:


config/locales/bg.yml
```diff
@@ -886,7 +886,7 @@ bg:
       h: 1ч
       d: 1д
       M: 1мес
-      y: 1г
+      1: 1г
```
kueda: @Nikerabbit, this has happened before and it causes a bug for us. The key should be `y`, not `1`. Is there a way you guys can prevent this from happening in the future?

Nikerabbit: The reason is clearly that `y` is intepreted as boolean true at some point, and then casted to a string at another point.

I believe this is correct according to the yaml spec, see for example https://makandracards.com/makandra/24809-yaml-keys-like-yes-or-no-evaluate-to-true-and-false

Unfortunately, the behavior between different yaml parsers vary a lot, so not all libraries handle this consistently.

The sure solution is to quote it as `'y'` in en.yml.

PR #1378, opened 2017-06-01
site specific announcements can also be locale specific

loarie:
The Canada site wants different announcements if the locale is French or English. Currently, if the Site is set, it fetches announcements of all locales for that site. This change is meant to only fetch announcements with a specific locale for a site if announcements have both site and locale set.


loarie:
@kueda can you check this out. It works, but not sure its the most elegant way to grab announcements with a given site param with locale = nil or locale = current locale

PR #1409, opened 2017-07-07
Index curator coordinate access

kueda:
One way to address the geoprivacy issue @tiwane raised: https://inaturalist.slack.com/archives/C02F8RN5Q/p1499397827435545

### DIFF
--- a/app/es_indices/observation_index.rb
+++ b/app/es_indices/observation_index.rb
@@ -168,7 +168,17 @@ class Observation < ActiveRecord::Base
           project_observations.select{ |po| po.curator_identification_id.nil? }.
             map(&:project_id)).compact.uniq,
         project_observations: (indexed_project_ids || project_observations).map{ |po|
-          { project_id: po[:project_id], user_id: po[:user_id], uuid: po[:uuid] } },
+          {
+            project_id: po[:project_id],
+            user_id: po[:user_id],
+            uuid: po[:uuid],
+            curator_coordinate_access:
+              po.respond_to?(:prefers_curator_coordinate_access?) ?
+                po.prefers_curator_coordinate_access?
+                :
+                po[:curator_coordinate_access]
+          }
+        },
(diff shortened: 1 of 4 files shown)


app/es_indices/observation_index.rb
```diff
@@ -168,7 +168,17 @@ def as_indexed_json(options={})
           project_observations.select{ |po| po.curator_identification_id.nil? }.
             map(&:project_id)).compact.uniq,
         project_observations: (indexed_project_ids || project_observations).map{ |po|
-          { project_id: po[:project_id], user_id: po[:user_id], uuid: po[:uuid] } },
```
pleary: We shouldn't touch this right now. Node will fetch project observation preferences dynamically from the DB, so you don't need to add anything to the index.

I'm currently working on adding preferences to the index, and lots of other stuff, but that's not ready right now. So for this fix you should continue to use the old way, which means you don't need to change anything on the Rails end

app/es_indices/observation_index.rb
```diff
@@ -240,12 +250,26 @@ def self.prepare_batch_for_index(observations, options = {})
     # fetch all project_ids store them in `indexed_project_ids`
     if options.blank? || options[:projects]
       connection.execute("
-        SELECT observation_id, project_id, curator_identification_id, uuid, user_id
```
pleary: The changes here can be removed too. You won't need to change anything in `observation_index`

spec/lib/elastic_model/indices/observation_index_spec.rb
```diff
@@ -162,6 +162,18 @@
     expect( json[:owners_identification_from_vision] ).to be true
   end
 
+  it "indexes project observation curator_coordinate_access in bulk" do
```
pleary: This can also be removed

kueda:
Pref. loading happens from the DB in the API, Patrick is working on more sweeping index changes.

PR #1450, opened 2017-08-11
Config changes

pleary:
Moving anything that isn't identical across all sites into the Site model/DB, and leaving everything else in `config.yml` (and removing them from site preferences)


app/models/provider_authorization.rb
```diff
@@ -157,10 +157,6 @@ def refresh_access_token!
        'Content-Type' => 'application/x-www-form-urlencoded'
      }
     }
-    if CONFIG.ca_path || CONFIG.ca_file
-      options[:ssl_ca_file] = CONFIG.ca_file
-      options[:ssl_ca_path] = CONFIG.ca_path
-    end
```
kueda: I don't think this will cause any probs in production, but it might in local dev environments where the CA file might be somewhere other than the default, right? Not a blocker, but something to keep in mind.

pleary: Yep, noted. We weren't using this in production or Staging, and I'm not using it locally, so I was just trying to trim the fat. Easy to add back if we want

PR #1474, opened 2017-09-13
Move project icons to s3.

kueda:
Something like the following to update the data:

```ruby
start = Time.now
updated = 0
test = true
Project.find_each do |project|
  puts project.id
  icon = Dir["#{Rails.root}/public/attachments/projects/icons/#{project.id}/original/*"].first
  if icon.present?
    project.update( icon: File.open( icon ) ) unless test
    updated += 1
    puts "\tuploading"
  else
    puts "\tno icon"
  end
end
puts
puts "Updated #{updated} projects in #{Time.now - start}s"
puts
```

### DIFF
--- a/app/models/project.rb
+++ b/app/models/project.rb
@@ -115,6 +115,7 @@ class Project < ActiveRecord::Base
       path: "projects/:id-icon-:style.:extension",
       url: ":s3_alias_url",
       default_url: "/attachment_defaults/general/:style.png"
+    invalidate_cloudfront_caches :icon, "projects/:id-icon-*"
   else
     has_attached_file :icon,
       styles: { thumb: "48x48#", mini: "16x16#", span1: "30x30#", span2: "70x70#", original: "1024x1024>" },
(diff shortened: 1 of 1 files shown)


app/models/project.rb
```diff
@@ -104,11 +104,25 @@ def aggregation_preference_allowed?
...
+      storage: :s3,
+      s3_credentials: "#{Rails.root}/config/s3.yml",
+      s3_protocol: CONFIG.s3_protocol || "https",
+      s3_host_alias: CONFIG.s3_host || CONFIG.s3_bucket,
+      bucket: CONFIG.s3_bucket,
+      styles: { thumb: "48x48#", mini: "16x16#", span1: "30x30#", span2: "70x70#", original: "1024x1024>" },
+      path: "projects/:id-icon-:style.:extension",
+      url: ":s3_alias_url",
+      default_url: "/attachment_defaults/general/:style.png"
+  else
```
pleary: I'd also add in the CloudFront cache invalidator, which will send invalidation requests when people update their icon
```invalidate_cloudfront_caches :cover, "projects/:id-icon*"```

Issue #1501, opened 2017-10-13
Add hyperlinks to URL formatted text in observation fields on obs/show


loarie:
https://www.inaturalist.org/observations/8381375
<img width="515" alt="screen shot 2017-10-13 at 6 44 17 am" src="https://user-images.githubusercontent.com/598026/31549100-f5c5953e-afe1-11e7-8f91-6ff310df9d64.png">

PR #1534, opened 2017-11-07
make observation fields text urls clickable

wykhuh:
@kueda When the observation field has format of `text`,  if value contains an url, make a link.

Issue https://github.com/inaturalist/inaturalist/issues/1501

screen shot 2017-11-07 at 7 56 55 am

### DIFF
--- a/app/webpack/observations/show/components/observation_field_value.jsx
+++ b/app/webpack/observations/show/components/observation_field_value.jsx
@@ -1,6 +1,7 @@
 import React, { PropTypes } from "react";
 import { Popover, OverlayTrigger } from "react-bootstrap";
 import SplitTaxon from "../../../shared/components/split_taxon";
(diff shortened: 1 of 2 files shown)


app/webpack/observations/show/components/observation_field_value.jsx
```diff
@@ -1,6 +1,7 @@
 import React, { PropTypes } from "react";
 import { Popover, OverlayTrigger } from "react-bootstrap";
 import SplitTaxon from "../../../shared/components/split_taxon";
+import ObservationFieldText from "../../../shared/components/observation_field_text";
```
kueda: This new component is probably unnecessary. https://github.com/inaturalist/inaturalist/blob/master/app/webpack/shared/components/user_text.jsx might do the trick.

kueda:
I'm guessing you don't have any plans to work on this further.

PR #1645, opened 2018-01-29
Translatewiki updates for 29 Jan 2018

kueda:


config/locales/qqq.yml
```diff
@@ -1528,6 +1528,18 @@ qqq:
           the vow_or_con block, please do not make @ references to keys that
           have not been translated.
         con: Translation of the word "consonant"
+        a: Only customize if these letters are not considered vowels. Do not translate
```
kueda: @translatewiki can you please stop re-adding everything in the `:@vow_or_con` block after `con` to the qqq file? For one thing the documentation is redundant, and for another it causes bugs for us. If there's a way to add a single message for the `:@vow_or_con` block without adding messages for all the keys within it please let me know. That's what I was trying to do initially but then I ran into these bugs.

PR #1714, opened 2018-04-05
Translatewiki updates for 5 April 2018

kueda:


config/locales/cs.yml
```diff
@@ -0,0 +1,5446 @@
...
+  find_your_current_location: Najít vaši aktuální lokalitu
+  finished: Dokončeno
+  finished_with_page: Práce na stránce dokončena
+  first_confirmed_inat_observation: První potvrzené pozorování %{site_name}
+  first_connect_your_provider_account_to_your_site: 'Nejdříve propojte svůj účet na
+    %{provider} se stránkou %{site_name}:'
+  first_inat_observation: První pozorování %{site_name}
+  first_observation: První pozorování
+  first_observation_by_date_added: První pozorování podle data přidání
+  fir
```
kueda: @translatewiki these appear to be mis-translated. It should be a yaml list like we have in `en.yml`:

```yaml
    abbr_month_names:
      - 
      - Jan
      - Feb
      - Mar
      - Apr
      - May
      - Jun
      - Jul
      - Aug
      - Sep
      - Oct
      - Nov
      - Dec
```

This problems seems to be present in `cs.yml` and `el.yml`.

kueda: Hm, I'm not sure this actually left a comment on the line that I wanted, but the problem is with `abbr_month_names` in `cs.yml` and `el.yml`.

kueda:
@translatewiki, `cs.yml` and `el.yml` still have numbered lists under `abbr_month_names` (and possibly elsewhere. Can you please fix this so they use the same formatting as `en.yml`, i.e.

```
abbr_month_names:
  - 
  - Jan
  - Feb
  - Mar
  - Apr
  - May
  - Jun
  - Jul
  - Aug
  - Sep
  - Oct
  - Nov
  - Dec
```

Nikerabbit:
Thanks for processing the updates.

All the languages exported from translatewiki.net have numbered lists under `abbr_month_names`. That is because the first element marked as ignored because there is nothing translate – and the expectation is that missing values are taken from the fallback language – and because there is no way to have a list with missing entries it has to be a map.

The developer who created `abbr_month_names` probably thought it is a clever way to do things, while in practice it causes a lot of problems instead of applying the `-1` to the array index in the code, or better yet using a map in the first place.

Yet another thing is that these translations should be in a library shared between different projects and not translated over and over again for each project.

This is not your fault, but nevertheless, I am not keen on adding hacky exceptions to our file format support code.

PR #1738, opened 2018-04-24
Index spam status

### LINKED ISSUE #1735: Index spam status, suspended status for users
We want to keep spam out of search results, so we should index whether or not something is spam. This would look something like `spam: known_spam? || owned_by_spammer?` and should be added to the index of any model that can be marked as spam (Observation, Place, Project, User, Comment, Identification, at a minimum). I suspect we could do this in `ActsAsElasticModel`, but perhaps it would be less magical to just add it to each index file?

We also want to hide suspended users (at least by default), so we should add `suspended: suspended?` to `User#as_indexed_json`, in the initial `no_details` part.

Finally, this means everything a user has made needs to get re-indexed after their spam or suspension status changes.

kueda:
Add `spam` attribute to several indices of models can that can be spam, and added `suspended` to the user index. Closes #1735

### DIFF
--- a/app/es_indices/identification_index.rb
+++ b/app/es_indices/identification_index.rb
@@ -62,7 +62,8 @@ class Identification < ActiveRecord::Base
       } : nil,
       vision: vision,
       disagreement: disagreement,
-      previous_observation_taxon_id: previous_observation_taxon_id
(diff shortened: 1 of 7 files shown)


pleary:
I added some preloading to save queries when bulk indexing. Otherwise looks good to me

PR #1790, opened 2018-05-17
add dwca options for generating file with all taxa, posting archive

pleary:
This add options to tools/dwca that allows creating a duplicate archive which includes a taxa.csv file that includes all active taxa, and allows sending a POST request to an endpoint when that archive is updated. This also modifies parameters for INatAPIService methods to use a hash of options rather than an ever increasing list of parameters

### DIFF
--- a/app/models/year_statistic.rb
+++ b/app/models/year_statistic.rb
@@ -128,7 +128,7 @@ class YearStatistic < ActiveRecord::Base
         params[:locale] = locale
       end
     end
-    JSON.parse( INatAPIService.get_json("/observations/tree_taxa", params, 3, 30 ) )["results"] rescue nil
+    JSON.parse( INatAPIService.get_json("/observations/tree_taxa", params, { timeout: 30 } ) )["results"] rescue nil
   end
 
   def self.observations_histogram( year, options = {} )
@@ -144,7 +144,7 @@ class YearStatistic < ActiveRecord::Base
     if site = options[:site]
       params[:site_id] = site.id
     end
-    JSON.parse( INatAPIService.get_json("/observations/histogram", params, 3, 30 ) )["results"][params[:interval]]
+    JSON.parse( INatAPIService.get_json("/observations/histogram", params, { timeout: 30 } ) )["results"][params[:interval]]
   end
 
   def self.identifications_histogram( year, options = {} )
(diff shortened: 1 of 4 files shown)


kueda:
Looks fine to me. It's a bit weird that we're adding non-DwC-A functionality to the DwC-A generation script, but I suppose this is the most reasonable place to put it. Also, I'm assuming it's not a problem that the additional file contains all taxa and not just taxa represented by the observations? You could imagine achieving something like this by adding a more conventional DwC taxa extension file to the archive with its attendant metadata (we actually have the `make_taxon_data` method for this, but it looks kind of janky, db-only, only species for some reason, etc), but I'm guessing you rejected that b/c it was missing something you needed.

pleary:
I agree about it being a little strange to add the POSTing in the dwca script. I considered adding it in a separate cron task, but thought it could be a little tricky since we never know when the dwca generation will be done. If we just check regularly then the checking script will have to keep some state so it knows when it has POSTed the most recent version to avoid making multiple requests per week. So combining them seemed like the solution with the fewest components.

I did consider and reject having taxa.csv be a proper dwca extension because of limitations of the star schema used by dwca. In our GBIF archive, the core of the archive is observations, and there are extensions off of that such as media. If we added taxa as an extension, that extension _should_ reference the core, meaning we _should_ include a core_id column in taxa. But I don't want to repeat taxa in the taxa file, I just want a taxon listed once even if it has thousands of observations. You can get away with extensions that don't reference the core, but that's only by exploiting an accidental omission in the DWC-a XML schema, and I think that is generally frowned upon. The users that want the data do also want all active taxa known to iNat even if they don't have observations, and you can't add extensions off core data that doesn't exist.

Lastly, if the script were instant I would have just had it run twice to generate the different versions, rather than what I have it doing here which is making the main files, archiving them, then adding the taxa file at the end and making a second archive. But its taking something on the order of 18 hours to generate the GBIF dwca, and I didn't want to create a second long-running script we need to run every week.

Something to reconsider if we refactor exports. Assuming we can generate exports quickly, I think it would be great to have in addition to the GBIF dwca, some separate exports for all core types people might want to download. So we could provide a weekly-generated standalone taxa archive, or if we allowed users to export other types than observations, point them to a new exporter (it only takes about 10 minutes to export all taxa using the API).

PR #1793, opened 2018-05-21
Remove annotations when they are no longer relevant (closes #1696)

### LINKED ISSUE #1696: Reassess annotations when obs.taxon changes
There are at least three unhandled two situations I can think of in which annotations can become irrelevant:

1. The `observation.taxon` changes after an annotation was added (e.g. it used to be Plantae and someone added a Flowering Phenology=flowering annotation, then shifted to Animalia, and now that annotation is absurd)
1. The taxon associated with an observation gets moved around the tree, e.g. the taxon _Bibim bap_ is grafted to Plantae, and someone adds a a Flowering Phenology=flowering annotation, then _Bibim bap_ gets moved to Animalia, and the annotation no longer makes sense.
1. Someone edits the `ControlledTermTaxon` entries associated with a controlled attribute after annotations have been added.

I propose we delete the irrelevant annotations in all these cases.

kueda:
Removes taxon-specific annotations when taxonomy or taxon associations change.

### DIFF
--- a/app/models/annotation.rb
+++ b/app/models/annotation.rb
@@ -175,6 +175,9 @@ class Annotation < ActiveRecord::Base
   def self.reassess_annotations_for_taxon_ids( taxon_ids )
     Annotation.
         joins(
+          controlled_value: [
+            { controlled_term_taxa: :taxon }
+          ],
           controlled_attribute: {
             controlled_term_taxa: {
               taxon: :taxon_ancestors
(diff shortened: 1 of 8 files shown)


app/models/annotation.rb
```diff
@@ -172,4 +172,32 @@ def touch_resource
...
+    Annotation.
+        joins(
+          controlled_attribute: {
+            controlled_term_taxa: {
+              taxon: :taxon_ancestors
+            }
+          }
+        ).
+        where( "taxon_ancestors.ancestor_taxon_id IN (?)", taxon_ids ).
+        includes(
```
pleary: I was looking at the queries this would generate if run on a large taxon like Animalia. Looks like we can reduce the number of queries needed for validation with a little more preloading. This worked for me locally:

```
        includes(
          { resource: :taxon },
          controlled_value: [ :taxa, :excepted_taxa, {
            controlled_term_taxa: :taxon
          } ],
          controlled_attribute: [ :values, :taxa, :excepted_taxa, {
            controlled_term_taxa: :taxon
          } ]
        ).
```

pleary:
This looks good to me. Validating annotations takes a handful of queries which are currently unavoidable with preloading, but hopefully taxa don't change much, particularly large clades.

PR #1815, opened 2018-06-12
Forget user

kueda:
Adds a method to destroy the user but also several others sources of data, including:

* DeletedUser
* DeletedObservation
* DeletedPhoto
* DeletedSound
* S3 photos
* S3 sounds
* Logstash activity (requires an additional config)


kueda:
@pleary, I think the only weird part of this is the added config of the logstash ES host. I'm pretty sure this represents the only direct connection between the Rails app and that ES instance, so if this seems like overkill we could just make that an optional parameter of `User.forget` or put it in a script or something.

PR #1843, opened 2018-07-06
Remove update subscribers table

pleary:
The main changes here are to remove the `update_subscribers` table and instead use the `UpdateAction` Elasticsearch index as a persistent store for subscriber and viewed_subscriber data. UpdateAction records will only be added to Elasticsearch when they are created, and any future changes will happen with `update_by_query` calls with inline scripts for data modification (adding/removing subscribers, or marking subscribers as having viewed the action). By modifying data in-place in ES, we should avoid race conditions and data conflicts (note the few added `try_and_try_again` blocks catching conflict errors mean a race condition was prevented).

The other main changes in the production application code are about querying these data, and fetching subscriber information from the ES docs instead of from the DB via AREL queries. 

The last big change is in specs, where HasSubscribers is now disable by default. This is to prevent having to declare `enable_elastic_indexing(UpdateAction)` all over the place in tests.


kueda:
Looking good to me.

PR #1870, opened 2018-08-03
React 16 update

kueda:
Updated react and all relevant dependencies both to keep up to date but also to hopefully address some mysterious bugs IE users have been having with the uploader.


kueda:
I'll merge this in next week when I'll be more available to fix the inevitable problems, just getting this up here so I don't forget.

PR #1892, opened 2018-08-27
Rank level validations closes #1889

### LINKED ISSUE #1889: Avoid taxa with rank_level >= parent | <= children
There are currently about 738 active taxa in the db with rank levels equal to or greater than their ancestors
```
Select COUNT(*)
from taxa a
  join taxa b on a.id = any(string_to_array(b.ancestry, '/')::int[])
  WHERE a.rank_level <= b.rank_level AND b.is_active = TRUE;
```
These create issues with models that require taxa with coarsening ancestries. While they can be fixed, new ones are continually introduced.

We need validations to prevent a taxon being saved if:
1) the rank is greater than or equal to its parent
2) the rank is less than or equal to any children

@kueda are there any other edge cases I didn't think of here?

loarie:
@kueda - here's a fix for #1889. this broke a bunch of specs but in all cases because we haven't been explicit in making spec taxa with ranks not because it broke anything in the non-spec code. Travis is now happy with this and I've done some test on staging, but would appreciate any other checks you can do


app/models/taxon.rb
```diff
@@ -872,6 +874,21 @@ def taxon_cant_be_its_own_ancestor
       errors.add(self.name, "can't be its own ancestor")
     end
   end
+  
+  def rank_level_must_be_finer_than_parent
+    return if parent.nil?
+    
+    if (parent.rank_level.nil? || rank_level.nil?) || (parent.rank_level.to_f <= rank_level.to_f)
```
kueda: I might get rid of the nil checks here. `rank_level.to_f` will be 0.0 if rank_level is nil, and if the rank_level or parent rank_level is nil, "rank level must be finer than parent" is a confusing error message. Better to add separate validations to check for that.

spec/models/taxon_merge_spec.rb
```diff
@@ -1,10 +1,10 @@
 require File.dirname(__FILE__) + '/../spec_helper.rb'
 
 def setup_taxon_merge
-  input_genus = Taxon.make!( rank: Taxon::GENUS )
-  @input_taxon1 = Taxon.make!( rank: Taxon::SPECIES, parent: input_genus )
-  @input_taxon2 = Taxon.make!( rank: Taxon::SPECIES, parent: input_genus )
-  @input_taxon3 = Taxon.make!( rank: Taxon::SPECIES, parent: input_genus )
+  @input_genus = Taxon.make!( rank: Taxon::GENUS )
```
kueda: If this isn't always going to be a genus and will be assigned a new rank in different tests, please call it `@input_ancestor` or something

spec/models/taxon_spec.rb
```diff
@@ -115,6 +115,16 @@
   it "should strip leading space" do
     expect( Taxon.make!( name: "   Leading space" ).name ).to eq "Leading space"
   end
+  
+  it "should prevent creating a taxon with a rank coarser than the parent" do
+    parent = Taxon.make!( rank: Taxon::GENUS)
```
kueda: Missing a space. Also, when writing new Ruby code, please use the new-style hash syntax and double quotes whenever possible. So below should be `taxon = Taxon.new( name: "balderdash", rank: Taxon::FAMILY, parent: parent )`

loarie:
@kueda ok fixed all of these, its passing travis. Re: plan for dealing with existing issues in DB. There are now none, but they keep getting introduced. I'll fix them again when we deploy

PR #1919, opened 2018-09-11
Atlases and comprehensive lists

kueda:
Currently, a comprehensive list causes problems with atlases, b/c atlases seem to assume listed taxa are only on a place's default checklist. However, the presence of a comprehensive list messes that up. Here's a concrete example: I was trying to do https://www.inaturalist.org/taxon_changes/39598 properly by makeing atlases for both of the output taxa. However, since there are comprehensive lists for Phrymaceae in California and for all counties of California, adding Erythranthe lewisii 777190 to any of the default checklists was impossible (if a comprehensive list exists, it's supposed to be comprehensive). In theory one could solve that problem by adding your new taxon to the relevant comprehensive lists, but it's currently impossible to choose an inactive taxon when adding to a list, and you can't make something like E. lewisii sensu stricto active b/c there's already an active E. lewisii.

My "solution" here is to make `AtlasesController#alter_atlas_presence` choose to add a listed taxon to a relevant comprehensive list if present, and update any other relevant comprehensive lists that might cause a new listed taxon to fail validations. Another, less invasive solution, might be to allow the addition of inactive taxa to comprehensive lists.

Anyway, open to suggestions. Happy to trash all these changes if you prefer an alternate approach, @loarie, but we need to do something or we can't properly execute https://www.inaturalist.org/taxon_changes/39598


app/models/atlas.rb
```diff
@@ -33,7 +33,7 @@ def places
   # All of the atlas places where there is a ListedTaxon demonstrating presence
   def presence_places
     exploded_place_ids_to_include, exploded_place_ids_to_exclude = get_exploded_place_ids_to_include_and_exclude
-    descendant_listed_taxa = ListedTaxon.joins( list: :check_list_place ).
+    descendant_listed_taxa = ListedTaxon.joins( list: :place ).
```
kueda: I should also point out that this change (and related changes below) might cause serious problems, but they didn't seem to cause any specs to fail.

loarie: Am I understanding this correctly that if there is a comprehensive list for Phrymaceae in Nevada County, and the Atlas editor lists Erythranthe lewisii in Nevada County, this change will only create a listing on the comprehensive list, but not on the default checklist?

kueda: Partially. Currently, if there is a comprehensive list for Phrymaceae in Nevada County, the atlas editor will not let you add Nevada County to the atlas for Erythranthe lewisii. You are correct in saying that this change will only create a listing on the comprehensive list, not the default checklist. But we could change that if you want.

kueda:
Or not? I don't really know how Cassi was able to make an atlas for https://www.inaturalist.org/taxa/777371-Diplacus-aurantiacus unless she went through and edited all the lists... which she might have.

boucur:
Yep--I had briefly activated the sensu stricto D. aurantiacus [as a child of Diplacus (genus) instead of Diplacus (sect)] in order to add it to the comprehensive lists.

app/models/atlas.rb
loarie: my preference would be for the taxon in question (e.g. Erythranthe lewisii) to get added to the default list. All of the atlas and comprehensive stuff to date only deals with the default lists and I think its likely there's unforeseen bugs and complications if they also have to be aware of non-default checklists

PR #1920, opened 2018-09-11
Inactive taxa validations

loarie:
The goal of this is to try to reduce the number of active taxa with ancestries that include inactive nodes by more strongly validating what curators can do. Specifically, this should make it so that:

1) taxon changes won't commit if the inputs have active children unless they are swaps/merges and 'move children' is checked (e.g. curator should manually move children before committing change)

1) active taxa can't be created/updated to be active with inactive parents (unless inactive parent is an output of a draft change since curators may have to manually move children to inactive output taxa before committing as described above)

2) taxa can't be updated to inactive if they have active children

@kueda I've done a pretty rigorous job testing this and playing with it on staging but would appreciate your feedback and any testing

### DIFF
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -930,7 +930,7 @@ class Taxon < ActiveRecord::Base
   def active_parent_if_active
     return if parent.nil? || !is_active
     
-    if !parent.is_active && parent.taxon_changes.where( "committed_on IS NULL" ).first.nil?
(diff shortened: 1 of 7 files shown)


app/models/taxon.rb
```diff
@@ -955,7 +981,7 @@ def complete_taxon
   def graftable_if_complete
     return true unless ancestry_changed?
     ct = complete_taxon
-    if ct && ( current_user.blank? || !ct.taxon_curators.where( user: current_user ).exists? )
+    if !@skip_complete && ct && ( current_user.blank? || !ct.taxon_curators.where( user: current_user ).exists? )
```
kueda: Are you 100% sure this is working? @skip_complete is not the same thing as the `attr_accessor :skip_complete`

PR #1946, opened 2018-10-02
Projects improvements

pleary:
- exclusion filters for taxa, places, users
- annotation filter


kueda:
This is all looking good to me. I say merge and release.

PR #1947, opened 2018-10-02
Specify Ruby version

greysteil:
The Ruby version iNaturalist is tested against (on Travis) is 2.3.1 - it probably makes sense to specify that everywhere, so that everyone is using the same Ruby that is used in production (which I presume is 2.3.1).


pleary:
This looks good to me. We are indeed using ruby 2.3.1 (p112) in production so it makes sense to specify that. OK with me to merge this in

PR #1988, opened 2018-10-11
SemVer compatible updates to a bunch of dependencies

greysteil:
Updates a small group of dependencies at once. All the updates are SemVer compatible, and shouldn't cause any problems.


greysteil:
@kueda - another batch update to help quieten Dependabot down. This will get conflicts if you merge any Dependabot PRs before merging it, so please merge it first!

Issue #1904, opened 2018-09-04
Allow network sites to have different featured projects


carrieseltzer:
Network members want to feature different projects on their sites than what is featured on iNaturalist.org. This outlines changes to the way projects are selected for featuring so that Site Admins and Staff can more easily control which projects are featured at /projects on each network site. 

*Note: all functionality described below should be visible only for Staff and Site Admins.*

How to feature projects:
- [x] For collection and umbrella projects, the About page should have clickable text to “Feature this project”. Clicking reveals a dropdown to select “New & Noteworthy carousel” or “Featured section”. Once a selection is made, the text should show which option was selected where it used to show “Feature this project”.
- [x] For traditional projects, a new section should be added below “Project Curator Tools” called “Site Admin Tools” with the option to “Feature this project”. Clicking reveals a dropdown to select “New & Noteworthy carousel” or “Featured section”. Once a selection is made, the Site Admin Tools section should show which option was selected where it used to show “Feature this project”.
- [x] The option to feature projects should be removed from “edit projects” views
- [x] Project featuring (using the term broadly) should be applied to the platform on which a user was logged in (e.g. if logged in on inaturalist.nz, selections would apply there). 
- [x] Staff should be able to feature projects on iNaturalist.org and any network sites
- [x] Site Admins should only be able to feature projects on their own network site.

What is shown on /projects:
- [x] The project most recently selected as “New & Noteworthy” will become the first in the carousel
- [x] The carousel will have a maximum of three projects
- [x] If more than three projects have been designated for a site as “New & Noteworthy”, the project that received that designation first will be the first one bumped out. Projects bumped from “New & Noteworthy” will appear as the first project shown in the “Featured” section.

Depends on #1897

Issue #1903, opened 2018-09-04
Functionality changes to announcements page & announcements


carrieseltzer:
Depends on #1897 

Changes to the /announcements page:
- [x] When logged in as a Site Admin, /announcements should only show announcements that apply *uniquely to their site*.  
- [x] Site Admins should see the option to edit all announcements that they can see, i.e. all announcements applied uniquely to their site.
- [x] Add ability to search the body of announcements

Changes to creating/editing announcements:
- [x] The default Placement selection for all new announcements should be users/dashboard#sidebar
- [ ] Add ability to select “default” as a locale
- [x] Staff should be able to select multiple sites (rather than either all or a single site)
- [x] Site Admins should only be able to select their site when creating or editing announcements
- [x] Site Admins should be able to edit all announcements applied uniquely to their site

Changes to display prioritization (visibility of announcements for all users):
- [x] Site-specific announcements should take priority over “all site” announcements
- [x] Locale-specific announcements should take priority over “all locale” announcements

kueda:
> Add ability to select “default” as a locale

What's the intent here? If you don't want to target a locale, you just don't choose a locale. Locale just *restricts* the audience of the announcement, so a "default" wouldn't do anything.

carrieseltzer:
It appears to me that users can set their locale to "default", so this piece of the issue is intended to be able to reach those users. 

Yes, Tony and I have discussed targeting multiple sites but not all of them for announcements. This would allow us for example to put the blog post announcements on all sites that are predominantly English while not interfering with those that are predominantly non-English.

kueda:
> It appears to me that users can set their locale to "default", so this piece of the issue is intended to be able to reach those users.

If a user chooses the "default" locale, a locale will still be set, they are just not explicitly choosing it. So if I choose "default" but I use Naturalista, my locale will be es-MX, and I will see both announcements with no locale and announcements with locale set to es-MX. If you use iNaturalist.org, we'll try to infer your locale from information your browser sends us, but if we can't we'll default to English. An announcement targeting *just* people who did not explicitly choose a locale would, in fact, reach no one, so I don't think adding that as an option will help anyone.

carrieseltzer:
Thanks for explaining how the "default" account setting works. I understand now that it is not needed as a locale option for announcements. 

[...]

Issue #1901, opened 2018-09-04
Restrict site editing to appropriate users


carrieseltzer:
Building on #1897, this issue concerns what users can see and do at /sites. 

Site Admins should be able to: 
- View details about other sites by clicking on their icons (no change to current functionality)
- edit information about the site for which they are a designated Site Admin.  (more restrictive than current functionality)

Only Staff should be able to:
- Edit iNaturalist.org (needs to be restricted to staff)
- Edit all sites (needs to be restricted to staff)
- Create new sites (needs to be restricted to staff)

Tasks
- [x] When logged in as a Site Admin, only show the "edit" button for their site
- [x] Ensure that editing endpoints can't be reached directly by other users
- [x] remove the "New Site" button from view for the Site Admins

PR #1990, opened 2018-10-11
New site features

### LINKED ISSUE #1897: Allow iNaturalist staff to assign Site Admin status to a specific network site
When new iNaturalist Network Members join, staff should be able to assign that status from the "curate user" page, and assign individuals to a specific network site. 

- [x] Add a button to "Make Site Admin" which reveals a dropdown listing the current sites.
- [ ] When a user already has site admin status, the button should say "Remove Site Admin Status"

This network site-specific designation should enable additional privileges detailed in subsequent tickets related to editing the site, blog, announcements, and featuring projects *specific to their site*. Additionally, they should have the ability to create and edit wiki pages (which will be viewable across all sites).

pleary:
- Staff can create site admins
- Site admin status featured in user profile
- Site admins can create announcements for their site
- Site admins can set featured projects
- Site admins can see list of all network sites

Closes #1897, #1901, #1903, and #1904


app/views/users/_show.html.haml
```diff
@@ -44,10 +44,20 @@
...
+      - admin_site = @user.site_admins.live.first.try(:site)
+      - if @user.is_admin? || @user.is_curator? || admin_site
+        %small.role
+          - info_link = help_path(anchor: "curators")
+          - if @user.is_admin?
+            = t(:site_staff, site_name: "iNaturalist")
+          - elsif admin_site
+            - site_link = link_to( admin_site.name, admin_site.url, class: "site")
+            = raw t(:site_site_admin, site_name: site_link)
+            = " (#{t("places_name.#{admin_site.place.name.parameterize.underscore}")})"
```
kueda: @pleary not all sites have a place (e.g. the default site), so theoretically if someone was a site-admin but not staff for a site without a place this would raise an error, right? It's doing that for me locally but my local sites are definitely a bit weirder than they are in prod, so maybe this will just never happen.

app/views/users/curation.html.haml
```diff
@@ -10,7 +10,34 @@
...
+      $( "#site-admins .dropdown-menu" ).on( "click", function( e ) {
+        e.stopPropagation( );
+      } );
+      $( "#site-admins input[type='checkbox']" ).on( "change", function( e ) {
+        $.ajax({
+          type: $( this ).is( ":checked" ) ? "POST" : "DELETE",
+          url: "/site_admins",
+          data: {
+            site_admin: {
+              user_id: #{ @display_user.id },
```
kueda: /users/curation is supposed to work without a user_id, so this causes a problem.

app/views/users/_show.html.haml
pleary: Thanks for catching that. I added a conditional to only render the place name if we have one

app/views/users/curation.html.haml
pleary: Thanks - this should be fixed with the latest commit

PR #2141, opened 2019-02-06
Escape $

albert-ziegler:
Hi,

This is a minor fix for an unescaped `$` in a regEx I noticed while browsing this project's site on [LGTM](https://lgtm.com/projects/g/inaturalist/inaturalist/alerts/?mode=list&severity=error) (I work for the company behind that website).

### DIFF
--- a/app/assets/javascripts/atlases/index.js
+++ b/app/assets/javascripts/atlases/index.js
@@ -15,8 +15,7 @@ $(function() {
       },
       error: function(data){
         console.log("error");
-      },
-      dataType: 'JSON'
+      }
     });
   });
 });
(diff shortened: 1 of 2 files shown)


kueda:
It's not an unesecaped `$`, it's a variable in a regex, so if `text` is `"foo (bar)"` and `status.status` is `"bar"`, `text.match( /\(${status.status}\)/ )` should match. The duplicate `dataType` seems legit, though, so if you can remove the `$` change I'll merge this.

albert-ziegler:
:+1: Ah, of course, that makes sense.

I'm not sure it's possible to combine template strings and regex literals though ([example](https://www.w3schools.com/code/tryit.asp?filename=FZY6G2OE3J76)).

Maybe the best way out is to use the RegExp constructor? I've suggested that one.

kueda:
I think you're right about the regexp literal. Oops. However, I think you want
```javascript
if ( !text.match( new RegExp( `(${status.status})` ) )
```
 to match a string like `(status)`, right? No need to escape parens in a normal template string.

albert-ziegler:
Good point, there's no need to escape the brackets.

kueda:
Cool, thanks!

PR #2219, opened 2019-04-05
Trips

loarie:
Trips have been giving 500 errors on show for the last few years. This PR, fixes those bugs, and makes some modifications to how trip data is stored/displayed to accomodate the UBIF collaboration and validating distribution models. TripTaxa are no longer used but haven't been fully removed in the interest of time/going back. Trips now use the existence of observations in the context of Trip parameters and trip_purposes to calculate presence/absence. Trips#tabulate isn't populating the presence/absence column yet - ticket for the ES changes involved in this coming soon

### DIFF
--- a/app/views/trips/index.html.haml
+++ b/app/views/trips/index.html.haml
@@ -1,5 +1,5 @@
 - content_for( :title ) do
-  - @title = t(:trips)
+  - @title = t( :trips )
   = strip_tags( @title )
 .container
   .row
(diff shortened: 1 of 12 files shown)


app/views/trips/index.html.haml
```diff
@@ -1,3 +1,60 @@
...
+  .row
+    .col-xs-12
+      %h2= @title
+      .panel.panel-info
+        .panel-heading
+          %h3.panel-title
+            = "About Trips"
+        .panel-body
+          %p
+            = "Trips describe survey effort by recording where and when an area was surveyed and what was targeted."
```
kueda: FWIW, you could just do
```haml
%p
  Trips describe survey effort by recording where and when an area was surveyed
  and what was targeted. The existence of observations in the context of trips
  can be used to tabulate presence absence data. Presence absence data can be
  useful for understanding samplibg bias in species distribution modeling.
  = link_to "Read more", "/pages/trips"

Or you could globalize it if you think this needs to be translated.

loarie: thanks - also not done internationalizing but will do that after 4/4

PR #2298, opened 2019-06-11
More branding

loarie:
added joint initiative branding to logged out home and log in/sign up


app/views/shared/_header.html.haml
```diff
@@ -120,3 +120,6 @@
             %li= link_to t(:account_settings), edit_person_path(current_user), :class => 'profile_link'
             %li.divider{role: "separator"}
             %li.signout= link_to t(:sign_out), logout_path
+  #initiative.navtabs
+    - if @site && @site.id == Site.default.id
+      = image_tag "https://static.inaturalist.org/wiki_page_attachments/1249-original.png", width: 179
```
kueda: Please check this image in to the repo and don't rely on a wiki page attachment that could change or disappear.

app/views/welcome/_index.html.haml
```diff
@@ -102,6 +102,14 @@
                 = link_to people_path, :class => "morelink explorelink" do
                   = t(:meet)
                   %i.icon-arrow-right
+  #info.homesection
+    .row
+      #ji_left.col-xs-6
+        .pull-right
+          %embed{ src: image_path( "https://static.inaturalist.org/wiki_page_attachments/1248-original.png" ), width: 307 }
```
kueda: Same here, check it in and give it a more descriptive name than 1248-original.png. Also, FWIW, we generally use SVGs on the home page and in the header to deal with varying display resolutions. Not a requirement, but would get in front of issues like the logos looking pixelated.

app/views/shared/_header.html.haml
```diff
@@ -120,3 +120,6 @@
             %li= link_to t(:account_settings), edit_person_path(current_user), :class => 'profile_link'
             %li.divider{role: "separator"}
             %li.signout= link_to t(:sign_out), logout_path
+  #initiative.navtabs
```
kueda: Can you wrap this in something like `if @responsive`? This doesn't need to be on every page on the site, even if its hidden. I'd also appreciate a more descriptive name, like `#supporter-branding` or something.

app/assets/stylesheets/responsive.scss
```diff
@@ -25,4 +25,7 @@
   #usernav {
     display: none;
   }
+  #initiative-branding {
+    display: block;
```
kueda: Can you make this `display: flex`? It's not vertically centered for me as `block`.

PR #2309, opened 2019-06-27
Support implicit flow for OAuth

tomsaleeba:
I'm building a client-side (in browser) only webapp as a client to iNat and want to use OAuth for users to connect to their iNat account.

As the [doco](https://www.inaturalist.org/pages/api+reference#auth) states, iNat currently supports Authorization Code, Resource Owner and some third party flows.

I don't want to deal with passwords so Resource Owner isn't suitable. Authorization Code flow isn't really suited but it could work. The secret wouldn't be secret and there's an extra HTTP call to get the token but it would work. The problem here is iNat doesn't have an OPTIONS method on the `/oauth/token` endpoint so the CORS preflight check fails and we can't POST to get the token.

Implicit Flow is the right flow for this situation. I have been able to enable it with the change in this PR and things work as expected. That is, on the client if we use `...&response_type=token` in our URL that we bounce the user to, we get a token back after redirect.

There should probably be an update to the dev doco to talk about this new flow and something on the OAuth client app page, perhaps a second link, that shows how to use this flow. I'm happy to make those changes, I wanted to make sure I'm on the right path first.

I'm unsure if there are any security implications for this change. DoorKeeper is doing the heavy lifting so I'm thinking it should all be ok.

### DIFF
--- a/config/initializers/doorkeeper.rb
+++ b/config/initializers/doorkeeper.rb
@@ -73,7 +73,7 @@ Doorkeeper.configure do
   #   client.superapp? or resource_owner.admin?
   # end
 
-  grant_flows %w(authorization_code client_credentials password)
+  grant_flows %w(authorization_code client_credentials password implicit)
 
   # Don't force devs to use SSL redirects
   force_ssl_in_redirect_uri false


kueda:
The IETF does not recommend use of the implicit grant flow (https://tools.ietf.org/html/draft-ietf-oauth-security-topics-09#section-2.1.2) so I'm not too keen on allowing that. We could allow cross-origin requests to `/oauth/tokens`, but as you point out, you basically have to expose your client secret in your JS code to make use of that, which is a bit dodgy. From what I can tell, the current best practice is to use PKCE, [which Doorkeeper also supports](https://github.com/doorkeeper-gem/doorkeeper/wiki/Using-PKCE-flow) with a bit of modification. In terms of our Rails code, I think it involves a migration and some config changes. I haven't tried it so I don't know, might involve a new controller or something. It makes your client code somewhat more complicated, b/c now you have to generate an additional "proof" key and include it with your auth code and access token requests.

So long story short, we're not ok with the implicit grant flow, but we would support using Doorkeeper's PKCE flow support if you want to try working on that instead. One major caveat is that it would have to *not* mess with the existing authorization code flow, i.e. PKCE should be possible, but not to the exclusion of the auth code flow.

If you try working on that and it becomes onerous figuring out how we've layered some stuff in front of Doorkeeper, please holler and we can take a look.

tomsaleeba:
Thanks for the quick and comprehensive response @kueda. I wasn't aware that implicit grant flow is not recommended, so I've learnt something there.

I'll explore options for bringing a server-side into the mix so I can do the authorization code flow as it's meant to be. I'm trying to limit the moving parts for this project so hopefully there's a SaaS for this, otherwise a simple AWS Lambda will handle it.

I'll take a look at PKCE too but I think that requires a sizeable amount of effort and introduces more complexity without offering corresponding value.

Keep up the great work with iNat :+1:

PR #2411, opened 2019-09-24
Extract several methods from observation

hedgesky:
This PR extracts 6 methods (`#munge_observed_on_with_chronic`, `#englishize_month_abbrevs_for_locale`, `#get_quality_grade`, `. place_guess_from_latlon`, `#update_stats`, `#dedupe_for_user`) from `Observation` model into dedicated classes, reducing its size from 3296 to 2974 lines.
It also fixes a bug in `observed_on_string` parsing when `Time.zone` value wasn't restored in case of parsing failure.

### Details
I was surprised to find out that iNaturalist is written in Ruby and checked out Github repo to see if I can help with something. Started exploration from Observation model as it's a central entity in the app and found it takes 3296 lines. For me as a newcomer to the project that was a bit overwhelming and I decided to check if it makes sense to extract a bunch of methods from model to dedicated classes. 

I ended up having 5 additional classes almost all of which are essentially just copy-paste from original implementation. However, there's one exception: while working on `ParseObservedOnString` I realized that implementation of `Observation#munge_observed_on_with_chronic` temporary changes `Time.zone`, which leads to two potential issues:

1. This behavior isn't thread-safe, I believe. So there's a chance that incorrect timezone will be used if two users from different locations will try to upload observations (race condition);
2. In original implementation, if an exception occurs during Chronic parsing, original timezone wouldn't be restored.

To outline the issues, make the code more readable and fix issue 2, I decided to not only extract the method to a class, but proceed with further refactoring and slight behavior modification. The behavior change is following:
- Now `Time.zone` is always restored (even if Chronic parsing fails with an error)
- relative observed_on_strings ("today", "yesterday", etc) are always handled (even if Chronic parsing fails with an error). Well, I assume there couldn't be a situation when date_string is relative and Chronic fails, but mentioned it in sake of completeness.

### Note on implementation

On the one hand, introducing the concept of service classes makes the system in general a bit more complex (earlier we have only models, controllers and views; now we also have services). But on the other hand, it brings additional benefits as well:
- It makes model smaller, more readable and easier to reason about;
- In suggested implementation services objects have only one method in public interface: `#call`. Thus, they follow SRP and leverage its benefits: testability, maintainability, reusability.
- Extracted behavior (obviously) relied heavily on `Observation` instance methods. To make things simpler, I used `Forwardable` module from Ruby core to delegate methods from service objects to the model.
- Added an additional test case to check issue 2 from above.


kueda:
@hedgesky, thanks for all this code. Unfortunately, we're not interested in the service classes. They certainly make the model smaller, but they also make it harder to find relevant code, and in my opinion they don't make anything any easier to test. Since we're not interested in that part of this PR, I'm going to close this.

However, we *are* interested in the thread-safe version of `munge_observed_on_with_chronic`, so if you want to refactor your work on that so it doesn't remove code from the model and make a new PR, we'll probably pull that in. If you need to add it to a separate file / class in order to be thread-safe, that's fine.

PR #2414, opened 2019-09-26
speed up Observation and Taxon specs

hedgesky:
This PR speeds up running specs for `Observation` and `Taxon` models. The improvements are two-fold:
1. It disables or stubs calls to Elastic Search where it's not required. With the change running observation and taxon specs is 1.5x as fast on my machine. On CI the build time dropped from ≈28min to ≈17min.
2. Speeds up `load_test_taxa` method. It wraps all creations in a single transaction so every single create/update won't be wrapped into a separate one. Also, for each taxon amount of `save` calls is decreased, thus all the callbacks and validations will be triggered less. 
After the changes, time spent for first call of `load_test_taxa` was reduced from 9s to 4.5s on my machine. The effect is negligible in context of whole suite run, but if you run a single example or a group several times in a row (during refactoring, for example), feedback loop becomes perceivably shorter.

### DIFF
--- a/spec/helpers/make_helpers.rb
+++ b/spec/helpers/make_helpers.rb
@@ -202,31 +202,33 @@ module MakeHelpers
   #           `--- Magnoliopsida
   def load_test_taxa
     Rails.logger.debug "\n\n\n[DEBUG] loading test taxa"
-    @Life = Taxon.find_by_name( "Life" ) || Taxon.make!( name: 'Life', rank: "state of matter" )
-    
-    set_taxon_with_rank_and_parent( "Animalia", Taxon::KINGDOM, @Life, is_iconic: true )
-    set_taxon_with_rank_and_parent( "Chordata", Taxon::PHYLUM, @Animalia )
-    set_taxon_with_rank_and_parent( "Amphibia", Taxon::CLASS, @Chordata, is_iconic: true )
-    set_taxon_with_rank_and_parent( "Anura", Taxon::ORDER, @Amphibia )
-    set_taxon_with_rank_and_parent( "Hylidae", Taxon::FAMILY, @Anura )
-    set_taxon_with_rank_and_parent( "Pseudacris", Taxon::GENUS, @Hylidae )
-    set_taxon_with_rank_and_parent( "Pseudacris regilla", Taxon::SPECIES, @Pseudacris )
+    ActiveRecord::Base.transaction do
+      @Life = Taxon.find_by_name( "Life" ) || Taxon.make!( name: 'Life', rank: "state of matter" )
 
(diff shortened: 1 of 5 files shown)


spec/helpers/make_helpers.rb
```diff
@@ -202,31 +202,33 @@ def make_threatened_taxon( options = {} )
...
+      set_taxon_with_rank_and_parent( "Calypte", Taxon::GENUS, @Trochilidae )
+      set_taxon_with_rank_and_parent( "Calypte anna", Taxon::SPECIES, @Calypte, common_name: "Anna's Hummingbird" )
+
+      set_taxon_with_rank_and_parent( "Plantae", Taxon::KINGDOM, @Life, is_iconic: true )
+      set_taxon_with_rank_and_parent( "Magnoliophyta", Taxon::PHYLUM, @Plantae )
+      set_taxon_with_rank_and_parent( "Magnoliopsida", Taxon::CLASS, @Magnoliophyta )
+      set_taxon_with_rank_and_parent( "Myrtales", Taxon::ORDER, @Magnoliopsida )
+      set_taxon_with_rank_and_parent( "Onagraceae", Taxon::FAMILY, @Myrtales )
+      set_taxon_with_rank_and_parent( "Clarkia", Taxon::GENUS, @Onagraceae )
+      set_taxon_with_rank_and_parent( "Clarkia amoena", Taxon::SPECIES, @Clarkia )
```
hedgesky: Earlier `Taxon::GENUS` rank was used here; after my changes validation error occurred complaining that "rank level must be finer than parent". I believe that `Taxon::SPECIES` is a correct one here so made the corresponding change. Didn't investigate why the validation error didn't show up earlier when parent was set during update, not create.

spec/models/observation_spec.rb
```diff
@@ -1511,49 +1508,53 @@
...
-      o.update_attributes(:taxon => t)
-      Delayed::Job.find_each{|j| j.invoke_job}
-      p.reload
-      expect(p.observations_count).to eq 1
-      Observation.elastic_index!(ids: [ o.id ], delay: true)
-      p.reload
-      expect(p.observations_count).to eq 1
-    end
+    describe "counter caches updating" do
+      with_es_enabled_for_group(Observation, Taxon)
```
hedgesky: Didn't change the specs here, just indented them inside `describe` block to enable ElasticSearch only for these examples.

spec/spec_helper.rb
```diff
@@ -185,36 +187,6 @@ def processing?
   end
 end
 
-# Turn on elastic indexing for certain models. We do this selectively b/c
```
hedgesky: Moved these two methods to `spec/support/es_toggling.rb` to place coupled piece of functionality in a separate place.

Issue #2330, opened 2019-07-11
Changing "private" geoprivacy to "open" of an auto-obscured observation doesn't take


tiwane:
1. Make an observation of an auto-obscured taxon. The example from the user was _Libellula vibrans_ in Ohio.

2. Before uploading, set geoprivacy to private.

3. After upload, edit observation's geoprivacy to open.

4. Map still shows private for other users, but under "Details" it says "open" for geoprivacy.

pleary:
I have an open PR for this https://github.com/inaturalist/inaturalist/pull/2424. Would be great to have someone double check the code to make sure it's only fixing this bug and not causing other problems

pleary:
This should be fixed now after this PR was merged

PR #2424, opened 2019-10-04
Potential fix for #2330

pleary:
Tracing through the error, I found that in this case the observation started with a private geoprivacy, so its lat/lng were nil, and then when setting the geoprivacy to open, with the taxon still being protected in the observation area, it would get to this point in `obscure_coordinates` with its lat/lng having been re-assigned in code to patch the private coords, but the `latitude_was` and `longitude_was` were nil. That leaves the obs in a state where geoprivacy is open, but the public coordinates are nil instead of being obscured.

I made a spec for that case, and all specs are passing, but it would be great to get a second set of eyes on this in case there are some consequences of this change I'm not seeing.

### DIFF
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -1546,7 +1546,7 @@ class Observation < ActiveRecord::Base
     if (
       ( latitude == private_latitude || longitude == private_longitude ) &&
       !( private_latitude_changed? || private_longitude_changed? ) &&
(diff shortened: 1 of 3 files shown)


kueda:
I took a quick look at this and I can't think of any obvious problems. This stuff is pretty thoroughly tested so if specs are passing I feel good about merging it in.

PR #2554, opened 2020-01-11
default to RG photos

jwidness:
I *think* this will cause the taxon photo chooser to default to RG observations instead of all observations.  The idea is to help prevent incorrect taxon photos.

image

### DIFF
--- a/app/webpack/taxa/show/components/photo_chooser_modal.jsx
+++ b/app/webpack/taxa/show/components/photo_chooser_modal.jsx
@@ -316,8 +316,8 @@ class PhotoChooserModal extends React.Component {
                       className="form-control"
                       onChange={e => this.setProvider( e.target.value )}
                     >
-                      <option value="inat">{ I18n.t( "observations" ) }</option>
                       <option value="inat-rg">{ I18n.t( "rg_observations" ) }</option>
+                      <option value="inat">{ I18n.t( "observations" ) }</option>
                       <option value="flickr">Flickr</option>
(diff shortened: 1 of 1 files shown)


kueda:
Thanks, but please see https://github.com/inaturalist/inaturalist/blob/master/CONTRIBUTING.md. Feature requests should get worked out in the Forum first, and if there's support someone (probably someone on staff) should make a github issues, and then labeling of the issue indicates that it should get done.

kueda:
Just realized CONTRIBUTING was not entirely clear on this point, so I just updated it.

jwidness:
We already discussed it repeatedly in the forum and @tiwane said he would get to it.  Seemed like small enough a change to shortcut the red-tape, but I guess I can start back at the beginning of the process.

kueda:
Ok, #2596 exists now so I'm re-opening this. Still need to confirm that this actually fixes the issue. @jwidness, do you have a working development environment in which this fixed the issue?

jwidness:
No, I currently only have access to a Windows environment and didn't feel like attempting to get things running in Windows.

kueda:
Roger that. Just gave it a test and it didn't quite work, but I'm pretty sure 4af4782ac67ec8498aa132752df80c753de52a17 will do the trick.

jwidness:
Thanks!  Generally how long is it between a commit and going live on the site?

kueda:
Between hours and days. This one will be more like hours. Just need to make sure all tests pass.

Issue #2673, opened 2020-05-14
Collection projects that have a date requirement in the future lack a flag link


tiwane:
From https://forum.inaturalist.org/t/unable-to-flag-unstarted-bioblitz-project-as-spam/7217

Some spammers have created projects which can't be easily flagged due to this.

1. Create a collection project with an observed-on date that is in the future, which has a date range where the starting date is in the future.
2. There is no option to flag the project at the bottom of the project's page.
3. Change the date to something in the past.
4. Flag link appears.

todtb:
I'll take a look at this. I just got this set up locally and this should be a pretty quick fix.

kueda:
Holler if you run into any blockers. This will require some React work.

PR #2684, opened 2020-05-28
Add flag link to project before_event_tab component

todtb:
#2673
Add flag link to project before_event_tab component

New around here, but I didn't see anything like `mocha` or `jest` so just from some local click testing, this seems to operate fine.


kueda:
Thanks again. Yeah, we haven't broached tests for React components yet.

Issue #2677, opened 2020-05-24
JS SyntaxError on obs field merge page


tomsaleeba:
**Describe the bug**
When the merge page (https://www.inaturalist.org/observation_fields/*ID*/merge) for an observation field is loaded, a JS syntax error is thrown.

**To Reproduce**
1. view the list of obs fields https://www.inaturalist.org/observation_fields
1. find one that you can edit
1. edit it
1. click the "merge" button

**Expected behavior**
The page loads without a JS SyntaxError.

**Actual behavior**
Looking in the browser dev tools shows an error in the console:
```
Uncaught SyntaxError: Unexpected token '&'
```
It refers to this line
```js
    <script>
  var ATTRS = [&quot;name&quot;, &quot;datatype&quot;, &quot;description&quot;, &quot;allowed_values&quot;, &quot;observations_count&quot;, &quot;projects_count&quot;] // <- this line
  $(document).ready(function() {
```
...which I think comes from here in the source: https://github.com/inaturalist/inaturalist/blob/44e8fce/app/views/observation_fields/merge.html.haml#L7

The values should be literally quoted, not HTML escaped.

**Context**
 - OS: Manjaro
 - Browser Chrome
 - Version 81

PR #2685, opened 2020-05-28
Fix js error on observation field merge page

todtb:
#2677

I think this will fix the js error and restore the `jquery.chooser` functionality. However since that introduces a functional change to the page, let me know what you think and if we want to tighten it up any.


kueda:
Looks good. Thanks!

Issue #2670, opened 2020-05-12
Only allow project users to be made project managers


kueda:
In collection project lingo, only allow project members to be made project admins. Currently some unpleasant shenanigans are possible. Pretty sure this requires a change to

- [x] `Project#add_admins`
- [x] The admin autocomplete on the projects form so it only shows project members, not any user
- [x] Project form UI text to make it clear you can only make people admins after they have joined the project

todtb:
I can take this on.

todtb:
PR is up as a draft until the API endpoint is merged here: https://github.com/inaturalist/iNaturalistAPI/pull/195

PR #2687, opened 2020-05-29
Add autocomplete users to project edit scoped by membership

todtb:
#2670 
~Do Not Merge~

~I've opened this as a draft as we will need the endpoint merged here: https://github.com/inaturalist/iNaturalistAPI/pull/195~

Also this adds text to the Project edit page which will need i18n.

Other than that it's working great with the ES scoped to users already in the project on the API end and the controller rejecting invalid options to be safe.

Note: I've left the ability to add admins on the Project new page as there are no members yet and this could save some time and page loads.


todtb:
@kueda @pleary Updated to match changes to https://github.com/inaturalist/iNaturalistAPI/pull/195

I suppose the API prerequisite is no longer applicable, but I still think it makes sense to get to that first.

todtb:
https://github.com/inaturalist/iNaturalistAPI/pull/198 should allow this to go forward until the index is added, unless the 1000 member cap would prevent doing so.

app/assets/javascripts/i18n/translations.js
```diff
@@ -17133,6 +17133,7 @@ I18n.translations["en"] = {
     "projects": {
       "collection": "Collection",
       "edit": {
+        "admins_must_be_existing_members": "Admins must be existing members of the project.",
```
kueda, line 17136: This needs to be specified in https://github.com/inaturalist/inaturalist/blob/master/config/locales/en.yml#L5675. This js file will get overwritten with every deploy.

app/models/project.rb
```diff
@@ -1038,16 +1038,16 @@ def flagged_with( flag, options = {} )
...
-        if pu = project_users.where( user_id: admin_attr["user_id"] ).first
-          pu.update_attributes( role: nil )
-        end
+      new_role = admin_attr["_destroy"] == "true" ? nil : "manager" 
+      if new
+        project_users.find_or_create_by( user_id: admin_attr["user_id"] ).update_attributes( role: new_role )
       else
-        pu = project_users.find_or_create_by( user_id: admin_attr["user_id"] )
-        pu.update_attributes( role: "manager" )
+        project_users.find_by( user_id: admin_attr["user_id"] )&.update_attributes( role: new_role )
```
kueda, line 1050: Neat, I didn't know about the `&.` operator.

app/assets/javascripts/i18n/translations.js
todtb: Ahh, I see `generate_translations_js` rake task. Thanks for the info. I guess changing this output is irrelevant and unnecessary.

kueda:
Looking good. Thanks again!

Issue #2623, opened 2020-03-24
Prevent creation of observations with more than 20 photos in the Uploader


kueda:
In the Uploader, make it so a photo or a card will not drop on a card that already has >= 20 photos.  In a perfect world, we would show some kind of error or popup that says, "Observations can only have 20 photos. To add more you will need to remove some" when the user tries to do this, but I'm not sure we can dispatch actions in response checking droppability.

todtb:
I'll take a look at this.

We also want to prevent "combining" cards (ex: "Select All" > "Combine") into a stack with over 20 photos too, right? Is that the same as dragging a card on top of another?

PR #2689, opened 2020-05-30
Limit photos uploaded or combined in upload to observations

todtb:
#2623 

This should handle file limits for drag and drop, combine cards, and move a single photo from one card to another.

Let me know if I've missed any other ways these might exceed the limit.

### DIFF
--- a/app/webpack/observations/uploader/actions/actions.js
+++ b/app/webpack/observations/uploader/actions/actions.js
@@ -29,6 +29,13 @@ const actions = class actions {
     };
   }
 
+  static fileCounts( files ) {
+    return {
+      photos: _.filter( files, f => /^image\//.test( f.type ) ).length,
+      sounds: _.filter( files, f => /^audio\//.test( f.type ) ).length
+    };
+  }
+
   static enforceLimit( src, dest ) {
     const cardFileDropLimit = 20;
     const obeysLimit = ( src + dest <= cardFileDropLimit );
@@ -43,7 +50,7 @@ const actions = class actions {
           }
         } ) );
       }
-      return ( obeysLimit );
+      return obeysLimit;
     };
   }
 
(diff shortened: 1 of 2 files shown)


app/webpack/observations/uploader/actions/actions.js
```diff
@@ -29,6 +29,24 @@ const actions = class actions {
...
+        dispatch( actions.setState( {
+          confirmModal: {
+            show: true,
+            message: I18n.t( "observations_can_only_have_n_photos", { limit: cardFileDropLimit } ),
+            confirmText: I18n.t( "ok" ),
+            hideCancel: true
+          }
+        } ) );
+      }
+      return ( obeysLimit );
```
kueda, line 46: Can you remove the parens here?

app/webpack/observations/uploader/actions/actions.js
```diff
@@ -149,7 +167,9 @@ const actions = class actions {
 
   static onFileDropOnCard( droppedFiles, obsCard ) {
     return function ( dispatch ) {
+      const cardFileCount = Object.keys( obsCard.files ).length;
```
kueda, line 170: Here and below at 242 I think you need to make this check specific to photo files. Currently, if you have a card with 20 photos and you try to drag a sound on to it, it will give you the error message about only allowing 20 photos, which is confusing if you're adding a sound.

todtb: Makes sense. Are you thinking 20 total files (photos and sounds) or 20 of each?

kueda: 20 of each.

app/webpack/observations/uploader/actions/actions.js
```diff
@@ -357,6 +397,10 @@ const actions = class actions {
 
   static movePhoto( photo, toObsCard ) {
     return function ( dispatch ) {
+      const { photos: targetPhotos, sounds: targetSounds } = actions.fileCounts( toObsCard.files );
+      const { photos: movedPhotos, sounds: movedSounds } = actions.fileCounts( [photo.file.file] );
```
kueda, line 401: I think this should be `[photo.file]`. Otherwise I get a bug when I add a sound, duplicate it, then try to drag the sound (not the card) onto the original card.

todtb: Huh, nice catch. Just out of curiosity, did you find this by click testing different edge cases, or some other means?

kueda: Just click testing. As I mentioned, we don't have any automated testing of React stuff.

todtb: All good. I think a lot is just my lack of familiarity with all the cases in these features. I've got nowhere to go but up on that front!

kueda:
Thanks, @todtb!

PR #2693, opened 2020-06-05
Allow taxon photos to be locked

### LINKED ISSUE #2396: Allow a taxon's photos to be locked
Anyone can edit the photos for a taxon, and that is usually a good thing, but it can lead to several issues:

- edit wars, where maybe two or more people are constantly changing the photo back and forth, without dialogue https://www.inaturalist.org/flags/406465 I'll note that @bobby23 [proposed](https://forum.inaturalist.org/t/track-and-view-changes-to-taxon-photos/5630) a tracking system for taxon photos changes, but I think that might be better implemented with a  larger tracking system that includes geoprivacy changes and such. 

- the default photo for _Homo sapiens_. Because we all know what humans look like, a diagnostic photo isn't necessary. However, since we're a global site, any photo of a human that is chosen as the standard depiction of our species can't be representative of all humans and thus all users, which causes user to change the default photo often. I suppose this in and of itself isn't a huge problem, but it seems like it will be a constant thing and, to me, a waste of users' time, and could be the cause of bad blood and hurt feelings. 

Other potential issues with our current system would be users posting obscene, offensive, or willfully incorrect photos.

So I propose that we have a functionality that allows a taxon's photos to be locked and only staff can edit them, should such situations arise.

kueda:
Closes #2396

### DIFF
--- a/db/migrate/20200604181750_add_photos_locked_to_taxa.rb
+++ b/db/migrate/20200604181750_add_photos_locked_to_taxa.rb
@@ -1,0 +1,5 @@
+class AddPhotosLockedToTaxa < ActiveRecord::Migration
+  def change
+    add_column :taxa, :photos_locked, :boolean, default: false
+  end
+end
--- a/app/assets/javascripts/i18n/translations.js
+++ b/app/assets/javascripts/i18n/translations.js
@@ -4176,6 +4176,7 @@ I18n.translations["ca"] = {
   "are_you_sure_remove_photo": "Esteu segur que voleu eliminar aquesta foto?",
   "are_you_sure_you_want_to_remove_all_tags": "Esteu segur que voleu eliminar totes les etiquetes?",
   "are_you_sure_you_want_to_remove_these_x_taxa?": "Esteu segur que voleu eliminar aquests %{x} tàxons?",
+  "are_you_sure_you_want_to_remove_this_observation_from_project": "Esteu segur que voleu eliminar aquesta observació de %{project}?\n",
   "asc": "asc",
   "ascending": "ascendent",
   "atom": "Atom",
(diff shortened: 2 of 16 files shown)


db/migrate/20200604181750_add_photos_locked_to_taxa.rb
```diff
@@ -0,0 +1,5 @@
+class AddPhotosLockedToTaxa < ActiveRecord::Migration
+  def change
+    add_column :taxa, :photos_locked, :boolean, default: false
```
todtb, line 3: With a default set, could this also be `null: false`?

kueda: Yup. Might be wise. This is also going to be a rarely-used column, though, and in most situations, a `null` value will evaluate as `false` anyway.

todtb:
Though with the current logic in the controller, this isn't strictly necessary, it would provide some defense since Taxa `accepts_nested_attributes_for :taxon_photos`.

What about a validation something like:

```ruby
def cannot_edit_locked_photos
  return unless taxon_photos_changed?
  return unless photos_locked? 
  return if current_user.is_admin?
  errors.add(:taxon_photos, t('Message'))
end
```

(I didn't test any of this, so it may be off, just a thought)

kueda:
Re: the `cannot_edit_locked_photos ` validation, I thought about a model-level validation, but ultimately decided against it because, on the whole, models aren't aware of the `current_user`, and I'm not sure they *should* be. Generally I feel like the `current_user` is a part of the session, and that's the controller's responsibility. Maybe if we stored the person who made the last edit in the db (e.g. an `updater` relationship or something) then it would be the model's concern, but that seemed like overkill for this minor feature.

We've certainly done things like `attr_accessor :current_user` in the Taxon model and elsewhere, but I've never felt great about it.

todtb:
True, I saw `current_user` in there so thought it was possible, though, yeah, it is unconventional. Probably best without the model validation.

PR #2698, opened 2020-06-12
Preliminary integration for API V2

kueda:
Pending https://github.com/inaturalist/iNaturalistAPI/pull/196, and should be updated if/when we merge https://github.com/inaturalist/inaturalistjs/pull/31

Basically uses the alpha v2 API on obs detail when specified by URL param or by an opt-in test. Parity isn't complete yet, but the basics are there.


pleary:
This all looks good to me. I fixed two small things
 - the identifiers weren't loading properly on obs show
 - there was an undefined constant in `webpack/observations/show/ducks/controlled_terms.js` which I presume was legacy from early in the implementation. The `fetchAllControlledTerms` method that contained it appears to only be used on Identify and ProjectCreate, so not yet using the V2 api anyway

Issue #2706, opened 2020-06-22
Extract metadata from PNGs


kueda:
Apparently we're only doing it from JPGs. Not sure what the best lib to do this is, but maybe https://github.com/wvanbergen/chunky_png. Should try to extract all the same fields we get from JPGs with exifr and xmp.

todtb:
I'm going to try to give this a go with `chunky_png`.

todtb:
TL;DR: after some experimentation it looks like the ruby wrapper for exiftool may prove to be more useful here. I'm going to carry on using that.

---

After some experimentation it looks like `chunky_png` is using metadata stored in `zTXt`. See http://www.libpng.org/pub/png/spec/1.2/PNG-Chunks.html#C.zTXt 

This was confirmed first by converting a `.jpg` file with known exif data using `magick`(ImageMagick "preserves" exif data by shoveling into `zTXt`) and observing the fields with both [`exiftool`](https://exiftool.org/) and `chunky_png` that later of which yields:
```ruby
"exif:GPSLatitude" => "39/1, 55/1, 5801/100",
"exif:GPSLatitudeRef" => "N",
"exif:GPSLongitude" => "75/1, 12/1, 4317/100",
```
It essentially namespaces the keys in zTXt under "exif:*"

`exiftool` shows both sets of tags (zTXt top, exif bottom):

```bash
Exif GPS Latitude               : 39/1, 55/1, 5801/100
Exif GPS Latitude Ref           : N
Exif GPS Longitude              : 75/1, 12/1, 4317/100
Exif GPS Longitude Ref          : W
...
GPS Latitude                    : 39 deg 55' 58.01" N
GPS Longitude                   : 75 deg 12' 43.17" W
````

Using `exiftool` to manually add the tags to a `.png` file yields no results from `chunky`.

Luckily there is a ruby wrapper for `exiftool`: https://github.com/exiftool-rb/exiftool.rb which is showing promising results along with some useful casting of data types: 

```ruby
:gps_date_time => 2020-06-07 19:12:56 UTC,
:gps_date_time_civil => #<Date: 2020-06-07 ((2459008j,0s,0n),+0s,2299161j)>,
:gps_latitude => 39.93278056,
:gps_longitude => -75.21199167,
```

kueda:
Cool, exiftool will work for us.

todtb:
[...]

PR #2716, opened 2020-07-03
2706 extract png metadata

todtb:
#2706 

This allows extraction of exif (and common `tTXt` and `zTXt` metadata) from `.png` files.

The extraction of this data allows not only for fields such as `Copyright`, `Comment`, etc on the photo's view page, but also full exif feature functionality currently available to JPEGs (timestamp, location, taxon, and comment autofill on Observation uploader, metadata serialization)

The `exiftool_vendored` gem should come with the required `exiftool` however if this tool is missing, that will raise a specific exception (`Exiftool::ExiftoolNotInstalled`) that will be handled. Something to keep an eye on when deploying to different environments. That, along with other exceptions raised by exif metadata extraction are currently rescued and logged.

The tags of interest extracted from the `tTXt` and `zTXt` chunks are in `PNG_CHUNK_MAP`, many `.png` files may not be utilizing the relatively new support for exif data, so I wanted to make sure to grab this info from both places with exif taking precedence. 

Consistency in the serialized metadata was a goal here and therefore exif data is cast/serialized/etc to the existing `EXIFR` implementations. Additionally, only tags brought in by `EXIFR` are preserved (though the full supported list can be found here https://exiftool.org/TagNames/EXIF.html)

Manual testing was done with a variety of sample images either downloaded, generated, or captured on a device. That said, there may be an odd format here or there not accounted for. This should be handled/rescued and a baseline of metadata retained (as the whole process during and upload iterates and is additive in nature).

Also, while the data models aren't currently set up to do so, similar data for `.wav` files could in theory be extracted as well, though overall support/use of this by other libraries and devices seems spotty.  

Lastly, a note on `ComputerVisionDemoUpload`. I've updated it's metadata extraction as well, but it does not appear that it is currently using that method (relying instead on front-end exif libraries) and thus does not store any serialized metadata in the database.


kueda:
Will take a look at this next week. FWIW, we have exiftool installed on production machines so we probably don't technically need `exiftool_vendored`, but that probably does make setup a lot easier (one less external dep).

app/models/computer_vision_demo_upload.rb
```diff
@@ -33,22 +33,8 @@ def extract_metadata( path )
...
-              unless xmp.dc.send(dcattr).blank?
-                self.metadata[:dc][dcattr.to_sym] = xmp.dc.send( dcattr )
-              end
-            rescue ArgumentError
-              # XMP does this for some DC attributes, not sure why
-            end
-          end
-        end
-      end
+      exif_data = ExifMetadata.new(path: path, type: photo_content_type ).extract
```
kueda, line 36: FYI, our general Ruby style is to include a space within parens, so this should be `ExifMetadata.new( path: path, type: photo_content_type )`

spec/lib/exif_metadata_spec.rb
```diff
@@ -0,0 +1,77 @@
+require File.dirname(__FILE__) + '/../spec_helper.rb'
+
+describe ExifMetadata, 'extract' do
```
kueda, line 3: Another style issue: we use double quotes pretty much across the board.

todtb: Have you ever considered adding in a structured linter/style guide such as a `rubocop.yml`? Even if it wasn't wired up to run locally or in CI/CD or even make any changes, it could serve as reference for the style guide and be incorporated into many editors that support it and help all new code conform to the style specified.

pleary:
> 1. Do we want to include a vendored version of exiftool or assume what ever machine the app is running on has it (true of our current production machines)?

I don't mind including the `exiftool_vendored` gem. If we see there are compatibility issues across platforms, or if the gem doesn't update fast enough to address security vulnerabilities that pop up we can reconsider and fall back to the linux distribution

> 2. This is currently only using tags from `exiftool` that match the tags we get from `EXIFR`, which is good for consistency and compatibility, but bad b/c we're discarding a lot of potentially useful metadata (but also good b/c we're not wasting space on a bunch of stuff we'll probably never use). Keep it consistent or try to capture everything?

The photos table is by far our largest table, and part of that is all the metadata we store. I'd rather not start collecting more metadata if we're not going to use it. Or if you do want to keep it so it doesn't get lost, I'd really like to see a proposal for how we can store that outside postgresql

spec/lib/exif_metadata_spec.rb
kueda: I'll think about it. We use eslint for JS and while it does encode our opinions about style, I also feel like I spend a lot of time making non-functional changes to enforce style compliance, so I'm not sure it's a net gain.

todtb: Oh for sure. Definitely wouldn't want to start enforcing the guide until there's some major time set aside and even then it would probably need to be phased. More as an unenforced documentation of style guide than anything. Is it documented elsewhere that I may have missed?

kueda:
> I'd rather not start collecting more metadata if we're not going to use it

Theoretically we'd only be storing more metadata for PNG uploads, which constituted 0.38% of all photos uploaded in June 2020 (23,612 total), so I don't think that's too much of a burden to bear, right? I'm more concerned about consistent behavior, i.e. if we allow people to upload PNGs, we should treat the metadata the same as we treat JPGs. Fair argument for keeping it minimal though.

pleary:
Maybe I misunderstood. I was thinking this PR causes us to get metadata for PNGs, which we hadn't done before, and the metadata we get from PNGs would be the same as we have been getting from JPEGs, and that we wouldn't change the metadata we're getting from JPEGs. I thought the question was - we can get more data from exiftool than we had been getting, and we could get it from both PNGs and JPEGs, and that this would be new stuff we hadn't collected before. To that I was suggesting if we hadn't been collecting it before maybe we don't need it.

I agree we should treat PNGs and JPEGs the same

todtb:
So some clarification to the remarks that opened this PR @pleary . 

The support we have for PNG and JPEG is different. For most of its history the PNG standard did not support exif data. Instead it supported [a limited number](https://dev.exiv2.org/projects/exiv2/wiki/The_Metadata_in_PNG_files#2-Textual-information-chunks-the-metadata-in-PNG) of metadata tags. It now supports exif data, so we want to pull that out too. However `EXIFR` does not actually pull out [all supported exif data](https://exiftool.org/TagNames/EXIF.html), just a [subset](https://github.com/remvee/exifr/blob/master/lib/exifr/tiff.rb#L46). 

Therefore, any exif data pulled from PNGs will only be the subset we are pulling from JPEGs. Additionally we want to pull in the `tTXt` and `zTXt` metadata from PNGs and map them to the corresponding exif tags for serialization purposes. We pull `tTXt` and `zTXt` from PNGs as many may not be using the relatively newly support exif.

PR #2719, opened 2020-07-08
Attempts to improve og tags on obs search

kueda:
This attempts to improve the Open Graph tags we include at https://www.inaturalist.org/observations with some customization based on the search parameters. That requires loading some data in Rails, which might be too much of a performance burden to bear, hence the PR for review. If we don't want to do that, we can show the generic site description as the `og:description` instead.


pleary:
this all looks good and is working well for me locally. I'm not worried about the data that needs to be fetched to generate these

Issue #2722, opened 2020-07-08
Support keyboard shortcuts in the TextEditor


kueda:
Specifically

- [x] `CMD-b` should be the same as clicking the "bold" button
- [x] `CMD-i` should be the same as clicking the "italic" button
- [x] `CMD-k` should be the same as clicking the "link" button (pilfered from Github, but so useful)

Using the `CTRL` versions should work in Windows, and it looks like [mousetrap](https://www.npmjs.com/package/mousetrap) supports a `mod` helper to do that, or we can just use `"ctrl-b", "cmd-b"`.

todtb:
I'll work on this in tandem (as separate branches, of course) so I can better understand and make sure the work done on #2723 lines up with the other upcoming improvements.

PR #2725, opened 2020-07-09
Add mousetrap bindings for bold, italics, and links in TextEditor

todtb:
#2722 

This adds mousetrap bindings for bold, italics, and links in the `TextEditor` component. The use of refs required converting `TextEditorFormatButton` from a functional component. 

Additionally, this can't really be handled in the `keyboard_shortcuts.js` file as we need to bind to an element rendered by the component rather than the default `document`. 

I'm not 100% certain this is the cleanest approach, but it seems to click-test well. Open to any suggestions to make it more idiomatic.

### DIFF
--- a/app/webpack/shared/components/text_editor.jsx
+++ b/app/webpack/shared/components/text_editor.jsx
@@ -1,5 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
+import mousetrap from "mousetrap";
 import TextEditorFormatButton from "./text_editor_format_button";
 import UserText from "./user_text";
 
@@ -17,9 +18,18 @@ class TextEditor extends React.Component {
   }
 
   componentDidMount( ) {
-    Mousetrap( this.textarea.current ).bind( 'mod+b', ( ) => { this.boldButton.button.current.click( ); } );
-    Mousetrap( this.textarea.current ).bind( 'mod+i', ( ) => { this.italicButton.button.current.click( ); } );
-    Mousetrap( this.textarea.current ).bind( 'mod+k', ( ) => { this.linkButton.button.current.click( ); } );
+    mousetrap( this.textarea.current ).bind( "mod+b", e => {
+      e.preventDefault();
+      this.boldButton.button.current.click( );
+    } );
+    mousetrap( this.textarea.current ).bind( "mod+i", e => {
+      e.preventDefault();
+      this.italicButton.button.current.click( );
+    } );
+    mousetrap( this.textarea.current ).bind( "mod+k", e => {
+      e.preventDefault();
+      this.linkButton.button.current.click( );
+    } );
   }
 
   render( ) {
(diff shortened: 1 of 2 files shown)


app/webpack/shared/components/text_editor.jsx
```diff
@@ -7,12 +7,21 @@ class TextEditor extends React.Component {
...
+    this.italicButton = React.createRef();
+    this.linkButton = React.createRef();
     this.state = {
       textareaChars: 0,
       preview: false
     };
   }
 
+  componentDidMount( ) {
+    Mousetrap( this.textarea.current ).bind( 'mod+b', ( ) => { this.boldButton.button.current.click( ); } );
```
kueda, line 20: While this is working, Mousetrap should be imported or declared as a global in an eslint comment. Also, double-quotes only, please. React-specific eslint rules are at https://github.com/inaturalist/inaturalist/blob/master/app/webpack/.eslintrc

app/webpack/shared/components/text_editor.jsx
```diff
@@ -43,6 +52,7 @@ class TextEditor extends React.Component {
                 label={<i className="fa fa-bold" />}
                 template={text => `**${text}**`}
                 placeholder={I18n.t( "bold_text" )}
+                ref={(button) => {this.boldButton = button;}}
```
kueda, line 55: Parens around `button` are extraneous and curly braces around the function definition need spaces on the inside (also in the eslint rules).

app/webpack/shared/components/text_editor_format_button.jsx
```diff
@@ -1,63 +1,74 @@
...
+      <button
+        type="button"
+        tabIndex="-1"
+        className={className}
+        disabled={disabled}
+        title={tip}
+        aria-label={tip}
+        ref={this.button}
+        onClick={() => {
+          const {selectionStart} = textarea;
```
kueda, line 32: Spaces inside the braces, please.

app/webpack/shared/components/text_editor_format_button.jsx
```diff
@@ -1,63 +1,74 @@
...
+          if (textarea.selectionStart !== undefined && textarea.selectionEnd !== undefined) {
```
kueda, line 33: Spaces inside parens.

app/webpack/shared/components/text_editor_format_button.jsx
```diff
@@ -1,63 +1,74 @@
...
-    }}
-  >
-    { label }
-  </button>
-);
+class TextEditorFormatButton extends React.Component {
+  constructor(props, context) {
+    super(props, context);
+    this.button = React.createRef();
+  };
```
kueda, line 8: Class methods don't need semi-colons.

todtb:
I tested `preventDefualt( )` on 78.0.2 (64-bit) and that should do the trick. I think that's acceptable in this case given that it is only bound to the textarea.

Not sure if we need to support IE, but it (of course) uses a slightly different approach I've left out for now.

kueda:
Looking great. Thanks again!

Issue #2723, opened 2020-07-08
Use the TextEditor in Identify


kueda:
Use the [`TextEditor`](https://github.com/inaturalist/inaturalist/blob/master/app/webpack/shared/components/text_editor.jsx) For comments and IDs like we do on obs detail.

todtb:
I can take a look at this if you didn't already have a plan of action @kueda. Do you think it makes more sense to get this in "Identify" first or approach #2721, #2722, and #2724 first?

kueda:
They all kind of complicate each other equally, so if you want to take a crack at this, go for it. Much obliged.

PR #2727, opened 2020-07-10
Use TextEditor in observations identify

todtb:
#2723 

~It looks like the older `INatTextArea` supported the "mentions" boolean, but it didn't appear to be utilized in `CommentForm` or `IdentificationForm`.~

I was thinking it would be neat to bring in the whole side-by-side comment/id container from observations show, but it might be too much of a change in the UI too quickly. Additionally, it would add a non-trivial amount of work with redux and refactoring to make it work both in observations/show and /observations/identify (observations modal). 

From a UI side, I just kinda kept the overall style of the /observations/identify page, but do note the preview functionality (bg-color is different from /observations/show, but consistent for the page).


app/webpack/observations/identify/components/observation_modal.jsx
```diff
@@ -36,6 +36,11 @@ class ObservationModal extends React.Component {
     };
   }
 
+  componentDidMount( ) {
+    const domNode = ReactDOM.findDOMNode( this );
+    $( ".IdentificationForm textarea,.CommentForm textarea", domNode ).textcompleteUsers( );
```
kueda, line 41: This isn't working for me, I suspect b/c the textareas this is selecting don't exist when it runs. Might be better to add a `mentions` option to `TextEditor` like we were doing in INatTextArea and do this when *that* component mounts.

todtb:
> This looking mostly great. In addition to the mentions issue, the `c` keyboard shortcut is focusing on the bold button, when it should focus on the comment form `<textarea>`.

Oh good catch. Just saw the change you made to `observations/show/keyboard_shortcuts.js:12` recently. I'll add that in for `observations/identify/keyboard_shortcuts.js` as well now that we're using the new component here too. Also the change to the TextEditor to support "mentions" bool.

kueda:
Thanks again!

Issue #2724, opened 2020-07-08
Inline editing of comments and ID bodies using the TextEditor


kueda:
Currently to edit these things the ActivityItems loads an old Rails edit form in a new window/tab. Performing this action inline would be an improvement, IMO, and allow us to use the TextEditor for editing in addition to creating.

- [ ] When you click "Edit" in the ActivityItem menu, replace the UserText with a form that includes a TextEditor populated with the `body` of the comment or ID. `href` should remain the same, but the click event should show the form and stop propagation.
- [ ] Editing form should have "Save" and "Cancel" buttons
- [ ] If we're making it so you can't get to the edit identification form we need to support identification deletion in the context menu. We don't *really* want people to do that, though, so that should be at the bottom, in gray, and maybe fire an "Are you sure?" confirm dialog

todtb:
I can take a look at this next if you weren't already planning to. When you say stop propagation, you mean in react (`stopPropagation( )`) to prevent the click event from doing anything else, right? I'm not very familiar with it, but if it's just a matter of calling it, I can throw that in before showing the TextEditor.

kueda:
Go for it. I was just recommending that the "Edit" menu item continue to be a link identifications/edit or comments/edit but that if all the JS is working normally, it should open the inline editor instead of opening the link.

PR #2733, opened 2020-07-13
2724 inline edit activity item

todtb:
#2724 

I'm opening this as a draft because it's so very close to functionally complete but with one perplexing issue. 

What I'm seeing is when using the hard delete functionality on an Identification (previously only on the id edit page), there seems to be some unpredictable results perhaps from a race condition.

From the observation.js duck, `deleteID` is called. It sends off `dispatch( callAPI( inatjs.identifications.delete, Object.assign( {}, { id }, options ) ) );` Now `callAPI` will await the promise, however it appears that the response is returned BEFORE the backend has truly deleted the identification (sometimes). The next thing that happens is that `setObservation()` will go out and fetch the current (perhaps sometimes deleting and sometimes deleted) information and apply that to the (front end) state. Thus, sometimes this includes the the record marked `api_status: "deleting"`. Stepping through the redux state diffs in the browser I can see it deleting and "restoring" it from the latest pull for identifications in the same step:

```
0(pin):{id:251,uuid:'436e08bf-6cfd-4f…ent:false} (ADDED)
2(pin):{id:251,uuid:'436e08bf-6cfd-4f…api_status(pin):"deleting"'} (REMOVED)
```
I'm not really sure what the best way to deal with this. Any ideas @kueda ?


kueda:
In theory [this](https://github.com/inaturalist/inaturalist/blob/master/app/controllers/identifications_controller.rb#L250) should ensure that the observation index in ES is up-to-date before `DELETE /identifications/:id` returns, and [this](https://github.com/inaturalist/inaturalist/blob/master/app/webpack/observations/show/ducks/observation.js#L279) should ensure you're bypassing any caching, so yeah, that's kind of mysterious. Any ideas @pleary?

That said, I think that's probably an issue with our API and not within the scope of this UI work, I wouldn't let that hold you up, Brian.

pleary:
Yes this sounds similar to what was happening in https://github.com/inaturalist/iNaturalistAPI/issues/200#issuecomment-652652554 - the classic Elasticsearch refresh issue. Brian would you mind testing something and see if it fixes this problem for you - on the line after https://github.com/inaturalist/inaturalist/blob/master/app/controllers/identifications_controller.rb#L250 which is `@identification.wait_for_obs_index_refresh = true` could you try adding `@identification.wait_for_index_refresh = true`? And you'll need to add `attr_accessor :wait_for_index_refresh` to the Identification model too.

With this change we will be asking the ID to wait until both the observations and identifications indices are refreshed. I wouldn't think this would be needed since I think the observation indexing happens last, and if we're waiting for a refresh on that I would think the ID index would also have been refreshed. But it would be an interesting test. If that works - great. If not, then as @kueda said this is an issue beyond the scope of this ticket and we can dig into it further later on.

todtb:
Hey @pleary I tested that out without luck. In addition to it happening last, there's also `include Shared::TouchesObservationModule` which should in theory be setting `wait_for_index_refresh` to `true` using the `after_destroy` hook anyway, so I'm not sure.

I'll open up as a non-draft PR, but I do want to reiterate the behavior that gave me hesitation in the first place:

If a user uses the new inline hard delete functionality on an Identification, there's a good chance the Identification card will stay in place which would be confusing. Any attempts to edit/withdraw/delete it then will result in an error modal popping up and then the card disappearing, but obviously not ideal. Additionally, refreshing the page will clear it out as would I imagine any subsequent front end action which calls `setObservation()`.

Just wanted to make sure that was clear.

todtb:
Just saw I messed up and somehow included commits from https://github.com/inaturalist/inaturalist/pull/2727 here. Shouldn't be a problem if https://github.com/inaturalist/inaturalist/pull/2727 is merged first.

app/webpack/shared/components/text_editor.jsx
```diff
@@ -34,30 +35,31 @@ class TextEditor extends React.Component {
...
           this.setState( { textareaChars: e.target.value.length } );
         }
-      };
-    }
+      }
+    };
     return (
       <div className={`TextEditor ${className} ${preview && "with-preview"}`}>
-        { this.textarea && this.textarea.current && (
+        { this.textarea && (
```
kueda, line 58: Hm, why isn't `current` necessary? Don't you need to be sure the ref is pointing at an actual DOM element?

todtb:
Thanks for all the feedback. I'll be able to address the issues early next week but totally understand if you're looking to get this out sooner so feel free to take it from here if urgent. Otherwise I'll pick it up then. Have a great weekend!

app/webpack/shared/components/text_editor.jsx
todtb: So I guess I had the understanding that with a ref, the actual DOM element represented by `.current` is only truly set after the render but before the `componentDidMount` in the component's lifecycle. What I thought that meant was that `.current` is best used in functions outside of render such as: `onClick={

todtb: I moved this check to the handle click function of the TextEditorFormatButton itself, still operating under the assumption that refs are set after render. I also passing a function to setState of the TextEditor's content as I noticed that simply clicking the button without any other changes (for example selecting text and clicking bold or putting in a template/placeholder text) wasn't updating the content correctly.

app/webpack/observations/show/components/activity_item.jsx
```diff
@@ -15,496 +16,618 @@ import UserImage from "../../../shared/components/user_image";
...
-        </Panel.Body>
-        { footer ? <Panel.Footer>{ footer }</Panel.Footer> : null }
-      </Panel>
-    </div>
-  );
-};
+}
 
 ActivityItem.propTypes = {
+  inlineEditing: PropTypes.bool,
```
todtb, line 620: Ideally this is temporary. 

The way that the `ObservationModal` in /identify works vs the observation show page is significant enough that I think it should be treated separately.

Mainly it comes down to `Activity`  component vs `DiscussionList` component. I'm not sure if there are already plans to do so, but maybe the older `DiscussionList` can be replaced with the newer `Activity` feed? Even if not, it feels like completing it for the latter while leaving the former with existing functionality is a good point to separate the work.

kueda: Yeah, this is kind of an ongoing issue. Historically, Identify was our first React page and I made it. Later, we started using React and Redux a bit differently, consolidating actions and reducers into ducks, and still later Patrick made the obs detail page with a mostly-novel set of components. Since then I've worked some of Patrick's obs detail work into Identify, like `ActivityItem`s, but not everything, like the `Activity` component or the `callAPI` action. I think Identify should move in that direction, but it would be kind of a big overhaul and out of scope for this issue, so I think your approach here is good.

todtb:
>There are a few UI issues that I was planning on fixing myself, but if you want to address them, I'll just list them

I'll take a look.

> Regarding Identify, if there's a way to replace DiscussionList with Activity in a separate PR, that would be great. However, I think the reason I haven't done so is that I do think the "Add ID" and "Comment" button should remain where they are in the ObservationModal, which is kind of incompatible with the way Activity works. If you want to propose a way to make Activity work with that UI, I'd certainly be interested in seeing / hearing what you have in mind. Maybe open a separate issue for that if you have an idea.

Was thinking a little bit about it, but yeah, wanted to break it off into something separate.

> extreme social distancing (aka camping)

Haha, nice! That was me last weekend. Thanks for the feedback and enjoy the trip!

todtb:
Very minor thing I noticed while updating styles.

```scss
# observations/show.scss
@import "../shared/activity_item";
...
@import "./identify/observation_modal";
```

```scss
# identify/observation_modal.scss
@import "../../shared/activity_item";
```

This results in the styles for `ActivityItem` being included in the compiled `show.css` a couple times.

We recently both independently added commits doing something similar with `shared/text_editor`.

Hopefully a quick enough fix to restructure in a separate issue.

Issue #2735, opened 2020-07-14
"Log in" and "Sign Up" links on obs show pages point to incorrect addresses


tiwane:
On obs show, if you're not logged in, you'll see this:

![image](https://user-images.githubusercontent.com/15896181/87482478-53f5f780-c5e6-11ea-868e-8c40b9f37ead.png)

But "Log In" points to https://www.inaturalist.org/log_in It should point to https://www.inaturalist.org/login

And "Sign Up" points to https://www.inaturalist.org/sign_up when it should be https://www.inaturalist.org/signup

pleary:
Looks like @todtb's fix is live now. Thank you!

tiwane:
https://forum.inaturalist.org/t/observation-page-broken-link-to-login-page/18587

Looks like it's fixed for English, but not for other languages. eg in Bulgarian I see https://www.inaturalist.org/log_in

When I switch to Arabic it works, but I _think_ that's because those links haven't been translated into Arabic. In every language where those links have been translated, they go to the wrong pages.

kueda:
Needed to sync translations, so I think it's working now, at least for Arabic (not translated) and Bulgarian (translated).

PR #2737, opened 2020-07-15
Update login and signup links in activity

todtb:
#2735

Quick fix. Technically `translations.js` isn't strictly required to be changed as it is regenerated each build right?


kueda:
Right, `translations.js` well get re-written on deploy. However, all of the non-English translation files will also get re-written the next time we merge in changes from Crowdin, so there's not much point in updating those. No harm in doing so either.

Spec failure is my fault, hoping my last commit fixes that.

PR #2743, opened 2020-07-17
Removed agree functionality in a few places

kueda:
* Removed agree button from ident activity item when
  * ident is not leading
  * observation is RG
* Disabled agree keyboard shortcut in Identify when
  * it's your obs
  * obs is RG
  * CID exists and is species or lower

### DIFF
--- a/app/webpack/observations/identify/actions/identification_actions.js
+++ b/app/webpack/observations/identify/actions/identification_actions.js
@@ -84,7 +84,18 @@ function agreeWithCurrentObservation( ) {
       return null;
     }
     const currentObservation = s.currentObservation.observation;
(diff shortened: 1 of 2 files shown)


boucur:
Looks like it's encouraging click-agreeing based on the thumbnails on Identify now though ("thumbnailing"), instead of opening to review the photo(s) larger, which now requires manually typing in the name.

kueda:
@boucur, I literally released this an hour ago. What makes you think people are clicking those Identify grid buttons more? That said, I did totally forget about those buttons and will remove them where appropriate.

boucur:
oh, just personal inclination

Issue #2751, opened 2020-07-28
Adding ID to a "unknown" obs with withdrawn ID makes your ID a disagreement to withdrawn ID


tiwane:
From https://forum.inaturalist.org/t/disagreeing-with-a-withdrawn-id-doesnt-change-it-from-unknown/7964

1. Find an "unknown" obs that has a withdrawn ID (or create one).
2. Add an ID to it.
3. Your ID is added, then is changed to a disagreeing ID.

![image](https://user-images.githubusercontent.com/15896181/88710846-0f7d5800-d0cc-11ea-848b-d60cd6193a3b.png)

If original ID is deleted, the second ID is still a disagreeing one.

![image](https://user-images.githubusercontent.com/15896181/88711000-54a18a00-d0cc-11ea-90b2-320b67449f74.png)

todtb:
Looks like a few things going on here:

- [ ] An identification added to an observation without any current identifications should not be a disagreement

- [ ] Deleting an identification should update any other identification that was in disagreement to no longer be disagreements

#2769 Addresses the first

kueda:
While #2769 seems harmless and sensibly defensive, I'm pretty sure the more serious problem is at https://github.com/inaturalist/inaturalist/blob/main/app/models/identification.rb#L182, where we are trying to handle the situation when you are disagreeing with yourself and you are the only identifier, i.e. when there may not be a `probable_taxon` or a `community_taxon`. I think what's needed there is an additional test that covers the case where you are identifying an obs with one withdrawn identification by another user, and then those lines that look up a previous identification need to only look up the previous identifications by the user who is adding the *current* identification.

> Deleting an identification should update any other identification that was in disagreement to no longer be disagreements

I'm not entirely sure what this means. Could you provide a concrete example of when this seems appropriate? In my mind, deleting an identification should never change the disagreement state of other identifications. Identifications aren't really disagreements with other identifications. They are statements of what taxon the user thinks the observation depicts, and the disagreement is with another *taxon*, so if I say it's a flying fish and not a pickleweed, no one else's identification should change that statement.

todtb:
[...]

PR #2769, opened 2020-08-09
Prevent disagreeing with non-current identifications

todtb:
#2751 

Fixes first issue
> An identification added to an observation without any current identifications should not be a disagreement

### DIFF
--- a/app/models/identification.rb
+++ b/app/models/identification.rb
@@ -195,6 +195,12 @@ class Identification < ActiveRecord::Base
       return true
     end
 
+    # Can't disagree if no current identifications
+    if observation.identifications.current.empty?
+      self.disagreement = nil
+      return true
+    end
+
     # Can't disagree when suggesting an ungrafted taxon
     if !taxon.grafted? && !taxon.children.exists?
       self.disagreement = nil
--- a/app/models/identification.rb
+++ b/app/models/identification.rb
@@ -121,7 +121,6 @@ class Identification < ActiveRecord::Base
     c[0] = "taxa.id = #{taxon.id} OR #{c[0]}"
     joins(:taxon).where(c)
   }
-  scope :older_than, -> (date) { where("created_at < ?", date) }
   scope :on, lambda {|date| where(Identification.conditions_for_date("identifications.created_at", date)) }
   scope :current, -> { where(:current => true) }
   scope :outdated, -> { where(:current => false) }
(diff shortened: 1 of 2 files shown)


spec/models/identification_spec.rb
```diff
@@ -1046,6 +1046,13 @@
     i.reload
     expect( i ).not_to be_disagreement
   end
+  it "should not be automatically set to true if no other identifications are current" do
+    o = make_research_grade_candidate_observation
```
kueda, line 1050: Not a major problem, but all this test requires is an observation with one withdrawn Identification, so instead of incurrent the overhead of making a research-grade candidate, you could do something like `o = Identification.make!( current: false ).observation`.

app/models/identification.rb
```diff
@@ -179,8 +180,8 @@ def set_previous_observation_taxon
     end
     unless previous_observation_taxon_id
       working_created_at = created_at || Time.now
-      previous_ident = observation.identifications.select{|i| i.persisted? && i.current && i.created_at < working_created_at }.last
-      previous_ident ||= observation.identifications.select{|i| i.persisted? && i.created_at < working_created_at }.last
+      previous_ident = observation.identifications.by(user_id).older_than(working_created_at).current.select(&:persisted?).last
+      previous_ident ||= observation.identifications.by(user_id).older_than(working_created_at).select(&:persisted?).last
```
kueda, line 184: Again, not blocking, but using scopes here means that if the second line runs here, you will query the database a second time. `observation.identifications.select` will hit the database once (or maybe not at all if it was already called) and load all the observation's identifications into memory, and all subsequent calls to `.select` will probably be faster than their database-based equivalents. All of the Identification callbacks could probably use an audit to minimize db queries, and I realize I'm picking nits here, but identification creation is a high-volume, low performance operation as it is.

todtb: Hmm, so this is not something I've benchmarked (or probably something I realistically could in a dev environment) but my initial thought was kind of the opposite. 

I was thinking that by delegating the filtering to SQL we could avoid hydrating many instances that would be immediately thrown out by the select. That said, yeah, since the `persisted?` call will resolve the AR relation anyway it is possible to require two queries. Also, now that I'm thinking about it more, the number of associated identifications per observation is probably not great enough to really see an issue with loading them all into memory/deserializing/etc. 

I'll change it back.

Issue #2675, opened 2020-05-21
Clean up taxon name lexicons


kueda:
They're a mess, b/c we allow people to enter anything.

- [ ] Add a `unique_lexicon` column to the `taxon_names` table, give it a unique index, populate it with `parameterize(lexicon)` on save, and add a `validates_uniqueness_of` validation for it. When a user tries to add a lexicon that doesn't pass this validation, show them the lexicon they were about the duplicate
- [ ] Add a validation to check if `parameterize(lexicon)` is blank, and if so, remind the user that lexicons are supposed to be in English and translations should be handled like other translated text, i.e. "Japanese" instead of "日本語"
- [ ] Reject any new names with lexicons that match any translations of the `lexicons` translation keys, e.g. if a user tries to enter "Anglais" as a lexicon, it should fail b/c `I18n.t( "lexicons.english", locale: "fr" ) == "Anglais" && I18n.t( "lexicons.english", locale: "en" ) != "Anglais"`
- [ ] Clean up / combine all the lexicons that don't pass these new rules (this is the hard part)

todtb:
I'd gladly to take a look at this, but I'm a bit confused by the desired functionality.

We could add (with null for the data migration stage):
```ruby
add_column :taxon_names, :unique_lexicon, :string
add_index :taxon_names, :unique_lexicon, unique: true, where: "unique_lexicon IS NOT NULL"
```
but this means we would only support a single `taxon_name` per lexicon. 

Should this be a composite index of lexicon and name instead? It couldn't be on taxon_id as we still want to support multiple common names in each lexicon right?

Also if adding a new lexicon such as "Ido", this would be allowed since there is no lexicon translation key for the language? How do we differentiate this from "伊多語"?

Ultimately the goal is to avoid something like this?
| id  | name  | lexicon  | taxon_id  | 
|---|---|---|---|
| 1  |  Fungi |  German | 9  |
| 2  | Pilze  |  German  |  9  | 
| 3 | Pilze  |  Deutsch  |  9 |
| 4 | Pilze  |  deutsche sprache  |  9 |

kueda:
> Should this be a composite index of lexicon and name instead? It couldn't be on taxon_id as we still want to support multiple common names in each lexicon right?

Yeah, that sounds better. Sorry, I clearly didn't think that through very well.

> Also if adding a new lexicon such as "Ido", this would be allowed since there is no lexicon translation key for the language? How do we differentiate this from "伊多語"?

[...]

PR #2774, opened 2020-08-13
Add validations for custom lexicon for taxon names

todtb:
What do you think of this @kueda for checking off a few boxes from #2675?

The first validation will check and recommend the English name for any lexicon we have a translation for.

The second validation will restrict custom lexicon names to the latin character set plus punctuation, spaces, and right now digits. Since these should all be English, we shouldn't allow other character sets through. Let me know if you want to tweak the regex though.

So before jumping further into the list on the path to cleaning up the data, I'd like to know what db constraints, indexes, etc. might be introduced non-disruptively and what kind of real data we will be looking at. In order to do so I was wondering if it would be possible to get an export of select columns (`name`, `lexicon`, and `taxon_id` and possibly `source` and `position`)? Or, since the table doesn't seem to be sensitive in nature, everything but the with the `creator_id` and `updater_id` scrubbed. If not for either data or logistic reasons, I would completely understand.

### DIFF
--- a/app/models/taxon_name.rb
+++ b/app/models/taxon_name.rb
@@ -16,8 +16,7 @@ class TaxonName < ActiveRecord::Base
   validates_format_of :lexicon, with: /\A[^\/,]+\z/, message: :should_not_contain_commas_or_slashes, allow_blank: true
   validate :species_common_name_cannot_match_taxon_name
   validate :valid_scientific_name_must_match_taxon_name
-  validate :english_lexicon_if_exists
-  validate :latin_character_set
+  validate :english_lexicon_if_exists, if: Proc.new { |tn| tn.lexicon && tn.lexicon_changed? }
   NAME_FORMAT = /\A([A-z]|\s|\-|×)+\z/
   validates :name, format: { with: NAME_FORMAT, message: :bad_format }, on: :create, if: Proc.new {|tn| tn.lexicon == SCIENTIFIC_NAMES}
   before_validation :strip_tags, :strip_name, :remove_rank_from_name, :normalize_lexicon
@@ -415,6 +414,17 @@ class TaxonName < ActiveRecord::Base
     name = name.gsub(/\s+[A-Z].*/, '')
     name.strip
   end
+  
+  def self.find_lexicons_by_translation(translation)
+    lex_by_loc = I18n.available_locales.each_with_object({}) do |loc, hash|
+      hash[loc] = I18n.with_locale(loc) { I18n.t(:lexicons) }
+    end
+    match_loc, match_lexes = lex_by_loc.reject{|l| l.match("en") }
+                                 .transform_values { |loc| loc.transform_values { |lex| lex.downcase.strip } }
+                                 .find { |_loc, lexes| lexes.values.include?(translation.downcase.strip) }
+    
+    { locale: match_loc, lexicons: match_lexes }
+  end
 
   private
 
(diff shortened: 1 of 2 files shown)


app/models/taxon_name.rb
```diff
@@ -413,4 +415,27 @@ def self.strip_author(name)
...
   end
+
+  private
+
+  def english_lexicon_if_exists
+    en_lexicons = I18n.with_locale(:en) { I18n.t(:lexicons) }.values
+    translated_lexicons = I18n.available_locales.map { |loc| I18n.with_locale(loc) { I18n.t(:lexicons) } }
+    non_en_lexicons = (translated_lexicons.collect(&:values).flatten.uniq - en_lexicons).map!{ |l| l.downcase.strip }
+
+    if non_en_lexicons.include?(lexicon.downcase.strip)
```
kueda, line 426: I'm not entirely sure *why* we allow the lexicon to be blank, but apparently we do, so guarding against that here will get you past the spec failures in Travis.

config/locales/en.yml
```diff
@@ -7425,6 +7425,10 @@ en:
           attributes:
             lexicon:
               should_not_contain_commas_or_slashes: should not contain commas or slashes
+              should_match_english_translation: should match English translation, did you mean "%{suggested}"?
```
kueda, line 7428: This is a tricky concept to convey. How about "should be in English, but currently it matches the translation of "%{suggested}" into %{locale}. Use "%{suggested}" if that is what you meant and it will be translated when you view the site in that language"

app/models/taxon_name.rb
```diff
@@ -16,6 +16,8 @@ class TaxonName < ActiveRecord::Base
   validates_format_of :lexicon, with: /\A[^\/,]+\z/, message: :should_not_contain_commas_or_slashes, allow_blank: true
   validate :species_common_name_cannot_match_taxon_name
   validate :valid_scientific_name_must_match_taxon_name
+  validate :english_lexicon_if_exists
+  validate :latin_character_set
```
kueda, line 20: IMO, this is probably taking it a bit too far. People add names not only so they can "use" them on iNat when searching for taxa but also so they can be listed on the taxon page where people can learn about them or at least be made aware they exist. While I want to avoid duplication, I don't want to suppress the addition and correct spelling of languages that we do not currently support for translation. There are vanishingly few such lexicons that have been added, but I want to avoid a situation where someone wants to see their language represented on the website and can't do it properly b/c we force them to use Latin characters.

app/models/taxon_name.rb
```diff
@@ -16,6 +16,8 @@ class TaxonName < ActiveRecord::Base
   validates_format_of :lexicon, with: /\A[^\/,]+\z/, message: :should_not_contain_commas_or_slashes, allow_blank: true
   validate :species_common_name_cannot_match_taxon_name
   validate :valid_scientific_name_must_match_taxon_name
+  validate :english_lexicon_if_exists
```
kueda, line 19: This should have `on: :create`. Without that that none of the other properties of an existing "invalid" name could be updated, like `is_valid` or `position`.

app/models/taxon_name.rb
todtb: A little digging found app/helpers/default_form_builder.rb:137 includes the option "Unknown" value blank which app/models/taxon_name.rb:181 in turns sets to nil. If you can't think of an immediate use case as to why nil is allowed, it makes me wonder if all nil values should default to `""` or `"Unknown"` with the appropriate db constraint `NOT NULL`. Still investigating though, so not sure. Maybe the data dumps will reveal some trends.

app/models/taxon_name.rb
todtb: Good call! I'm sure there will be additional considerations that arise after inspecting the existing data.

app/models/taxon_name.rb
todtb: Hmm, yeah definitely don't want to prohibit anyone's language from being transliterated into English properly. This is an extended latin character set with all sorts of diacriticals, but still I won't pretend to know if the thousands of languages can be represented with even that. I'll pull this.

todtb:
Hey @kueda. Quick look at the data and initial research reveals the following. Not too bad. I think that this PR's changes (pending review) should address any more cases like this from popping up. I don't see why this couldn't be incorporated ahead of further work if everything looks good.

```sql
select count(*), lexicon from taxon_names
where LOWER(lexicon) IN #TRANSLATABLE_LEXICONS#
group by lexicon;
```

|Count|Lexicon                      |
|-----|-----------------------------|
|195  |Nombres Científicos          |
|168  |Русский                      |
|119  |學名                           |
|115  |Deutsch                      |
|104  |Português                    |
|95   |Español                      |
|57   |Научные названия             |
|49   |Français                     |
|28   |Italiano                     |
|28   |学名                           |
|22   |Nomi Scientifici             |
|22   |Noms Scientifiques           |
|21   |Nomes Científicos            |
|16   |Norsk                        |
|11   |Català                       |
|3    |Waray-Waray                  |
|2    |Iloko                        |
|2    |Tieteelliset Nimet           |
|2    |Videnskabelige Navne         |
|2    |Wetenschappelijke Namen      |
|1    |Davawenyo                    |
|1    |Hiligaynon                   |
|1    |Misima-paneati               |

Optilete:
If I add the lexcion Khanty twice I should expect that the error message would occur the second time I add the lexicon, right?

kueda:
@Optilete I don't really know what you mean by adding it "twice." You can use it twice and there should be no error. "Khanty" is not a translation of a language name in any of our files, so I wouldn't expect that to raise any errors.

Issue #2779, opened 2020-08-19
Web Uploader: Warn users if they are submitting an observation without an ID


tiwane:
From https://forum.inaturalist.org/t/warn-me-about-missing-ids-when-uploading-observations-on-website/13597

Currently, we have two warning pop-ups if a user submits observations which are lacking attributes:

`You are submitting observations without photos and taxon names. Observations without this information are often impossible to identify and will not be visible by default in observation searches.`

and 

`You are submitting observations without dates or locations. Observations without this information are often impossible to identify and will not be visible by default in observation searches.`

If the user submits observations which meet both of those cases, they see both warnings, one after the other.

Users have requested that we add a warning if they have not added an identification to any of their observations to prevent them from the purgatory of taxonomic limbo. I'm loathe to add a third attribute-related warning, so my suggestion would be to change "and" to "or"  for the first pop-up and have it appear if any observation is missing a photo or a taxon name/placeholder.

`You are submitting observations without photos or taxon names. Observations without this information are often impossible to identify and will not be visible by default in observation searches.`

I _think_ this makes sense...

todtb:
As mentioned in the forum:
>  (the) error message only appeared for observations missing IDs and media evidence like a photo

and a quick glance shows that is the logic in place.

It shouldn't be difficult to add a check for taxon id (or species guess) on each observation just on the web uploader. I can start working on that.

To confirm @tiwane, display a warning "You are submitting observations without photos or taxon names. Observations without this information are often impossible to identify and will not be visible by default in observation searches." if 1) the observation does not have any files attached, or if 2) the observation does not have either a taxon id or species guess. Sound right?

kueda:
@todtb's PR is working as you described, Tony, but now the error message is misleading, particularly the part about observations not being "visible by default in observation searches," which is not technically true for observations without media. How would you feel about something more verbose but accurate like this?

**Some observations are missing media or identifications.** Observations without media will not be visible by default in searches, and those without any identification may not be seen by experts looking for specific taxa. Even a very coarse identification like "plants" can help a plant expert find your observation.

tiwane:
Sounds good, thanks.

PR #2781, opened 2020-08-19
Add warning to obs uploader

todtb:
#2779 

Updates submit check in obs uploader to check for either no photos OR no taxon name (or species guess)


todtb:
I've updated to your suggested text @kueda. However, the current modal doesn't render html from the passed prop. I could add `dangerouslySetInnerHTML` to the modal, however the prop `message` could be anything (such as a surprise `<script>` tag). I would imagine most props being passed through are strings from translations, however I haven't thoroughly checked and there wouldn't be anything guarding against future use. Therefore I figured it would be better to hold off for now on the `<strong>` tag, unless you feel comfortable with that or have a good solution.

PR #2786, opened 2020-08-25
Add parameterized lexicon to taxon_name

todtb:
#2675

I think this is the next step to reducing lexicon variation as well as duplicate names in the same lexicon.

* The backfilling of `parameterized_lexicon` took 2-3 minutes for me if that needs to be a consideration in your planning on when to run the migration.

* Right now it's tough to put in the index as there are already several, albeit only a small fraction (2964/2M+), records that would violate the composite uniqueness constraint. It looks like most if not all are coming from imports. 

* Also, might be worth reexamining the role of `normalize_lexicon` now that `parameterized_lexicon` exists. I think the former will first change "Chinese (TraDitional)" -> Chinese (Tra Ditional)". With `parameterized_lexicon` this results in "chinese-traditional" and "chinese-tra-ditional". 

* Nil/""/null values still need to be allowed for some edge cases in the imports as well as that currently representing "Unknown" for user submitted taxon_names

Let me know if there are any special considerations I should be aware of for iNat DB migrations, as I believe this will be the first I've contributed.

### DIFF
--- a/spec/models/taxon_name_spec.rb
+++ b/spec/models/taxon_name_spec.rb
@@ -80,9 +80,14 @@ describe TaxonName, 'creation' do
   end
 
   it "should not be valid with a non-English translation of a lexicon" do
-    tn = TaxonName.make(lexicon: "简体中文", name: "common")
+    tn = TaxonName.make(lexicon: I18n.with_locale(:"zh-CN") { I18n.t("lexicons.finnish") }, name: "common")
     expect(tn).to_not be_valid
-    expect(tn.errors.messages[:lexicon].count).to eq 2
+    expect(tn.errors.messages[:lexicon]).to include(
+      I18n.t("activerecord.errors.models.taxon_name.attributes.lexicon.should_match_english_translation", 
+             suggested: I18n.with_locale(:en) { I18n.t("lexicons.finnish") }, 
+             suggested_locale: I18n.t("locales.zh-CN")
+      )
+    )
   end
   
   it "should parameterize and store lexicon" do
@@ -93,7 +98,9 @@ describe TaxonName, 'creation' do
   it "should not parameterize and store lexicon with invalid characters" do
     tn = TaxonName.make(lexicon: "测试", name: "common")
     expect(tn).to_not be_valid
-    expect(tn.errors.messages[:lexicon].count).to eq 1
+    expect(tn.errors.messages[:lexicon]).to include(
+      I18n.t("activerecord.errors.models.taxon_name.attributes.lexicon.should_be_in_english")
+    )
   end
   
   it "should not validate presence of a parameterizable lexicon if lexicon not present" do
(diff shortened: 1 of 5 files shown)


kueda:
Cool, mostly looking good here. Not sure why Github isn't displaying Travis build info for PRs anymore, but you have a spec failure: https://travis-ci.org/github/inaturalist/inaturalist/jobs/721150920#L3072

Regarding migrating the data, it might be considerably faster on our production machines, I guess we'd have to test. When we need to migrate data and it's going to take a while, sometimes we just do it in the console or in a script and add something like this to the migration"

```ruby
say <<-EOT
You're going to have to run this to migrate the data. Takes 2-3 mins on dev hardware.

    TaxonName.uniq.where.not(lexicon: nil).pluck(:lexicon).each do |l|
      next unless l.parameterize.present?
      TaxonName.where(lexicon: l).update_all(parameterized_lexicon: l.parameterize)
    end
EOT
```
So if you want we can do that, or try to run the migration and see how bad it is.

todtb:
> Cool, mostly looking good here. Not sure why Github isn't displaying Travis build info for PRs anymore, but you have a spec failure: https://travis-ci.org/github/inaturalist/inaturalist/jobs/721150920#L3072

I noticed that too and double checked. Fix forthcoming.

> Regarding migrating the data, it might be considerably faster on our production machines, I guess we'd have to test. When we need to migrate data and it's going to take a while, sometimes we just do it in the console or in a script and add something like this to the migration"

If it's not going to be an issue on prod, I think we're good. Probably not likely most dev setups are going to have the numbers (full table dump) where this is a concern.

spec/models/taxon_name_spec.rb
```diff
@@ -79,6 +79,28 @@
     expect( TaxonName.make!( lexicon: " Foo" ).lexicon ).to eq "Foo"
   end
 
+  it "should not be valid with a non-English translation of a lexicon" do
+    tn = TaxonName.make(lexicon: "简体中文", name: "common")
```
todtb, line 83: Would be cool to pull this from `t()` instead, but not sure how to get i18n into specs easily.

spec/models/taxon_name_spec.rb
```diff
@@ -79,6 +79,28 @@
...
+  
+  it "should parameterize and store lexicon" do
+    tn = TaxonName.make!(lexicon: TaxonName::LEXICONS[:CHINESE_SIMPLIFIED], name: "common")
+    expect(tn.parameterized_lexicon).to eq "chinese-simplified"
+  end
+  
+  it "should not parameterize and store lexicon with invalid characters" do
+    tn = TaxonName.make(lexicon: "测试", name: "common")
+    expect(tn).to_not be_valid
+    expect(tn.errors.messages[:lexicon].count).to eq 1
```
todtb, line 96: Same as above but for `expect(tn.errors.messages[:lexicon]).to include(t(:should_be_english))`

spec/models/taxon_name_spec.rb
kueda: I *think* it should work: https://github.com/inaturalist/inaturalist/blob/main/spec/models/observation_spec.rb#L438

Issue #2566, opened 2020-01-22
moving children followed by taxon change not updating indexes


loarie:
to rearrange the taxonomy as describe here:
https://inaturalist.github.io/trees/tree_figure.html?tfr_id=100795
all active children of Tyrannidae 15984 (except Terenotriccus Onychorhynchus & Myiobius) were moved to Tyrannidae 980017. And then immediately after https://www.inaturalist.org/taxon_changes/67564 was committed

However observations belonging to the moved children (e.g. https://api.inaturalist.org/v1/observations/37631285 belonging to moved child Pyrocephalus) is indexed with a stale copy of the taxon ancestry in the production ES index that references 15984 not 980017

This sequence of actions:
(1) move children from input to output
(2) commit taxon change
shouldn't prevent the indexes from updating

loarie:
note to self use ```Taxon.elastic_search( filters: [ terms: { id: [o.taxon.id] } ] ).response.hits.hits[0]._source.ancestor_ids``` to see indexed ancestry of an obs

PR #2804, opened 2020-09-09
Reassess community taxa after taxon activation

loarie:
As reported here https://forum.inaturalist.org/t/obs-with-only-subspecies-ids-can-get-stuck-without-a-cid/15992
The CID algorithm for an observation ignores inactive taxa
But if an taxon is activated the CID of its observations are not reassessed
This means that if a curator does the following set of actions which is common in curating taxa: (1) inactivates a taxon and moves it to a new taxon (this wipes the CIDs because moving a taxon reassesses CID and inactive taxa are ignored)
(2) reactivates the taxon
the CID's are not reassessed/restored because this doesn't happen when a taxon is activated

This pull request introduces a callback that reassesses the CIDs of obs associated with a taxon if the taxon is_active status changes

### DIFF
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -19,8 +19,6 @@ class Taxon < ActiveRecord::Base
   # Skip the more onerous callbacks that happen after grafting a taxon somewhere else
   attr_accessor :skip_after_move
 
-  attr_accessor :skip_after_activate
-
   attr_accessor :locale
 
   # set this when you want methods to respond with user-specific content
@@ -526,7 +524,6 @@ class Taxon < ActiveRecord::Base
 
   def handle_after_activate
     return true unless is_active_changed?
-    return true if skip_after_activate
     Observation.delay( priority: INTEGRITY_PRIORITY, queue: "slow",
       unique_hash: { "Observation::update_stats_for_observations_of": id } ).
       update_stats_for_observations_of( id )
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -19,6 +19,8 @@ class Taxon < ActiveRecord::Base
   # Skip the more onerous callbacks that happen after grafting a taxon somewhere else
   attr_accessor :skip_after_move
 
+  attr_accessor :skip_after_activate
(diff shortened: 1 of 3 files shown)


spec/models/observation_spec.rb
```diff
@@ -3095,6 +3095,31 @@
       expect(o.identifications.count).to eq 2
     end
 
+    it "change should be triggered by activating a taxon" do
+      load_test_taxa
+      o = Observation.make!
+      i1 = Identification.make!( observation: o, taxon: @Pseudacris_regilla_regilla )
```
kueda, line 3101: If this test doesn't rely on the taxon being a subspecies, could you switch to using `@Pseudacris_regilla` and remove the additional taxon you're adding in `load_test_taxa`? `load_test_taxa` is really onerous and runs all over the place, so adding more work there will slow down a lot of tests. Or, if you only need a genus -> species -> subspecies taxonomy and not a full branch from subspecies down the root, it might be better to use the blueprints to set up the taxonomy you need in this test instead of using `load_test_taxa`

app/models/taxon.rb
```diff
@@ -19,6 +19,8 @@ class Taxon < ActiveRecord::Base
   # Skip the more onerous callbacks that happen after grafting a taxon somewhere else
   attr_accessor :skip_after_move
 
+  attr_accessor :skip_after_activate
```
kueda, line 22: Is this actually getting used or was it just copied from `skip_after_move` / `handle_after_move`?

spec/models/observation_spec.rb
loarie: sure - I was trying to replicate Jane's case - but I since learned that this isn't ssp specific so will change it

app/models/taxon.rb
loarie: I guess thats not being used, I can remove it

loarie:
@kueda - if you can give this another look and/or merge thank you

kueda:
Any chance this was at the root of #2566?

PR #2805, opened 2020-09-10
Taxon split with same input output

loarie:
I altered taxon splits so that if one of the output taxa is the same as the input taxon the split is still allowed. It leaves the input taxon active and doesn't touch content (identificiations etc.) determined (by atlases) to stay with that taxon. Content determined to go to the other output taxon behaves normally.

The idea here is that with a taxon lump (e.g. Taricha torosa is folded into Taricha granulosa thereby broadening Taricha granulosa) we don't recommend making a new taxon for the broadened Taricha granulosa. Instead we simply swap Taricha torosa with Taricha granulosa and advise curators to make sure that any content may no longer match the broadened Taricha granulosa (e.g. taxon ranges, common names) be manually altered. And with lumps, IDs don't become misspecified - an ID of narrow Taricha granulosa still applies to broadened Taricha granulosa.

With splits (e.g. splitting Taricha torosa off from Taricha granulosa), in the past we have always recommended creating a new taxon for the narrowed taxon (e.g. Taricha granulosa). However, sometimes this creates a lot of disruption with content associated with the narrowed taxon (e.g. removing IDs of Taricha granulosa only to replace them with IDs of Taricha granulosa when its not always clear to the community why they are different) for the sole benefit of being able to preserve a record of how the broader Taricha granulosa compares to the narrower Taricha granulosa. But since we're not doing this for lumps (as explained above) it doesn't seem totally consistent/necessary to do them for splits and the system seems to be working fine with lumps.

Also, in many cases with splits the carved off taxon (e.g. Taricha torosa) has an extremely small range (e.g. it occurs on just a tiny island) carved off from the narrowed taxon (e.g. Taricha granulosa) which is largely unchanged. It seems overkill to touch all the often thousands of IDs associated with such only slightly narrowed taxa such as this to accomodate carving off an island.

In case its not clear, this functionality still uses 'normal split' functionality to properly handle identifications and other content moving towards the carved off taxon (e.g. Taricha torosa) so IDs don't become misspecified anymore than they would with a normal split. The only cost we're incurring here is a slightly less concrete record of how the broadened taxon as distinct from the narrowed taxon. But I think thats well worth it to minimize disruption caused by these massive splits.

### DIFF
--- a/app/models/taxon_change.rb
+++ b/app/models/taxon_change.rb
@@ -119,6 +119,7 @@ class TaxonChange < ActiveRecord::Base
     # inputs can't have active children
     if ["TaxonSwap", "TaxonSplit", "TaxonDrop"].include? type
       return false if !input_taxa.map{|t| t.children.any?{ |e| e.is_active }}.any?
+      return false if is_a?( TaxonSplit ) && is_branching?
     # unless they are also inputs
     elsif type == "TaxonMerge"
       return false if !input_taxa.map{|t| t.children.any?{ |e| e.is_active && (!input_taxa.map(&:id).include? e.id) }}.any?
@@ -129,7 +130,7 @@ class TaxonChange < ActiveRecord::Base
   end
   
   def is_branching?
-    taxon_change_taxa.map(&:taxon_id).include? taxon_id
+    taxon_change_taxa.map( &:taxon_id ).include? taxon_id
   end
 
   def committable_by?( u )
@@ -185,11 +186,11 @@ class TaxonChange < ActiveRecord::Base
       raise RankLevelError, "Output taxon rank level not coarser than rank level of an input taxon's active children"
       return
     end
-    unless is_a?(TaxonSplit) && is_branching?
+    unless is_a?( TaxonSplit ) && is_branching?
       input_taxa.each {|t| t.update_attributes!(is_active: false, skip_only_inactive_children_if_inactive: (move_children? || !active_children_conflict?) )}
     end
     output_taxa.each do |t|
-      next if is_a?(TaxonSplit) && t.id == taxon_id
+      next if is_a?( TaxonSplit ) && t.id == taxon_id
       t.update_attributes!(
         is_active: true,
         skip_only_inactive_children_if_inactive: move_children?,
--- a/app/models/taxon_change.rb
+++ b/app/models/taxon_change.rb
@@ -127,7 +127,11 @@ class TaxonChange < ActiveRecord::Base
     end
     return true
   end
-    
+  
+  def is_branching?
+    taxon_change_taxa.map(&:taxon_id).include? taxon_id
+  end
+
   def committable_by?( u )
     return false unless u
     return false unless u.is_curator?
@@ -181,11 +185,11 @@ class TaxonChange < ActiveRecord::Base
       raise RankLevelError, "Output taxon rank level not coarser than rank level of an input taxon's active children"
       return
     end
-    unless type == "TaxonSplit" && input_taxa.count == 1 && output_taxa.map{|a| a.id == input_taxa[0].id}.any?
+    unless is_a?(TaxonSplit) && is_branching?
       input_taxa.each {|t| t.update_attributes!(is_active: false, skip_only_inactive_children_if_inactive: (move_children? || !active_children_conflict?) )}
     end
     output_taxa.each do |t|
-      next if type == "TaxonSplit" && input_taxa.count == 1 && t.id == input_taxa[0].id
+      next if is_a?(TaxonSplit) && t.id == taxon_id
       t.update_attributes!(
         is_active: true,
         skip_only_inactive_children_if_inactive: move_children?,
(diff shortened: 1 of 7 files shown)


loarie:
ok thanks - I'd prefer to leverage the existing Split type and UI for this rather than add a whole new type. but I can add that refactor with an is_branching method to repeat myself less

loarie:
@kueda sorry about all the commits here, embarrassing, I guess my head wasn't in the game yesterday. But this is now well tested on staging and I think ready to merge if you're cool with it

app/models/taxon_change.rb
```diff
@@ -126,7 +128,15 @@ def active_children_conflict?
     end
     return true
   end
-    
+  
+  def is_branching?
+    taxon_change_taxa.map( &:taxon_id ).include? taxon_id
```
kueda, line 133: Since only a `TaxonSplit` can be "branching" can you move this to TaxonSplit? Putting it in `TaxonChange` means *all* taxon changes will respond to this, even thought `is_branching?` is meaningless for a `TaxonMerge`.

Also, I tend to find it more readable when we use the `input_taxa` and `output_taxa` methods, e.g. `output_taxa.map(&:id).include?( input_taxon.id )`. It does require loading those associates which is a bit of a performance burn, but I think they're probably already loaded anyway in most cases where this method would be called.

app/models/taxon_change.rb
```diff
@@ -126,7 +128,15 @@ def active_children_conflict?
...
+  end
+
+  def need_not_push_to?( output_taxon_id )
```
kueda, line 136: I don't know about you but I find negative boolean checks like this kind of confusing. How about

```ruby
def automatable_for_output?( output_taxon )
  output_taxon_id = output_taxon.is_a?( Taxon ) ? output_taxon.id : output_taxon
  if is_a?( TaxonSplit ) && is_branching?
    input_taxon.id == output_taxon_id
  else
    automatable?
  end
end
```

Then you'd need to negate this in most of your usage, e.g. `!taxon_change.automatable_for_output?( output_taxon )`, which also presents a readability challenge, but I think is slightly more readable. Personal preference I guess. If you prefer your current language, could you at least add a comment to the method to make it clear what you mean? Because "push" is not a verb we're currently using to describe the process of updating records associated with the taxa involved in a taxon change.

app/models/taxon_change.rb
```diff
@@ -180,8 +190,11 @@ def commit
       raise RankLevelError, "Output taxon rank level not coarser than rank level of an input taxon's active children"
       return
     end
-    input_taxa.each {|t| t.update_attributes!(is_active: false, skip_only_inactive_children_if_inactive: (move_children? || !active_children_conflict?) )}
+    unless is_a?( TaxonSplit ) && is_branching?
+      input_taxa.each {|t| t.update_attributes!(is_active: false, skip_only_inactive_children_if_inactive: (move_children? || !active_children_conflict?) )}
+    end
     output_taxa.each do |t|
+      next if is_a?( TaxonSplit ) && t.id == taxon_id
```
kueda, line 197: Is it possible to make a branching TaxonSplit with an inactive input? If so, we *do* want to activate it when committed, right?

kueda (commented):
Again, nothing blocking, mostly stylistic / readability concerns.

app/models/taxon_change.rb
loarie: I'm fine with that language but I want the method to do this
```ruby
def automatable_for_output?( output_taxon )
    output_taxon_id = output_taxon.is_a?( Taxon ) ? output_taxon.id : output_taxon
    if is_a?( TaxonSplit ) && is_branching?
      return input_taxon.id == output_taxon_id
    end
    return false
  end
```
because if its not a branching taxon split with the output under consideration matching the input , then I don't want to skip handling the input taxon content regardless of whether its automatable. So I'm not sure if adapting this term automatable for this new context adds confusion or not. But if you're ok with the above I'm fine with it

kueda: If the change is a `TaxonSwap` or a `TaxonMerge`, then isn't it always automatable for the output? Or if it only ever applies to a `TaxonSplit`, maybe move it to that class and get rid of the type check in the method?

loarie: I'm not trying to get at automatable here though - unless I'm misunderstanding
automatable in the past means is there a single output or are atlases in place to shuttle content to a single output
what I want here is 'is the output I'm considering the same taxon as the input for this taxon change?'
I'm fine with calling that 'automatable for output' I was just mentioning that it might be confusing since its kind of a different thing than how we use the term automatable elsewhere. Maybe 'in_place_for_output?' or something?

kueda: > what I want here is 'is this output I'm considering the same taxon as the input for this taxon split?'

I understand, but what I'm getting at here is that it's not immediately clear *why* that is important. Like, if I'm reading this, `input_matches?( output taxon )` provides no more insight into the purpose of the check than `input_taxon == output_taxon` does, but naming the method `automatable_for_output?` implies the purpose of that check, i.e. "for this split, is it ok for us to automatically update records associated with the input for this output?" Generally the answer should be yes, but not if that output taxon is a part of a branching split and is the same as the input taxon.

Again, I'm picking nits and I don't mean for this to block your merge, so go ahead if you want. I'm just trying to preserve my sanity the next time I have to dig into this stuff.

loarie:
@kueda ok - I think everything above is addressed -

Issue #2810, opened 2020-09-20
Some links on /people result in a whale shark


tiwane:
From: https://forum.inaturalist.org/t/whaleshark-when-clicking-on-observations-in-leaderboards/16493

Clicking on "observations" or "identifications" below any user on https://www.inaturalist.org/people results in a whale shark.

Clicking on "observations" on a page like https://www.inaturalist.org/people/leaderboard/2020/7 results in a whale shark.

However, choosing a single day (eg https://www.inaturalist.org/observations?on=2020-07-04&user_id=alexis_orion) works.

todtb:
I can start looking into this @tiwane. Maybe something with the params, do you have an example of one that works?

boucur:
I think `on=` now only works with exact dates, so instead of being `on=2020` or `on=2020-09`, it'd need to be `year=2020` and `year=2020&month=9`.

todtb:
Thanks! I'll start looking into it.

todtb:
Hey @kueda, when picking this up I thought it would be a simple date parsing issue with the params, but it looks like after shifting from `by_login` to `index` with user_id param, it works significantly different. 

While I've dug around a little bit and come up with some solutions for making OpenGraph work with partial iso8601 strings, (a year or month of obs), the best approach to the issue I believe needs some Angular knowledge, of which I lack. While I could keep poking at it, since this bug is now affecting many links from 5a3567bc383cb454890e08111370f528bee86dd3, I'd hate to hold up a solution.

If you have an approach in mind, I'd be glad to put it in place, otherwise I think I'll defer the issue to the experts.

kueda:
> the best approach to the issue I believe needs some Angular knowledge

By this do you mean that the UI at, say, https://www.inaturalist.org/observations?on=2015-09, should show the user that they are viewing observations from September 2019? I wouldn't worry about that. For one thing, as long as `params[:on]` is set to something like `2015-09`, the "Exact Date" field should get populated with `2015-09`. On top of that, this kind of date filtering is pretty niche, so if the UI doesn't perfectly reflect the state of the search, I don't think that's a huge deal.

IMO, the priorities here are

1. A spec to ensure that these kinds of date params don't cause obs/index to crash
1. A fairly aggressive `begin` / `rescue` block to prevent any code exclusively dedicated to figuring out values for Open Graph from interfering with the primary purpose of showing observations
[...]

PR #2819, opened 2020-10-01
Handle malformed date params on obs index

todtb:
#2810 

This should gracefully handle any exceptions that come out of the OG description generation. I'm still concerned that these partial iso8601 params are not being parsed correctly for the obs retrieval being done on the front end, however let me know your thoughts. I suppose it's the difference of landing on the obs index with no results vs whale fail. `by_login` action still is parsing these dates correctly, so I think a similar approach could be taken by the angular front end.

FWIW I also have a WIP branch https://github.com/inaturalist/inaturalist/compare/main...todtb:2810-date-params-for-obs that would start to generate this description based on partial dates, but that I think is lower priority.

### DIFF
--- a/app/controllers/observations_controller.rb
+++ b/app/controllers/observations_controller.rb
@@ -100,9 +100,10 @@ class ObservationsController < ApplicationController
     # API methods, and HTML/views and partials
     if human_or_scraper && !showing_partial
       @shareable_description = begin
-                                generate_shareable_description
-                               rescue StandardError
-                                ""
+                                 generate_shareable_description
+                               rescue StandardError => e
+                                 Logstasher.write_exception( e, request: request, session: session, user: current_user )
+                                 ""
                                end
     else
       h = observations_index_search(params)
--- a/app/controllers/observations_controller.rb
+++ b/app/controllers/observations_controller.rb
@@ -100,11 +100,11 @@ class ObservationsController < ApplicationController
     # API methods, and HTML/views and partials
     if human_or_scraper && !showing_partial
       @shareable_description = begin
-                                 generate_shareable_description
-                               rescue StandardError => e
-                                 Logstasher.write_exception( e, request: request, session: session, user: current_user )
-                                 ""
-                               end
+        generate_shareable_description
+      rescue StandardError => e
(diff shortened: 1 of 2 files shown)


app/controllers/observations_controller.rb
```diff
@@ -99,41 +99,11 @@ def index
...
-          I18n.t( :observations_by_user, user: search_user.try_methods(:name, :login) )
-        end
-      else
-        I18n.t( :x_site_is_a_social_network_for_naturalist, site: @site.name )
-      end
+      @shareable_description = begin
+                                generate_shareable_description
+                               rescue StandardError
+                                ""
+                               end
```
kueda, line 106: Could you a) fix the indentation here, and b) put `Logstasher.write_exception( e, request: request, session: session, user: current_user )` in the `rescue` block so if we *are* having a lot of problems with generating the shareable description it will at least get logged?

kueda (changes requested):
One more thing: not sure what the best etiquette is for requesting minor changes like this. Would it be better to just make them myself in your branch or does that feel intrusive?

app/controllers/observations_controller.rb
todtb: Sure, coming right up. Also, my fork is purely for contributing, so if there's something you'd like to get a quick change to, go right ahead! Doesn't bother me at all, but thanks for checking 😄 

One thing I noticed on some of the older files is that there is trailing whitespace on many lines so I turned off the auto correct indentation/whitespace for the project, so sorry I missed this one. 

In the future, if I touch a file with say one line change and a hundred minor whitespace changes, is that a problem?

kueda: Not a major problem but also not the best practice for record-keeping / git blaming. I do similar stuff when I try to correct JS for eslint compliance, but I try to keep those to single commits so it's clear I wasn't trying to change any functionality. In the ruby code, I generally try to leave things alone unless I'm editing that line anyway.

kueda (approved):

todtb:
Ahh, _that_ indentation. Noted and saved to project config.

Issue #2822, opened 2020-10-02
On User Curation Page, add link to user's profile and projects pages


tiwane:
At help@inat, people often don't send their usernames and thus I have to use /admin/users to find their account. From there two of the pages I need to get to the most are the user's profile page or their projects page. Can we add links to those pages on the right-side nav bar?

![image](https://user-images.githubusercontent.com/15896181/94955651-c10b8180-049f-11eb-9636-ea209707b6fb.png)

PR #2823, opened 2020-10-04
Add profile and project links to user curation page

todtb:
#2822 

Adds links to the user curation page to view profile and a user's projects. 

The glance section didn't seem like the right fit for the profile link so it's up top as shown:
Screen Shot 2020-10-04 at 9 35 56 AM

Hope this helps @tiwane


tiwane:
That's fine for me. I suggested the profile link be added to the "glance" (now I know what that's called!) because it seems analogous to the this subnav bar: https://www.inaturalist.org/home

image

But as long as the link's there, I don't have strong feelings.

Issue #2820, opened 2020-10-01
Comment body / ID remark no longer persisting when switching tabs on obs detail


kueda:
**Describe the bug**
On the obs detail page, we used to ensure the text of the textarea for the comment form and the identification form would stay the same when switching between the forms (i.e. clicking on the different tabs), but that doesn't seem to be working any more.

**To Reproduce**
1. Go to any observation detail page
1. Enter some text in the comment form
1. Click on the "Suggest an Identification" tab

**Expected behavior**
The identification textarea will be blank, when it should have the text from the comment form

**Context**
 - OS: Mac OS 10.15.6
 - Browser: Brave 1.14.84

todtb:
I can give this a go. I'm going to try converting from jQuery DOM manipulation to state based, with a ref on the parent `Activity` component to find and update the child `TextEditor`

PR #2828, opened 2020-10-08
Sync comment-id panel editor text

todtb:
#2820

I think this should take care of syncing these by using react state instead of DOM manipulation. I'm still having weird issues in the activity feed locally(?) where I go to edit or delete an identification or comment and it does not process. I think it may not be dispatching correctly, but I'm not too familiar with Redux and how these ducks are working.

Is this something you've observed in a dev environment @kueda? Existing edit/delete functionality were the only things I couldn't really test properly due to this odd behavior.


kueda:
Hey Brian. Is this working for you? B/c it isn't for me. A cursory look would suggest that your changes to `editorContent` might be getting passed to the `<TextEditor>` components but those changes aren't updating the state of those components.

> I'm still having weird issues in the activity feed locally(?) where I go to edit or delete an identification or comment and it does not process.

This sounds unrelated, but can you describe the problem in more detail, or provide repro steps? I haven't had any problems deleting comments or idents in production, FWIW.

todtb:
Turns out I had some other local issues going on that were confusing my efforts, including inat api still running `master`. 

Regarding the issue updating, it's still there though.

For me, reproducing is simple. Any attempt to update the comment body is immediately reverted. That is, `SET_ATTRIBUTES` sets the new state (which I can observe in Redux browser tools). However, no call is made to the API. Instead `SET_OBSERVATION` runs and reverts the state and thus the text.

I did notice this: 
`app/webpack/observations/show/ducks/observation.js:514` catches an exception `Error: id required`. Since the payload is just `UUID: "xxxx-xxxx-xxxx", comment: { body: 'text' }` I think that may have something to do with it. 

Is there something I need to setup locally as part of the ongoing transition to APIv2 where UUID has replaced ID? I've tried setting the `config.yml` in rails and `config.js` in node to `v2` but then existing calls do not work such as loading the obs itself.

todtb:
Marked as WIP as this is also affecting obs modal in "Identify"

kueda:
> Is there something I need to setup locally as part of the ongoing transition to APIv2

Maybe. Make sure iNaturalistAPI is up to date and using the main branch, and that the inaturalistjs dependency is up-to-date in the Rails and node repos. When I successfully update a comment locally the payload looks the same (except `uuid` is lowercase, not sure if yours is really in caps and if so how it got that way).

todtb:
So I've got this managed in the state in both places now in a similar fashion. I'm pretty new to redux so any and all feedback is much appreciated. Thanks!

I'll check to make sure Rails is using the up-to-date inatjs too just in case that's the issue.

kueda:
This is working for me and storing the content in Redux is probably the right way to go.

My remaining critique, which doesn't need to be blocking, is just that I think your duck should be named `text_editor.js` or `text_editor_content.js` instead of `update_text_editor.js` and should probably be refactored to handle the state of multiple text editors in a single React application, even if that's not something we currently need. Alternatively, you could store this data using the `config` reducer, which we often use for one-offs like this, though if we do want to support multiple text editors like this on a single page, a new reducer is probably more appropriate. Re: naming, IMO, a reducer manages a piece of the state which is generally something noun-y, whereas individual actions should be verb-y.

So I guess at a minimum, please update the name of that file and I think we should be good to go.

todtb:
Hey @kueda, I pluralized the duck name as it now tracks the editors based on a provided key. Let me know what you think.

kueda:
Looking good. Thanks for the updates! We'll get this deployed soon.

PR #2854, opened 2020-10-30
Use Wikidata to fetch the correct Wikipedia taxon page

kueda:
Specifically, this uses HEAD requests to https://hub.toolforge.org to map the iNat production taxon ID to a Wikipedia URL in a specific locale, so even if that language's Wikipedia doesn't have a page corresponding to that name, we can still show Wikipedia content.

Note that this starts storing the page content *and* the URL in a 24-hour cache, but doesn't cache requests to the hub service with an `ApiEndpointCache` like we do for some other services, b/c this service doesn't actually return a request body to cache.

The hub service is hosted by Wikimedia so presumably it won't stress them out too much if we hit it a lot, but if it does we'll need to start pre-caching these data from Wikidata. Some notes on how best to get this data from Wikidata at

https://forum.inaturalist.org/t/use-wikidata-to-link-to-appropriate-wikipedia-articles-in-all-languages/5538/22
https://forum.inaturalist.org/t/use-wikidata-for-place-names-and-wikipedia-descriptions/7702

### DIFF
--- a/app/assets/stylesheets/taxa/show/wikipedia.scss
+++ b/app/assets/stylesheets/taxa/show/wikipedia.scss
@@ -16,27 +16,43 @@
   table.infobox, 
   .infobox_v2,
   table.infobox_v2,
-  .thumb, 
-  .tright {
+  .tright,
+  .tleft {
     float: right;
     clear: right;
-    margin-left: 20px;
-    margin-bottom: 1em;
     background-color: #eee;
-    padding: 1em;
+    border-left: 20px solid white;
+    border-bottom: 20px solid white;
+    padding: 10px;
   }
 
   ul.gallery {
     list-style: none;
     display: flex;
     flex-wrap: wrap;
+    padding: 0;
   }
 
-  .gallerybox .thumb {
-    float: none;
-    margin: 0;
-    margin-bottom: 0.6em;
-    padding: 0;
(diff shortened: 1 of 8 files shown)


kueda:
I think this could be improved by moving the Wikidata property name into config so we can get rid of the `production?` blocks, but this works for now.

maxlath:
Coming here a bit late, but note that if you want to get a request body from hub.toolforge.org, you can set `format=json` in the querystring, and get the Wikipedia article URL in the response body at `.destination.url`. Example: https://hub.toolforge.org/P3151:506262?lang=en&format=json

Moreover, in case there is no article in Wikipedia for the desired language, the URL will be one of another Wikimedia project in the requested language, if available. Example: https://hub.toolforge.org/P3151:506262?lang=en&format=json

If you would actually prefer to get a URL on a Wikipedia in another language (rather than a URL from another Wikimedia project), you can specify `site=wikipedia`. Example: https://hub.toolforge.org/P3151:506262?lang=en&site=wikipedia&format=json

kueda:
Good to know, @maxlath. Our current approach is working for us right now, and we haven't heard from anyone at hub.toolforge.org suggesting that we're sending them too much traffic, but if any of those things should change, we can use your suggestions to start caching the responses.

Issue #2498, opened 2019-12-03
Order Locale pop-up on Account Settings page to match footer language chooser


tiwane:
From https://forum.inaturalist.org/t/change-order-of-languages-in-your-account-settings/8375

Currently the order of locales in the Locale chooser on the Account Settings page is determined by one's current locale. So if my Locale is Spanish, the languages are ordered by their names in Spanish, eg:

<img width="282" alt="image" src="https://user-images.githubusercontent.com/15896181/70093826-1522ac80-15d6-11ea-8b5e-6c09489addeb.png">

This is a bit confusing and disorienting for multi-lingual users, so it's been requested that the Local chooser on the Account Settings page work more like the footer language chooser, wherein each language is displayed in its own name/characters, and the order is static, eg:

<img width="197" alt="image" src="https://user-images.githubusercontent.com/15896181/70094002-70ed3580-15d6-11ea-9e69-43506fe6da50.png">

PR #2865, opened 2020-11-13
User settings UI and controller/routes for blocking users

albullington:


app/webpack/users/edit/components/relationships.jsx
```diff
@@ -0,0 +1,205 @@
...
+  relationships,
+  filteredRelationships,
+  updateFilters,
+  sortRelationships,
+  showModal,
+  loadPage,
+  page
+} ) => {
+  const showRelationships = ( ) => filteredRelationships.slice(
+    page * 10 - 10, page * 10 // only display up to 10 items on current page
```
kueda, line 21: Does this mean API pagination isn't working? Theoretically you should be able to make it 10 per page with `per_page=10` in your request.

app/webpack/users/edit/components/relationships.jsx
```diff
@@ -0,0 +1,205 @@
...
+                name="trust"
+                label={I18n.t( "trust_with_private_coordinates" )}
+                id={user.id}
+              />
+              {user.reciprocal_trust && <dfn>{I18n.t( "user_trusts_you_with_their_private_coordinates", { user: friendUser.login } )}</dfn>}
+            </div>
+            <div className="col-md-6">
+              <dfn className="relationship-row-margin">
+                {`${I18n.t( "added" )} ${moment( friendUser.created_at ).format( "LL" )}`}
+              </dfn>
```
kueda, line 49: Have to admit I need to look up the `<dfn>` tag. How is the date added a definition?

Also, it would be better to use the `added_on_datetime` translation key or to add a new one like `added_datetime` so a) translators can translated added" in a way appropriate to the context, and b) translators can control word order.

app/webpack/users/edit/ducks/relationships.js
```diff
@@ -0,0 +1,332 @@
...
+}
+
+export function setPage( page ) {
+  return {
+    type: SET_PAGE,
+    page
+  };
+}
+
+export function filterRelationships( ) {
```
kueda, line 98: Not sure where to put this so I'll put it here: are you only filtering here in the client and not making a new API request? I think that's what's happening, which means if you have a hundred relationships, the filters you apply will only get applied to the 10 relationships on the first page. If I'm misreading this, my apologies. Local dev env is hosed atm so I can't quite test as thoroughly as I'd like.

app/webpack/users/edit/components/relationships.jsx
```diff
@@ -0,0 +1,205 @@
...
+        </div>
+      </div>
+    </div>
+  );
+
+  const showEmptyState = ( ) => (
+    <SettingsItem>
+      <p className="bold">
+        {I18n.t( "youre_not_following_anyone_on_inat", { site_name: SITE.name } )}
+      </p>
```
kueda, line 136: This `bold` class is probably unnecessary, and could be replaced here with the slightly more semantic

```html
<p>
  <strong>blah</strong>
</p>
```

app/webpack/users/edit/components/relationships.jsx
```diff
@@ -0,0 +1,205 @@
...
+        <BlockedMutedUsersContainer
+          headerText={I18n.t( "blocked_users" )}
+          id="blocked_users"
+          placeholder={I18n.t( "add_blocked_users" )}
+          buttonText={I18n.t( "unblock" )}
+          htmlDescription={{
+            __html: I18n.t( "views.users.edit.blocking_desc_html", {
+              site_name: SITE.name,
+              // noting that this help_email isn't populating
+              help_email: "help@inaturalist.org" // SITE.help_email
```
kueda, line 177: Use `SITE.help_email` anyway. You can configure it locally at http://localhost:3000/sites/1/edit, but in production, different sites may have different help email addresses. If we want to ensure people are using *our* help address, we can do something later like `DEFAULT_SITE.help_email`.

app/webpack/users/edit/components/settings_item.jsx
```diff
@@ -0,0 +1,30 @@
...
+  required,
+  children,
+  htmlFor
+} ) => (
+  <div className="row settings-item">
+    <div className="col-xs-12">
+      {( header && htmlFor ) && (
+        <label htmlFor={htmlFor}>
+          {header}
+          {required && " *"}
```
kueda, line 15: This is fine, but you could make a more RTL-friendly version by assigning a `required` class to the `<label>` element and then doing something in css like
```css
label.required:after {
  content: "*";
  margin-left: 5px;
}
```

While the margin-left is still RTL-unfriendly, it makes it easier to support RTL later with something like

```css
label.required:after {
  content: "*";
  margin-left: 5px;
}
.rtl label.required:after {
  margin-left: 0px;
  margin-right: 5px;
}
```

app/webpack/users/edit/components/toggle_switch.jsx
```diff
@@ -0,0 +1,28 @@
...
+        />
+        <span className="slider round" />
+      </label>
+      <label htmlFor={name}>{I18n.t( "on_toggle" ).toLocaleUpperCase( )}</label>
+    </div>
+  </div>
+);
+
+ToggleSwitch.propTypes = {
+  profile: PropTypes.object,
```
kueda, line 23: This component is great, but this prop is keeping it from being more generally useful. Maybe replace it with `checked`? The other thing we'd need to do is pull its styles out into a separate file that could be used elsewhere.

app/controllers/users_controller.rb
```diff
@@ -597,6 +597,7 @@ def edit
       format.html do
         @monthly_supporter = @user.donorbox_plan_status == "active" &&
           @user.donorbox_plan_type == "monthly"
+        render :edit2, layout:"bootstrap"
```
kueda, line 600: Let's change this to support testing without changing the current behavior just yet, e.g.

```ruby
if is_admin? && params[:test] == "profile"
  render :edit2, layout: "bootstrap"
else
  render :edit
end
```

kueda (changes requested):
Overall, things are looking great, and this is a herculean effort for your first web project. Not sure why I thought this would be straightforward.

This is a first pass based on reading the code, since my local env is too messed up to run things right now. I'll do another pass later. One thing I did notice when I was testing before my local apocalypse was that adding a profile pic showed a broken image and a JS error, but I think reloading the page actually showed the pic. Not sure if that's still happening.

app/webpack/users/edit/components/account.jsx
```diff
@@ -0,0 +1,173 @@
...
+  setModalState
+} ) => {
+  const createTimeZoneList = ( ) => (
+    TIMEZONES.map( zone => <option value={zone.value} key={zone.value}>{zone.label}</option> )
+  );
+
+  const createLocaleList = ( ) => {
+    const locales = I18n.t( "locales" );
+
+    const excludeLocalizedName = ["br", "en", "eo", "oc"];
```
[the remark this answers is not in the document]
albullington: Nope, this was me being overly literal while looking at how we do things on the current account settings page. I think what should actually happen here is that we either: 

1) show *all* names as English / English (language name in user's current locale / language name in original language) or 
2) show only the user's current locale with a single language name & no hyphen (which is less redundant than English / English)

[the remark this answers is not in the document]
kueda: Given https://github.com/inaturalist/inaturalist/issues/2498, I think there's a general preference for approach 2.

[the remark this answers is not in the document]
albullington: Okay, per #2498, fixed this to display each language in their own name/characters: 

Screen Shot 2020-11-13 at 5 11 24 PM

app/webpack/users/edit/components/relationships.jsx
albullington: Fixed this here and also in the third party tracking modal.

app/webpack/users/edit/components/content.jsx
```diff
@@ -0,0 +1,301 @@
...
+          {createRadioButtons( )}
+          <p />
+          <p className="text-muted">{I18n.t( "views.users.edit.project_settings_desc" )}</p>
+          <p
+            className="text-muted"
+            // eslint-disable-next-line react/no-danger
+            dangerouslySetInnerHTML={{
+              __html: I18n.t( "views.users.edit.this_only_applies_to_traditional_projects", {
+                url: "blog/15450-announcing-changes-to-projects-on-inaturalist"
+              } )
```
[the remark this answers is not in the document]
albullington: Oops. Fixed.

app/webpack/users/edit/components/profile.jsx
```diff
@@ -0,0 +1,133 @@
...
+
+    if ( profile.icon.preview ) {
+      return profile.icon.preview;
+    }
+    return profile.icon;
+  };
+
+  // this gets rid of the React warning about inputs being controlled vs. uncontrolled
+  // by ensuring user data is fetched before the Profile & User page loads
+  if ( profile.login === undefined && profile.email === undefined ) {
```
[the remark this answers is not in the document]
albullington: Fixed.

app/webpack/users/edit/components/profile.jsx
```diff
@@ -0,0 +1,133 @@
+import React, { createRef } from "react";
+import PropTypes from "prop-types";
+import Dropzone from "react-dropzone";
+
+import CheckboxRowContainer from "../containers/checkbox_row_container";
+import SettingsItem from "./settings_item";
+import ChangePassword from "./change_password";
+
+const emptyProfileImage = "https://www.inaturalist.org/attachment_defaults/users/icons/defaults/thumb.png";
```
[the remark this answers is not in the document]
albullington: Neat. I can change this, but I suspect what we actually want to do here is add an attribute to `users.me` called `icon_url`, per my code comments on lines 54-55. That would let us use the same shared `UserImage` component that we're using elsewhere. Right now there is only an `icon` attribute.

app/webpack/users/edit/components/relationships.jsx
albullington: The API pagination is working, but I couldn't figure out a good approach to be able to use the API pagination and also apply filters and sorting.

If a user sorts by A-Z, for example, I'm assuming the user would want to see the first ten results with names starting with A, B, C, etc. across *all* of their relationships, but if I'm using `per_page=10` I would only have enough information to apply that sorting for the 10 names currently on display.

I'm also making an assumption that if a user is on page 3 and chooses an additional sort/filter, he/she would automatically start back at page 1 to see the newest sorted/filtered results... but maybe I need more clarity on how the UI is supposed to work here.

app/webpack/users/edit/ducks/relationships.js
albullington: I'm filtering in the client, and it is applied to all relationships. I'm basically fetching all relationships, then storing the result in the `relationships` state, and using that list as the base of truth to apply filters, which avoids only filtering the 10 displayed relationships.

This code all felt awkward to write though, so I'm not surprised it's unclear to read. Open to suggestions and also curious if it would make more sense for the relationships endpoint to handle the sorting and filtering?

app/webpack/users/edit/components/profile.jsx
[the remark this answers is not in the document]
kueda: https://github.com/inaturalist/iNaturalistAPI/issues/224

app/webpack/users/edit/components/relationships.jsx
kueda: IMO, if the user changes the search, sort, or other filters, they should go back to page 1, and all of those should be performed in the API request, not the client. So if the user wants to sort by name, you would request `/v1/relationships?order_by=name&order=asc&page=1`.

That said, another approach might be to load all the user's relationships and then perform these operations in the client. For most users that will be fine. For people with a lot of relationships, there will be some cost in bandwidth and data held in memory.

app/webpack/users/edit/ducks/relationships.js
kueda: Hah, I probably should have read this before writing my other reply. FWIW, I don't think [fetchRelationships](https://github.com/inaturalist/inaturalist/pull/2865/files#diff-1f1c2c9fea07980531091f52d8b3d905b2db0fbf026a63ed3ffdf20091c33b31R210) is fetching all relationships, just the first page. And yes, I think the search / filter / page logic should be handled via the API. FWIW, the API is passing things through to the Rails app, which ends up getting handled in [`FriendshipsController#index`](https://github.com/inaturalist/inaturalist/blob/main/app/controllers/friendships_controller.rb#L8).

app/webpack/users/edit/components/account.jsx
```diff
@@ -0,0 +1,173 @@
...
+        eventKey={number}
+        className="inat-affiliation-width"
+      >
+        <span className="flex-no-wrap">
+          <img
+            className="inat-affiliation-logo margin-right-medium"
+            alt={`inat-affiliation-logo-${number}`}
+            src={showINatAffiliationLogo( number )}
+          />
+          <div className="text-muted small">{location.toLocaleUpperCase( )}</div>
```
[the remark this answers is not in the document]
albullington: Removed `.toLocaleUpperCase( )` here and also switched to using translation strings appended with `_caps` in the other places where I was using that method (save button, remove relationship modal, toggle switch). I'm surprised I've never had complaints about this in Seek!

app/webpack/users/edit/components/account.jsx
```diff
@@ -0,0 +1,173 @@
...
+  { number: 6, location: I18n.t( "places_name.colombia" ) },
+  { number: 8, location: I18n.t( "places_name.portugal" ) },
+  { number: 9, location: I18n.t( "places_name.australia" ) },
+  { number: 13, location: I18n.t( "places_name.panama" ) },
+  { number: 14, location: I18n.t( "places_name.ecuador" ) },
+  { number: 15, location: I18n.t( "places_name.italy" ) },
+  { number: 16, location: I18n.t( "places_name.argentina" ) },
+  { number: 18, location: I18n.t( "places_name.chile" ) },
+  { number: 20, location: I18n.t( "places_name.finland" ) }
+];
```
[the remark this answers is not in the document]
albullington: Fixed this by fetching from the API, which means we'll need to add the image URLs to this API per your comment below.

app/webpack/users/edit/components/content.jsx
```diff
@@ -0,0 +1,301 @@
...
+          <div>
+            <DropdownButton
+              id="preferred_observation_license"
+              onSelect={e => handleCustomDropdownSelect( e, "preferred_observation_license" )}
+              title={showDefaultLicense( profile.preferred_observation_license )}
+            >
+              {createLicenseList( "preferred_observation_license" )}
+            </DropdownButton>
+          </div>
+          <p />
```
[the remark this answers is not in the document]
albullington: Switched to `.stacked` here and elsewhere in the user settings code.

app/webpack/users/edit/components/menu.jsx
```diff
@@ -0,0 +1,32 @@
...
+  I18n.t( "applications" )
+];
+
+const Menu = ( { menuIndex, setContainerIndex } ) => menuItems.map( ( item, i ) => (
+  <div key={item}>
+    <button
+      type="button"
+      name={i}
+      id="LeftNav"
+      className={`left-nav-item ${menuIndex && "green"}`}
```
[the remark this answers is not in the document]
albullington: Fixed.

app/webpack/users/edit/components/applications.jsx
```diff
@@ -0,0 +1,76 @@
+import React from "react";
+import PropTypes from "prop-types";
+import moment from "moment";
+
+import SettingsItem from "./settings_item";
+
+const iNatAppIds = [2, 3, 315, 333];
```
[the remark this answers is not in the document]
albullington: Slight preference for `official` but I'm good with either one.

app/webpack/users/edit/components/relationships.jsx
albullington: Switched to `<em>` and the full sentence `added_on_datetime` translation key.

app/webpack/users/edit/ducks/relationships.js
albullington: I think I have everything working from the API except for search. Does [this line](https://github.com/inaturalist/iNaturalistAPI/blob/2dc16cb6357f15555c038343fccdf707be3361a6/lib/controllers/v1/relationships_controller.js#L30) mean that I should be able to request something like `/v1/relationships?q=kueda` to search user logins? (When I try adding that query I'm getting a 500 error).

kueda: Sorry about that, should be fixed in https://github.com/inaturalist/iNaturalistAPI/commit/05c6582a1d569eb1a9a0d098bfe9ffa91bffd860

app/webpack/users/edit/components/relationships.jsx
albullington: Fixed.

app/webpack/users/edit/components/settings_item.jsx
albullington: Updated as follows to incorporate the red asterisk in the Zeplin designs & avoid making the label deeppink. 

```
label.required {
  color: #333;
  &:after {
    content: "*";
    color: #ca5c53;
    margin-left: 5px;
  }
}
```

app/webpack/users/edit/components/toggle_switch.jsx
albullington: Replaced with a `checked` prop and moved styles into `users/_toggle_switch` file.

app/webpack/users/edit/components/settings_item.jsx
kueda: You might want to use [$alert-red](https://github.com/inaturalist/inaturalist/blob/main/app/assets/stylesheets/colors.scss#L12) here.

app/webpack/users/edit/components/about_licensing_modal.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+  return (
+    <Modal
+      show={show}
+      className="AboutLicensingModal"
+      onHide={onClose}
+    >
+      <Modal.Body>
+        <h4>{I18n.t( "creative_commons_licenses" )}</h4>
+        <ModalCloseButton onClose={onClose} />
```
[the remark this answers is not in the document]
albullington: Per the Zeplin convo, updated all four modals to use a header and close button similar to the one shown when a user clicks the question mark above the seasonality chart on taxon detail.

app/webpack/users/edit/components/applications.jsx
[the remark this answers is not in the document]
albullington: Fixed I believe, but I couldn't fully test this. When I go to `http://localhost:3000/oauth/applications/18/edit` to edit my sample app and toggle the `official` box in the admin section, that field doesn't seem to be saving or staying checked.

app/webpack/users/edit/components/profile.jsx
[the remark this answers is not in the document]
albullington: Updated this code to use shared UserImage component.

app/webpack/users/edit/components/settings_item.jsx
albullington: Done (and swapped in `$inat-green`)

app/webpack/users/edit/ducks/relationships.js
albullington: Finished this -- search is working now.

app/webpack/users/edit/components/account.jsx
```diff
@@ -0,0 +1,173 @@
...
+        {I18n.t( `locales.${locale}` )}
+        {!excludeLocalizedName.includes( locale )
+          && ` / ${I18n.t( `locales.${locale}`, { locale } )}`}
+      </option>
+    ) );
+  };
+
+  const showINatAffiliationLogo = num => {
+    const pngAssetList = [2, 6, 8, 13, 14, 18];
+    return `https://static.inaturalist.org/sites/${num}-logo.${pngAssetList.includes( num ) ? "png" : "svg"}`;
```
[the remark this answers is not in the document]
albullington: Now fetching these through the sites inatjs endpoint.

app/controllers/users_controller.rb
albullington: Done, and unless I'm missing something else in the comments, I believe this should be the last fix.

app/webpack/users/edit/components/content.jsx
```diff
@@ -0,0 +1,297 @@
...
+        id="user_preferred_project_addition_by"
+        type="radio"
+        value={button}
+        name="preferred_project_addition_by"
+        checked={profile.preferred_project_addition_by === button}
+        onChange={handleInputChange}
+      />
+      <label className="margin-left" htmlFor="user_preferred_project_addition_by">
+        {radioButtons[button]}
+      </label>
```
kueda, line 42: Currently if you click the label text, it doesn't change the setting, because each `<input>` in the radio group has the same ID. They should have unique IDs and each label's `for` attribute should reference the ID of the input it's labeling. However, you could just ignore the id attributes by making the input a child of the label.

app/webpack/users/edit/components/applications.jsx
```diff
@@ -0,0 +1,74 @@
+import React from "react";
+import PropTypes from "prop-types";
+import moment from "moment";
+
+import SettingsItem from "./settings_item";
+
+const Applications = ( { showModal, apps = [] } ) => {
+  const renderEmptyState = ( ) => (
+    <p className="col-xs-9">
+      {I18n.t( "you_have_not_authorized_any_applications", { site_name: SITE.name } )}
```
kueda, line 10: You need to add this to `en.yml`, and I don't think it needs a `site_name` variable.

app/webpack/users/edit/components/revoke_access_modal.jsx
```diff
@@ -0,0 +1,49 @@
...
+    <Modal.Body>
+      <p>{I18n.t( "this_will_sign_you_out_current_session" )}</p>
+    </Modal.Body>
+    <Modal.Footer>
+      <div className="buttons">
+        <Button bsStyle="primary" onClick={deleteApp}>
+          {iNatAppIds.includes( id )
+            ? I18n.t( "log_out_caps" )
+            : I18n.t( "revoke_caps" )}
+        </Button>
```
kueda, line 35: Clicking this isn't closing the modal for me.

kueda (changes requested):
Ok, this review focused more on functionality. Couple bugs, couple UI issues.

app/webpack/users/edit/components/relationships.jsx
kueda: `<em>` is fine, but fwiw, if you need an element and there isn't a good semantic match, `<div>` and `<span>` are usually what I use. A reader (machine or otherwise) assuming the tags were being used with their implied meanings might interpret this as a particularly emphatic date, when it's really not more important than the other text that surrounds it. `<em>` also comes with some default visual styling (italic in most browsers) that you may not want and/or may have to override in CSS.

kueda (approved):

Issue #2870, opened 2020-11-19
New Lifelists: On right pane in Species view, have the number and label change when Show value changes


tiwane:
Currently in new lifelists, if the user chooses Species view, we show the number of leaves by default (red circle). But if the user changes the "Show" menu to another taxonomic level, eg Family (red arrow) the number and label don't change to reflect the new Show value:

![image](https://user-images.githubusercontent.com/15896181/99730397-37286f80-2a71-11eb-82c0-c4002a58312f.png)

Can we make the number and label reflect the taxonomic level chosen for Show?

todtb:
I can take a look at this.

todtb:
PR #2873 is open for a quick workaround. Also while working on this it took a little bit to set up some realistic taxonomic hierarchies. The readme has some info on importing some seed taxa, but do you have any tooling that you're using for creating realistic relational data @kueda or @pleary? For example a species with all its ancestral clades/ranks? If not, I can put together something that would build out a development seed set of species with associated Class, Order, Family, Genera, etc., but I figured I'd check first.

kueda:
Hey @todtb sorry for the late reply. I don't have any tooling like that. I usually rely on [`Taxon.import`](https://github.com/inaturalist/inaturalist/blob/main/app/models/taxon.rb#L1695), which will attempt to import a taxon by name and all its ancestors from external taxonomy providers, so you could set up a basic taxonomy with `Taxon.import("Vulpes vulpes"); Taxon.import("Quercus agrifolia"); Taxon.import("Amanita muscaria")`. However, maybe it would be more useful for development purposes to have a script like `tools/import-inat-taxon.rb` that accepts a name and imports a branch of the taxonomy from iNat itself.

pleary:
Fixed in #2873. Thanks again @todtb!

PR #2873, opened 2020-11-29
Change lifelist count based on taxa

todtb:
#2870 

Dynamically changes count based on taxa in new lifelist. The count was
originally calculated in the 'species_noapi' component. Now it, along
with the taxa filter labeling logic, are in a shared util function.

### DIFF
--- a/app/webpack/lifelists/show/components/details_view.jsx
+++ b/app/webpack/lifelists/show/components/details_view.jsx
@@ -123,7 +123,7 @@ const DetailsView = ( {
     }
   } else if ( lifelist.detailsView === "species" ) {
     title = I18n.t( "views.lifelists.all_species" );
-    let count = lifelist.detailsTaxon ? _.size( filteredNodes( lifelist ) ) : lifelist.leavesCount;
+    let count = _.size( filteredNodes( lifelist ) );
     searchLoaded = true;
     if ( lifelist.speciesPlaceFilter ) {
       inatAPIsearch = inatAPI.speciesPlace;
@@ -145,7 +145,8 @@ const DetailsView = ( {
       <div className="stats">
         <span className="stat">
           <span className="attr">
-            { I18n.t( "views.lifelists.observed_rank", { rank: rankLabel( lifelist.speciesViewRankFilter ) } ) }
+            { I18n.t( "views.lifelists.observed_rank",
+              { rank: rankLabel( { rank: lifelist.speciesViewRankFilter, withLeaves: false } ) } ) }
           </span>
           <span className="value">
             { searchLoaded
--- a/app/webpack/lifelists/show/components/details_view.jsx
+++ b/app/webpack/lifelists/show/components/details_view.jsx
@@ -8,6 +8,7 @@ import PlaceAutocomplete from "../../../observations/identify/components/place_a
 import Observations from "./observations";
 import Species from "./species";
 import SpeciesNoAPIContainer from "../containers/species_noapi_container";
+import { filteredNodes, rankLabel } from "../util";
 
 const ObservationsGridContainer = APIWrapper( "observations", Observations );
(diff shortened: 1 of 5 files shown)


app/webpack/lifelists/show/components/details_view.jsx
```diff
@@ -122,8 +123,7 @@ const DetailsView = ( {
     }
   } else if ( lifelist.detailsView === "species" ) {
     title = I18n.t( "views.lifelists.all_species" );
-    let count = lifelist.detailsTaxon
-      ? lifelist.detailsTaxon.descendantCount : lifelist.leavesCount;
+    let count = lifelist.detailsTaxon ? _.size( filteredNodes( lifelist ) ) : lifelist.leavesCount;
```
kueda, line 126: We still want the count to reflect the chosen rank when `lifelist.detailsTaxon` is blank. To see the problem,

1. Go to /lifelists/:username where :username is that of a user that has observations in several kingdoms
1. Change the rank filter to "Show: Kingdoms"
1. "Observed Kingdoms" shows the rank you chose, but the count is still the leaf count

kueda (changes requested):
Just the one bug as far as I can see. @pleary wrote all this, though, not sure if he wants to take a look or not.

app/webpack/lifelists/show/components/details_view.jsx
```diff
@@ -145,7 +145,7 @@ const DetailsView = ( {
       <div className="stats">
         <span className="stat">
           <span className="attr">
-            { I18n.t( "views.lifelists.observed_species" ) }
+            { I18n.t( "views.lifelists.observed_rank", { rank: rankLabel( lifelist.speciesViewRankFilter ) } ) }
```
pleary, line 148: One thing here that's a little confusing - we use the term "Leaves" in the dropdown menu, and this is the default value for taxa filtering, but we'd like to use the term "Species" in this label when filtering by leaves or species. This is a little weird, and technically wrong, but the term "leaves" is confusing for people and we're already conflating the two terms in other places. Since "Leaves" is the default, we'd rather the first thing people see is "Observed species". So if the rank chosen is either "leaves" or "species" the label should read "Observed species".

It's also a little strange to me when "Children" is selected with there's no details taxon to see "Observed children" with no context. But that's probably outside the scope of this ticket. I think the best thing to do is not offer "Children" as a selectable value if there's no details taxon, or get rid of "children" entirely - not sure if anyone uses it

app/webpack/lifelists/show/components/details_view.jsx
pleary: Agreed. I think we can skip the ternary and just go with `_.size( filteredNodes( lifelist ) )`

pleary (changes requested):
Thanks @todtb . I added a couple comments, but this mostly looks great. Thanks for picking it up and working through this pretty confusing and messy bit of code. I'm happy to see some of it extracted into a shared library.

app/webpack/lifelists/show/components/details_view.jsx
todtb: I think I understand @pleary. Originally I thought "leaves" were shorthand for subsp., variety, cultivar, etc. It seems like it's more like terminal nodes for the selected taxon? Operating on that assumption I think the latest commit should address it.

pleary: Yes leaves are terminal nodes in the tree. We go into a little [more detail here](https://www.inaturalist.org/pages/how_inaturalist_counts_taxa) if you're interested. Thanks - this change is looking great and is updating the labels as request

pleary:
This is looking ready to merge to me. Thanks @todtb for working on it. The build failure wasn't about code but about our new Github Actions workflow. I think it's related to the fact this is a pull request from a remote repo and we're expecting a Slack webhook secret, which you obviously don't have. I think I've updated the workflow to address that and skip Slack notifications for that case

PR #2903, opened 2020-12-27
Containerize

todtb:
After playing around with `Procfile`s and custom bash scripts for local dev, I took a go at containerizing the dependencies.

It's working pretty well, still a WIP but almost there, but curious what you think @kueda @pleary and if it's something you'd be interested in getting into the repo. If interested, I can put together some docs (it's pretty simple/automated though). If not, I can add these things to my local ignore files.

https://github.com/inaturalist/iNaturalistAPI/pull/234


Makefile
```diff
@@ -0,0 +1,15 @@
...
+
+clean :
+	docker-compose down --remove-orphans
+	docker image prune
+	docker volume prune
+
+services :
+	docker-compose build --parallel es memcached redis pg
+	docker-compose up -d es memcached redis pg
+	docker-compose -f ../iNaturalistAPI/docker-compose.yml -f ../iNaturalistAPI/docker-compose.override.yml up --build
```
todtb, line 12: This is the main part I'm concerned with and still treating as WIP as it makes assumptions on file structures. Curious what your thoughts are. Maybe there's a workaround, or maybe documentation around container caveats/requirements.

pleary:
Thanks @todtb - this is very interesting and something I'm sure some people will find very helpful. I tried out the Docker stuff here and in the API repo and I think I got everything mostly working OK. Here's a couple of things I found along the way and a few questions about how you think this can be used in development.

- I believe I needed to explicitly expose ports 11211:11211 for the memcached instance to connect to it
- The Makefile here assumes there is a docker-compose.override.yml file in the node repo, which doesn't exist by default, and I didn't have one, so I needed to modify this Makefile
- The pg container only creates a 'development' version of the database. Do you have an idea of how developers will be able to use that container for development as well as running specs against a test database? If people are using these to create code, they should ideally be able to run specs before committing code. That would mean we'd minimally need a test database. The test ES indices might get created automatically - I haven't tested that yet
- The data volumes appear to work nicely to preserve data, which I feel is necessary for development so devs don't need to constantly recreate test data. Nice!
- Do you think the Node API docker container is a necessary component here? I wonder if just the dependencies (pg, es, memcached, redis) should be containerized and devs would be responsible for running rails and node locally if they want to. That would remove the need to reference node in the Makefile and avoid the relative path issue you mentioned. It seems a little inconsistent that the node API is containerized but the rails repo isn't - that if you want the full stack and you're working on rails you don't need a local instance of the API running, but if you're working on the API you'd need a local instance of Rails running. I was also having issues getting the node container to reference some of these containerized repos, but I'm sure that's just a config issue on my end getting the right IPs to reference

Other than those things - this is looking pretty great and thanks for taking the time to set it all up. I've been having issues with my local database recently, and maybe I could move over to the containerized pg at some point. I don't think there's anything here that is a blocker, but it would be great to fix the memcached ports, and have some plan for how to run tests in each repo before merging things in.

kueda:
@todtb is this something you're still planning on working on? If not, I'll close. Or maybe you think this is worth keeping open if you want to switch to Docker for local development, Patrick?

todtb:
Sorry about that @pleary and @kueda, looks like the GH notifications got lost in the shuffle. I'll address some of that feedback and get some docs together on how to do the initial setup.

todtb:
Thanks for the feedback @pleary, sorry I didn't see this sooner!

> * I believe I needed to explicitly expose ports 11211:11211 for the memcached instance to connect to it

I'll add that in to be explicit

> * The Makefile here assumes there is a docker-compose.override.yml file in the node repo, which doesn't exist by default, and I didn't have one, so I needed to modify this Makefile

I'm planning on adding docs that will go over the initial setup. Included will be instructions to copy this over from the provided example, modifying any credentials. Note that `docker-compose.override.yml` is in the `gitignore` as it will contain the credentials required by pg

> * The pg container only creates a 'development' version of the database. Do you have an idea of how developers will be able to use that container for development as well as running specs against a test database? If people are using these to create code, they should ideally be able to run specs before committing code. That would mean we'd minimally need a test database. The test ES indices might get created automatically - I haven't tested that yet

This is definitely something I'd like to get in there as well. As it stands testing would still require a non-containerized DB, which isn't ideal to require both setups.

> * Do you think the Node API docker container is a necessary component here? I wonder if just the dependencies (pg, es, memcached, redis) should be containerized and devs would be responsible for running rails and node locally if they want to. That would remove the need to reference node in the Makefile and avoid the relative path issue you mentioned. It seems a little inconsistent that the node API is containerized but the rails repo isn't - that if you want the full stack and you're working on rails you don't need a local instance of the API running, but if you're working on the API you'd need a local instance of Rails running. I was also having issues getting the node container to reference some of these containerized repos, but I'm sure that's just a config issue on my end getting the right IPs to reference

I've started working on two make commands that will allow this as an added bonus.

pleary:
Thanks for the documentation updates and small fixes. Things are working for me so I'll merge this into master. Thanks again for your work on this - hopefully it helps other developers get up and running a bit quicker

todtb:
Thanks @pleary! If https://github.com/inaturalist/iNaturalistAPI/pull/234 looks good too, I think we'd be all set for running them together.

kueda:
@todtb this might merit an issue, but I'll post here for now: I was just trying to set things up with Docker on a wiped machine and realized that trying to install the `pg` gem fails because there's no `pg_config` in the PATH. Does the containerized version of postgres actually have a working version of `pg_config` on in the filesystem somewhere that would be accessible to Ruby?

I'm guessing none of us ran into this previously b/c we all generally have postgres installed.

todtb:
Oh interesting. Come to think of it, I haven't tried from a blank slate either. Unfortunately, I don't think the volume is accessible to the host directly (to point to with `--with-pg-config=/path` or something similar). Reading [here](https://michaelrigart.be/install-pg-ruby-gem-without-postgresql/) that maybe only `libpq` is required, but still, kind of disappointing to need it local just for the gem build. Maybe there's some work around or at minimum automation of this dependency.

todtb:
Hey @seanclifford, I noticed you put together some pretty cool debian/ubuntu setup scripts which includes the aforementioned dependency `libpq-dev` required for building the `pg` gem. Maybe that's a route that the developer docs could go in? @kueda what do you think?

seanclifford:
I've used [my ubuntu scripts](https://github.com/seanclifford/inat-dev-setup) to install from a fresh OS install with docker -- though to be honest I've already forgotten why I put half the things I did in there! (hence the reason to script it out)
Yes - I think installing `libpq-dev` was so I could get past installing the `pg` gem...

PR #2923, opened 2021-01-13
Monitor Google map creation

kueda:
This hits a [new logging endpoint](https://github.com/inaturalist/iNaturalistAPI/commit/558475671abf2de90323f712851ee814e5668959) most times we create a Google Map instance so we can investigate ways to cut down on use. This should monitor almost every Google Map on the site that goes through our custom `createMap` method in `map3.js.erb`, which means it's ignoring the location choose in the Uploader since that's using a third-party React component. The node endpoint is filtering by origin, which means it's denying all requests that don't come from inaturalist.org, so if the source of our extreme usage is on one of the partner sites we're going to miss that, but I figure this will at least get us started.

### DIFF
--- a/app/assets/javascripts/ang/controllers/observation_search.js
+++ b/app/assets/javascripts/ang/controllers/observation_search.js
@@ -1163,6 +1163,7 @@ function( ObservationsFactory, PlacesFactory, shared, $scope, $rootScope ) {
     if( $scope.map ) { return; }
     var defaultMapType = PREFERRED_MAP_TYPE || google.maps.MapTypeId.LIGHT;
     $( "#map" ).taxonMap({
+      placement: "observations-search",
       urlCoords: false,
       mapType: defaultMapType,
       showAllLayer: false,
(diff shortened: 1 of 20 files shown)


pleary:
This all looks good to me. I think we could add a call to `iNaturalist.log` in the `componentDidMount` method of the uploader location chooser if we wanted. Also, why are you authenticating these log calls with the application JWT when the node API isn't checking authentication? Seems like we could remove the auth part of the call, or we should be checking it on the API end. We probably don't need to authenticate these calls unless problems arise

kueda:
> I think we could add a call to iNaturalist.log in the componentDidMount method of the uploader location chooser if we wanted

Done.

> Also, why are you authenticating these log calls with the application JWT when the node API isn't checking authentication?

Correct me if I'm wrong, but I was trying to authenticate the calls with the user's JWT, not the application JWT. I'm doing that because I want to be able to discriminate between sign-in and signed-out map loads. If we're getting a lot of map loads from signed out people or from bots that might demand a different remedy.

pleary:
yep you're right. I wasn't looking closely and thought `inaturalist-api-token` was the app token, but it's a user token. I did see the API and logstash properly mark requests as signed in, so makes sense to leave it as you have it

PR #2940, opened 2021-01-26
Remove private positional accuracy

loarie:
yanked private positional accuracy from the db and added public positional accuracy to exports


kueda:
@loarie, IMO, this looks fine to me. If you can confirm that the migration was pretty quick I'll merge.

loarie:
yes was quick - that column is empty

PR #2957, opened 2021-02-11
Multiple photo buckets

pleary:


app/controllers/observations_controller.rb
```diff
@@ -2113,8 +2114,10 @@ def retrieve_photos(photo_list = nil, options = {})
         photo = if photo_class == LocalPhoto
           if photo_id.is_a?(Integer) || photo_id.is_a?(String)
             LocalPhoto.find_by_id(photo_id)
-          else
-            LocalPhoto.new(:file => photo_id, :user => current_user) unless photo_id.blank?
+          elsif !photo_id.blank?
+            lp = LocalPhoto.new( user: current_user )
+            lp.file = photo_id
+            lp
```
kueda, line 2120: This stuff isn't related to the bucket changing, right?

app/models/shared/license_module.rb
```diff
@@ -31,6 +31,10 @@ module Shared::LicenseModule
     CC_BY_NC_SA,
     CC0
   ]
+  ODP_LICENSES = [
```
kueda, line 34: Can we get a comment here explaining what ODP means (I assume Open Data Project, but would be nice to leave a note).

app/models/local_photo.rb
```diff
@@ -84,7 +88,76 @@ class LocalPhoto < Photo
...
+  def s3_permissions
+    LocalPhoto.s3_permissions( self.s3_account )
+  end
+
+  def set_s3_account
+    self.s3_account = self.is_public? ? "public" : nil
+  end
+
+  def is_public?
+    self.large_url ? !! self.large_url.match( LocalPhoto.s3_bucket( true ) ) :
```
kueda, line 137: Why `large_url` instead of `original_url`? I guess it doesn't matter, but `original_url` is theoretically the one that absolutely must exist.

app/models/local_photo.rb
```diff
@@ -84,7 +88,76 @@ class LocalPhoto < Photo
...
+
+  def s3_permissions
+    LocalPhoto.s3_permissions( self.s3_account )
+  end
+
+  def set_s3_account
+    self.s3_account = self.is_public? ? "public" : nil
+  end
+
+  def is_public?
```
kueda, line 136: I know I'm picking nits, but all iNat photos are public. Could we use `is_s3_public?` instead for clearer semantics (if more awkward typing)? Alternatives might be `odp_candidate?` or `odp_eligible?`

app/models/local_photo.rb
```diff
@@ -394,4 +467,112 @@ def extrapolate_dimensions_from_original
...
+      secret_access_key: s3_credentials[:secret_access_key],
+      region: LocalPhoto.s3_region
+    )
+  end
+
+  private
+
+  def self.move_to_appropriate_bucket( photo )
+    return unless CONFIG.s3_public_bucket
+    return unless p = LocalPhoto.find_by_id( p ) unless photo.is_a?( LocalPhoto )
```
kueda, line 483: What is `p` here? I'm surprised this runs with the two `unless` statements on one line.

app/models/photo.rb
```diff
@@ -147,6 +147,10 @@ def update_default_license
   def update_all_licenses
     return true unless [true, "1", "true"].include?(@make_licenses_same)
     Photo.where(user_id: user_id).update_all(license: license)
+    Photo.where(user_id: user_id).pluck(:id).each do |photo_id|
```
kueda, line 150: Have you confirmed this will run before a request times out if a person has like 100k photos?

kueda (commented):
I don't see anything blocking here, though I feel like some renaming and comments might be in order to clarify what's happening, e.g. things aren't toggling between a visible "public" state and a restricted "private" state.

app/controllers/observations_controller.rb
pleary: I think this was related to a previous approach I tried. I've reverted this

app/models/shared/license_module.rb
pleary: I added a comment. Feel free to modify or expand on it

app/models/local_photo.rb
pleary: I just tend to avoid original in most bases, but I agree that is the canonical version. I've updated it

app/models/local_photo.rb
pleary: I made a change, but I'm open to other changes. I have used "public" a bunch, such as in the s3 config parameters, which maybe isn't the best. I'm open to any naming changes. We could go with odp across the board if we like that better

app/models/local_photo.rb
pleary: This was just me flubbing a refactor of this method. I think I've fixed this now

app/models/photo.rb
pleary: I haven't timed it, and I agree it's a liability. I heard we like delayed jobs, so I made a delayed job that will create these delayed jobs w/o affecting web response time

pleary:
thanks for the comments. I made an attempt a fixing the various issues

app/models/photo.rb
kueda: Sing a sea shanties and down some Tide Pods and we'll be at web scale. Something to keep an eye on during testing, I guess.

app/models/local_photo.rb
kueda: Thanks. I think `is_in_public_s3_bucket?` expresses its meaning more clearly, so that works for me. FWIW, I think it's usually preferable to avoid verbs in boolean methods like this. It looks the same in conditionals (e.g. `if in_public_s3_bucket?`) and better in rspec (e.g. `expect( photo ).to be_in_public_s3_bucket?`)

Issue #2738, opened 2020-07-15
Selecting all and deleting doesn't update the remaining chars indicator in TextEditor


kueda:
**Describe the bug**
If you add enough text to make the remaining characters indicator visible in a Text editor and then use a cursor or a keyboard shortcut to delete all the text, the indicator doesn't change and remains visible when it should disappear or at least reset to 0.

**To Reproduce**
1. Go to the obs detail page
1. Start adding a comment and paste in over 4000 characters. You should see the character indicator
1. Use CMD-A or CTRL-A to select all the text and DELETE to delete it all
1. The indicator remains visible with the previous character count even though the textarea is empty.
1. See error

**Expected behavior**
The TextEditor should reset to the state it was in initially when there was no text.

**Context**
 - OS: Mac
 - Browser Brave (Chromium)
 - Version 1.10.97 Chromium: 83.0.4103.116 (Official Build) (64-bit)

ThamiMemel:
I am interested by the project and i want to contribute and i think this is a good place for me to start.
i spent some time getting around the source code and i think the reason for it is the **if condition** in  **inaturalist/app/webpack/shared/components/text_editor.jsx** 

`
      textareaOnChange = e => {
        if ( e.target.value.length > showCharsRemainingAt ) {
          this.setState( { textareaChars: e.target.value.length } );
        }
      };
`
but i am not sure yet, i will setup a development environment later on today to try fixing it.

PR #2963, opened 2021-02-15
Fixes char counter not behaving as expected

rawcomposition:
Fixes https://github.com/inaturalist/inaturalist/issues/2738 where the comment textarea character counter would not disappear if the text was deleted with cmd+a then backspace. Furthermore, the counter wasn't even accurate if only a small amount of content was typed after deleting the original text.


kueda:
Thanks!

PR #2969, opened 2021-02-20
Index translated place names

### LINKED ISSUE #2951: Index translated country names
We've allowed country names to be translated for some time now, so if I'm a French speaker and type in "United States", I'll see the name displayed in French:

image

But, I would have to know the English name for "United States", as translated country names have not been indexed. If I type in "etats-unis", "United States" doesn't appear as a result. If we could index these translated names that would help users find a country in their own language.

kueda:
This adds a nested object to documents in the place index that uses the same fields we're using in the taxon index. This will allow our existing search API to search on these fields without modification, so people search for "china," "中国," or "Китай" will all be able to find the country China.

Not 100% sure if this is the right way to add a nested field to the index, though.

Closes #2951

### DIFF
--- a/app/controllers/photos_controller.rb
+++ b/app/controllers/photos_controller.rb
@@ -1,6 +1,6 @@
 #encoding: utf-8
 class PhotosController < ApplicationController
-  before_filter :load_photo, :only => [:show, :update, :repair, :destroy, :rotate]
+  before_filter :load_record, :only => [:show, :update, :repair, :destroy, :rotate]
   before_filter :require_owner, :only => [:update, :destroy, :rotate]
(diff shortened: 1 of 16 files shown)


pleary (approved):
Looks good to me. The index schema changes seem right and it ran successfully for me locally to create nested properties.

In recent years I have manually applied changes to the production index to avoid any potential timing issues since ES properties cannot be deleted (to avoid a case where the index modification fails for some reason but some process picks up the new code and writes to the index before the intended properties are in place). But this in theory should work since the migration should fail if there was a problem and the new code should roll back before any process uses it

PR #2972, opened 2021-02-24
Use compact notation instead of exponential notation in taxa charts

albullington:
Right now, everything is using whole numbers (i.e. 1M instead of 1.2M) for simplicity across locales. This means charts for higher level taxa like [Chordates](https://www.inaturalist.org/taxa/2-Chordata) will repeat 1M a few times along the y-axis, pretty similar to how those charts are currently repeating 1e+6.


config/locales/en.yml
```diff
@@ -974,6 +974,58 @@ en:
...
   community_moderation_settings: Community Moderation Settings
   community_supported_id: Community-supported Taxon?
+  compact_number_formatting:
+    x_thousand:
+      one: "%{count}K"
+      other: "%{count}K"
+    # Short notation representing a four-digit number, i.e. 1,000
+    x_ten_thousand:
+      one: "%{count}K"
+      other: "%{count}K"
```
kueda, line 984: I'm sure I missed some context about this in kickoff, a couple questions:

1. Can you provide some examples of why we need to pluralize these? We already have some of these under [`number.format.si`](https://github.com/inaturalist/inaturalist/blob/main/config/locales/en.yml#L7719) so if we don't need to pluralize we should probably build on those. If we do need to pluralize, then this separate set of abbreviations is probably gustified.
1. Why are you using the `one` and `other` and not `few` and `many`?
1. Isn't `x_ten_thousand` supposed to be for a five-digit number, e.g. 10,000? I think a lot of these context comments are off.

config/locales/en.yml
```diff
@@ -974,6 +974,58 @@ en:
...
+      one: "%{count}M"
+      other: "%{count}M"
+    # Short notation representing an eight-digit number, i.e. 10,000,000
+    x_hundred_million:
+      one: "%{count}M"
+      other: "%{count}M"
+    # Short notation representing a nine-digit number, i.e. 10,000,000
+    x_billion:
+      one: "%{count}B"
+      other: "%{count}B"
```
kueda, line 1004: I don't think I've ever seen a billion abbreviated as `B` in a chart. `G` for giga- seems much more common, as per https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes.

kueda (changes requested):
Haven't tried to run the code but already have some blocking questions / issues. In addition to my inline comments, instead of putting this logic in charts.jsx, can you put it in [`util.shortFormattedNumber()`](https://github.com/inaturalist/inaturalist/blob/main/app/webpack/shared/util.js#L185)?

albullington:
Assuming we're going for a more localized, less standardized number than SI notation, I moved the code into `util.compactDecimalFormattedNumber()` (because that's what it's called on [cldr.unicode.org](http://cldr.unicode.org/translation/numbers-currency/number-patterns)). I also fixed the comments in `en.yml`.

There aren't too many languages that need pluralization, but I have two examples. Arabic uses a different abbreviation for thousands [starting from 3000](https://github.com/unicode-cldr/cldr-numbers-modern/blob/cfc3fa31ad5d1539af6985048313d290ddee0d38/main/ar/numbers.json#L128), so I added `few`. Macedonian also seems to use a [different abbreviation](https://github.com/unicode-cldr/cldr-numbers-modern/blob/cfc3fa31ad5d1539af6985048313d290ddee0d38/main/mk/numbers.json#L79) for 100 billion and > 100 billion, which would fit into `other`. Those were the only pluralization examples I found out of all the iNat locales (though I couldn't find Breton, Esperanto, Occitan, or Luxembourgish).

kueda (changes requested):
Can you put that info about what `few` and `other` mean in this context into comments on the relevant strings? If I find it confusing, translators will definitely find that confusing. Unfortunately, I think that means repeating the same comment on all the relevant translations if we want it to actually show up on Crowdin.

albullington:
Alright, I revamped this one to use `Intl.NumberFormat` with the relevant polyfills for Safari compatibility. Hopefully, this makes the code much simpler and avoids a lot of potential translator confusion.

kueda (approved):

kueda:
This is looking good and seems to be working for me in multiple browsers and multiple languages. There is a trade-off in this approach in that by incorporating all these polyfills at compile time we are forcing all users to download all translations. Hopefully that doesn't increase the size of the files too much since these translations should be pretty small. Another things we're trading for simplicity of implementation and maintenance is control over support for languages (if there isn't a polyfill for a locale we choose to support... we just won't be able to support it). IMO, both trades work making right now.

PR #2986, opened 2021-03-16
Fix without annotation filters

albullington:


kueda (changes requested):
Alas, you jumped into a complicated one here, Amanda. The change you made will change the URL and the filters, but it won't do what is expected, which is show observations without a specific term/value combination. Instead, this will show observations without a term, regardless of value.

This is kind of UI problem, kind of an API problem: the UI problem is that the "Without Annotation" UI seems like it's the inverse of the "With Annotation" UI, but it's not: in "With Annotation" you can refine the results by choosing a term, and further refine those results by choosing a value, but for "Without Annotation" choosing a term is actually a narrower search than choosing a term AND and value.

The API confusion is that `without_term_value_id` won't even work *unless* you specify `term_id`. That's because values can be associated with many terms, e.g. if the value is "yes" it could be associated with terms like "Looks like Jeff Goldblum: yes/no" or "Ate part of my leg: yes/no," so asking for observations without a specific value doesn't make sense except in the context of a term. I'm not convinced this is the best approach to the API: might be better to always specify term and value at once, e.g. with_anno=1-2 or without_anno=3-4, but that's not how it currently works.

I actually took a try at this issue a few weeks ago and gave up because it was too much for my brain to handle on a Friday.

kueda:
Just going to assume this one is permanently tabled and might depend on some updates to the API, and/or a different approach to the design of these controls.

PR #2991, opened 2021-03-26
Project details: lazy loading / code delivery when needed

albullington:
Performance update experiments: the goal here is to spread out when queries run on project details pages and only fetch data when it's needed. On page load, we're now fetching the members list, total counts for observations, species, identifiers, and observers, and the visible section of the leaderboard for umbrella projects.

Everything else is fetched either when a user clicks a relevant tab (in `componentDidMount`) or when a user scrolls to a relevant section of the page (via a callback function in the `LazyLoad` wrapper component from `react-lazy-load`). Users should see a loading wheel on the observations, species, and identifiers tab until the full results have been fetched.


pleary (commented):
Everything seems to be working well for me with these changes. A lot of the lazy loading will save queries at page load time and I think that should help make things feel faster for most users. The part that I'm hesitant about is splitting up requests for observations, observations, and species_counts into two parts - one with per_page=0 and one that fetches the first results, at least for collections and less so for umbrellas. The problem is we have the leaderboard panels pretty high up on the page and we need the first results of observations, observers, and species_counts to render those panels. Given that section is so high up, I suspect that more often than not that section will be rendered and we'll end up making requests for those data. I think it'll be overall more performant to skip the per_page=0 requests to fetch only the counts, and make single requests for those 3 things.

We can try keeping it the way it is for identifiers since we don't need the identifiers list for anywhere other than the identifiers panel, and people may not be clicking on that a lot. We might even be able to run some stats after this is released to see how often the identifiers panel is loaded. If for example it were loaded for every session, then it may make sense to skip the per_page=0 request for identifiers as well, but if it's rarely used then splitting up the requests makes sense.

For umbrella projects, since it doesn't have the same leaderboard as collection projects and fetches data differently for the umbrella stats, it's possible it would be overall more performant to keep the split requests. But I'm not sure that the per_page=0 query is going to be any faster than one that fetches results (on the server side - it obviously would send less data to the user). So I'm kind of torn on that and am slightly leaning to - let's just fetch the data with the counts and save the potential additional queries which may initiate large Elasticsearch aggregations.

Everything else is looking good though - loading the posts, quality_grades, map, etc is all working well and could save on some queries. Thanks for the work on this - I hope it's not too much of a pain to change how observations, observers, and species_counts are loaded (if we think making the change I suggested is a good idea)

albullington:
@pleary sounds like a good plan to me! I didn't realize when implementing `per_page=0` that it might end up putting double the load on the database... I was making an assumption that fetching counts was less intensive than fetching full results and/or that users might not click on as many tabs as we think they do.

I reverted all the code relating to fetching observations, species, and observers, and I kept identifiers as is so we can learn a bit about how often that panel is loaded / whether that improves or decreases performance.

kueda:
@pleary do you want to take another look at this or shall we merge?

albullington:
Also reverted the umbrella stats code since it's not actually a query that uses pagination & therefore should only be fetched on page load.

pleary:
I think everything is looking good right now and ready to merge. It would be great if counts were less costly to run than fetching results and we could do the two-step approach, but unfortunately that's not the case right now. Thanks for your work on this

PR #2999, opened 2021-04-02
Taxon details: lazy loading

albullington:
The important change here is waiting to fetch the History histogram until that tab is clicked.

Otherwise, I also added loading text to the Seasonality and History histograms and added lazy loading for the map, recent observations, discoveries, and wanted sections.


pleary:
This all looks great to me. I tested it locally and confirm that the map tiles and history chart requests requests aren't being made until needed, and the lazy-loaded sections seem to be rendering well. This should be a nice improvement for the taxon pages. Thanks!

PR #3003, opened 2021-04-06
Obs detail: lazy load annotations

albullington:
This keeps the annotations section closed by default for all users & only fetches controlled terms once the annotations section is clicked open. I added a loading state which probably won't be seen very often but seemed helpful for observations with a lot of data like [Gerald](https://www.inaturalist.org/observations/5890862), and I also moved the empty state underneath the annotations header (instead of not showing the annotations section at all) since the data fetch needs to happen before the empty state can be determined.


app/webpack/observations/show/components/annotations.jsx
```diff
@@ -11,9 +11,8 @@ import { termsForTaxon } from "../ducks/controlled_terms";
 class Annotations extends React.Component {
   constructor( props ) {
     super( props );
-    const currentUser = props.config && props.config.currentUser;
     this.state = {
-      open: currentUser ? !currentUser.prefers_hide_obs_show_annotations : true
```
pleary, line 16: I think what we decided on is the open state of the annotations panel should be closed by default, but should remain sticky based on user preferences. For example if I open the tab and reload the page it should stay open. I also think it makes sense to stay sticky for a session even if the user is not logged in.

For example if I'm not logged in and I go to a page that has observations, the section will be closed by default. Then if I open it and user that arrows at the top of the page to go to another observation and that observation also has annotations, the panel should still be viewable. If I leave and come back, it makes sense it would revert to closed. Whereas if I were logged in and came back it would remain open because that preference stuck.

That may complicate things a bit and make it so we need to manage the open state of the annotations panel outside the component and in the larger context of the obs page.

pleary (changes requested):
I noticed that if I load the page, open the annotations panel, then navigate to a different observation w/o reloading the page in the browser (e.g. use the arrows at the top to go to another observation) the panel will be closed and when opened it will say "No Relevant Annotations". So it seems like after navigating the annotations panel contents won't load. The heading seems to have the right count, but I don't see the annotations themselves.

If we could fix that, and ideally make the open state sticky, everything else seems to be working well as far as lazy loading the content before navigating

albullington:
@pleary thanks for explaining how the UX is supposed to work here -- I definitely didn't understand that the annotations panel was supposed to be sticky and didn't realize a user could navigate with the left and right arrows to different obs pages when I made this pull request.

I believe this is now working as expected -- the panel stays open or closed during a session whether the user is logged in or logged out, and it also remains sticky if a logged in user navigates away from the page and comes back. To achieve this, the open/closed state of the panel is now being stored in the `controlled_terms` reducer instead of in the annotations component. And the state of controlled terms is now being reset when a user navigates to a new observation page, so the annotations for that observation can be loaded correctly.

app/webpack/observations/show/ducks/controlled_terms.js
```diff
@@ -125,3 +157,12 @@ export function setControlledTermsForTaxon( taxon, terms = [] ) {
     dispatch( setControlledTerms( termsForTaxon( allTerms, taxon ) ) );
   };
 }
+
+export function fetchAnnotationsPanelPreferences( ) {
+  return ( dispatch, getState ) => {
+    const { config } = getState( );
+    const currentUser = config && config.currentUser;
+    const open = currentUser ? currentUser.prefers_hide_obs_show_annotations : false;
```
pleary, line 165: minor thing here - prefers_hide_obs_show_annotations should be inverted here. The panel should be open if there's a user and that user does NOT prefer to hide the annotations panel (it's admittedly a bad preference name)

pleary:
Aside from the inverted user preference, everything seems to be working well for me now. I can navigate different observations within the same session and the annotations panel open state stays sticky, and API requests are made only if needed. Looks to me like this can be merged in whenever. Thanks!

app/webpack/observations/show/ducks/controlled_terms.js
albullington: Good catch, inverted this.

jwidness:
Apologies for butting in, but the sticky setting is a little wonky for some workflows.  For example, if I go to Identify, then right-click and load an observation in a new tab, the Annotations panel is open.  Then I right-click and load another observation from Identify and the panel is closed.  Then I right-click a third observation from Identify and the panel is open again.  I haven't opened or closed the panel myself  at all, it just toggles open/closed on its own.  It only does this if you allow the obs detail page to load -- if you right-click and open 3 observations quickly without letting any finish loading, they're all in the same toggle state.

PR #3011, opened 2021-04-09
Prevent Ratatosk from importing taxon names without taxa

### LINKED ISSUE #2730: TaxonNames exist without taxa
**Describe the bug**
This is not good:
```
2.6.5 :015 > TaxonName.where( "taxon_id is null" ).count
 => 3104 
```
It's causing a crash on `TaxonNamesController#edit` when the name has a synonym that is missing a taxon. Some of these records were created in July 2020 so it's definitely a current, active bug.

**To Reproduce**
Go to https://www.inaturalist.org/taxon_names/1372094/edit, or if that's been fixed, search for TaxonName records with no taxon.

**Expected behavior**
TaxonNames should always have a taxon

loarie:
Closes #2730
if EOL names provider returns name with no rank,
if its a trinomial call the blank rank "subspecies" (as we currently call the blank rank of binomials "species")
otherwise stop trying to import the name


kueda:
FWIW, I think this looks fine, and I like that you're raising an exception if we don't get a rank. The more troubling finding is that TaxonNames can even be saved without an associated Taxon record. If you're saying that the `taxon` associate is there but it's invalid due to the missing rank and thus never gets persisted, maybe we need something like

```ruby
class TaxonName
  validates :taxon_id, presence: true
end
```
in addition to the existing validation ensuring the associate is present. That way we would ensure there's an object there *and* it's persisted in the database, at the cost of not being able to create and persist a TaxonName and its associated Taxon in one database transaction. That might break other things, though

kueda (approved):

loarie:
hm - yeah that broke things. when trying to save the taxon_name record with an associated taxon in memory (ie both have nil taxon_ids) that validation prevents the taxon_name from being saved. I suspect its because the taxon_name wants to save first and then the associated taxon? Is there any way to reverse the order and have it try to save the taxon first?

kueda:
> I suspect its because the taxon_name wants to save first and then the associated taxon? Is there any way to reverse the order and have it try to save the taxon first?

I suspect that's an accurate diagnosis. I don't know of a way to automatically reverse the order of saving, but you can always do it manually, e.g. instead of `tn = TaxonName.create!( name: "Greatest Lizard", taxon: Taxon.new( name: "Foo bar" ) )` you might ensure the taxon is saved first:

```ruby
t = Taxon.create!( name: "Foo bar" )
tn = TaxonName.create!( name: "Greatest Lizard", taxon: t )
```
If that change is too disruptive, though, feel free to just remove it. I think your original work addressed the issue as described.

loarie:
@kueda - there are several places where we used `valid?` to assess whether an unsaved TaxonName/Taxon pair was otherwise valid. Is the approach I took here of adding `:skip_taxon_id_validation` to get round the new taxon-has-id validation in those cases ok to you? If so I can fix the broken specs and test this a bit more

kueda:
> Is the approach I took here of adding :skip_taxon_id_validation to get round the new taxon-has-id validation in those cases ok to you?

I don't love that. The point of the validation is to ensure we never save a TaxonName without a persisted Taxon. If that causes problems, we should address them by ensuring the Taxon record is persisted first, not carving out exceptions to the validation.

That said, as I mentioned, this is somewhat tangential to the specific problem this PR was trying to address, so if you want to just ignore the validation we can merge this and it will probably solve that problem.

loarie:
The exception isn't to allow the record to be saved - just to allow `valid?` to return true. But I get that its a bit funky to rely on a `valid?` call that has one of the validations excepted.

That `valid?` use case is to check the uniqueness of the taxon name (ie to make sure there's no other taxon that already has that taxon name) by leveraging the `validates_uniqueness_of :name, scope: %i[parameterized_lexicon taxon_id], message: :already_exists, case_sensitive: false` taxon name validation. But if you don't like leveraging an excepted `valid?` call to do that I could just find another way to check the uniqueness of the taxon_name.

so bluntly - do you prefer:
a) keep it the way I have it here (still using `valid?` with the exception to check taxon name uniqueness)
b) find some other way to check taxon name uniqueness
c) remove the taxon name must have taxon_id validation and keep this simple

kueda:
I prefer 
> c) remove the taxon name must have taxon_id validation and keep this simple

loarie:
@kueda this is ready to merge

PR #3024, opened 2021-04-22
Has subscribers after commit

pleary:
Changes:

- has_subscribers uses after_commit callbacks instead of after_save, after_update, after_create to avoid long-running transactions
- obs user mention DJs only run when mentions may have changed
- user site updating doesn't re-index all obs, just update site_id with update_by_query
- added test_after_commit gem to not have to manage spec transaction logic (slows down tests, but less spec configuration needed)
- a couple delayed jobs run less frequently


spec/controllers/identifications_controller_api_spec.rb
```diff
@@ -196,15 +196,6 @@
...
   describe "destroy" do
-    before(:all) do
-      # some identification deletion callbacks need to happen after the transaction is complete
-      DatabaseCleaner.strategy = :truncation
-    end
-
-    after(:all) do
-      DatabaseCleaner.strategy = :transaction
-    end
-
```
kueda, line 207: So does `test_after_commit` run everything in a complete transaction instead of using `ROLLBACK`? Is that going to slow down the test suite?

kueda (approved):
Not sure if you were looking for feedback or not. Looks good to me, just a few questions.

app/models/user.rb
```diff
@@ -593,8 +593,31 @@ def update_observation_sites_later
   end
 
   def update_observation_sites
-    observations.update_all( site_id: site_id, updated_at: Time.now )
-    index_observations
+    observations.update_all(site_id: site_id)
```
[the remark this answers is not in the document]
pleary: oops - not sure why that line got changed. I'll restore the change to updated_at

spec/controllers/identifications_controller_api_spec.rb
pleary: I don't think that using the gem is equivalent to running with the `truncation` strategy all the time, but I'm not also completely sure how the gem works. Specs are quite a bit slower with the gem. I was considering doing something different and using the `truncation` strategy any time has_subscribers was enabled, but that seemed like adding yet more special cases for spec. It sounds like in Rails 5 and above they do something internally to make sue after_commits run in specs, like the gem accomplishes. I also like that when I added the gem I found we had one spec that was passing but should be failing.

If we don't mind the gem, one thing we could do to offset the increase testing time is split up specs into a couple more test groups

kueda: Simpler tests + exposing bugs - slower tests > 0, I guess.

pleary:
Thanks for reviewing. I'll merge this in so we can try it out

PR #3042, opened 2021-05-10
Reindex obs after establishment change

kueda:
Previously it didn't look like we were actually re-indexing observations after establishment means changed, which meant if you changed something to "introduced" observations of that taxon in that place would not have the "introduced" marker on the obs detail pages.

This will definitely introduce some index churn, so if there's a better way to do this, please let me know.


app/models/listed_taxon.rb
```diff
@@ -964,4 +965,18 @@ def establishment_means_description
     I18n.t( "establishment_means_descriptions.#{ key }", default: default )
   end
 
+  def reindex_observations
+    return true if taxon_id.blank? || place_id.blank?
+    return true unless previous_changes[:establishment_means]
+    Observation.
```
pleary, line 971: I think we'll want to tweak this a little. With the `.delay().elastic_index!` setup, this is bypassing some of the logic in `.elastic_index!` for efficient delayed indexing. As written, the scope query will be run and all observations will be fetched and all of them will be serialized in the delayed job. We could improve on that with `Observation.elastic_index!( scope: ..., delay: true )` which will first fetch just the IDs of the obs and create jobs for them and only put the IDs in the delayed job handler - essentially don't call `.delay()` explicitly.

Another potential issue is that this will be using DB querying for obs in places, which can be very slow even for relatively small places, and prohibitively slow for large places. So I think we need to use Elasticsearch for determining which obs are in this taxon and place. I also think we should leave that querying to the delayed job so the user doesn't need to wait for us to figure out which obs to index later. So maybe there's a method like this called `reindex_observations` and another called `reindex_obsdervations_later`. The latter can be the rails callback and it just queues up the other one that'll query ES and kick off the indexing job

kueda: K, I think I've done what you described if you want to take another look.

app/models/listed_taxon.rb
```diff
@@ -964,4 +965,35 @@ def establishment_means_description
...
+    page = 1
+    per_page = 1000
+    while true
+      res = Observation.elastic_search(
+        source: ["id"],
+        filters: [
+          { term: { private_place_ids: place_id } },
+          { term: { "taxon.ancestor_ids": taxon_id } }
+        ]
+      ).page( page ).per_page( per_page )
```
pleary, line 991: Yep, this is the approach I was describing. Overall I think this will work much better. I did hit a problem here though with pagination of ES requests - if there are more than 10k taxa we'll hit the `index.max_result_window` error. We should use the `id_above` approach and filter by obs whose IDs are higher than the last ID of the previous batch.

I also noticed an existing small issue that we can address here or separately in the main brach. When given the delayed param, elastic_index! will still sleep in between queuing each batch to be indexed later when it doesn't need to - it just slows down creating all the sub-jobs. I think we should do something like `if batch_sleep && !options[:delay]` in acts_as_elastic_model.rb

pleary:
Latest changes look good to me. This is ready to merge/release from what I can see

PR #3049, opened 2021-05-14
DwC-A annotations

kueda:
Include annotation-based term values in the DarwinCore Archives.


kueda:
So far there's only support for http://rs.tdwg.org/dwc/terms/sex, but I'd like to do http://rs.tdwg.org/dwc/terms/lifeStage and http://rs.tdwg.org/dwc/terms/reproductiveCondition, hence the draft.

One thing worth discussing: do we want to continue adding database queries like this, or is that not sustainable? Annotation data is already in the ES index, which we're using to query observations for the DwC-A, but which we then kind of ignore when populating the values. If we did something like cache the ES response with the `Observation` instance, we could use that to populate these fields instead of executing multiple queries per batch on `annotations`, `controlled_terms`, `controlled_term_labels`, etc., just not sure that complexity is worth it.

pleary:
I agree that it's a bit of a waste to query Elasticsearch for observations then toss out all but the ID in favor of ActiveRecord. I think it would be great if we could move towards using the ES response directly. It would benefit creating these archives, and could also help a lot with observation search exports and various other things. I just don't know the easiest way to get there. There's still a good amount of rails logic that expect AR objects that would need updating.

For now, I don't think adding an additional query for annotations here will change run-time significantly. This all looks good to me. I did notice some extra queries for controlled value labels, which come from existing logic in ControlledTerm at https://github.com/inaturalist/inaturalist/blob/main/app/models/controlled_term.rb#L80 . I think we should change to `all_labels = labels.order(:id)` to `all_labels = labels.sort_by(&:id)`. The former will always run an additional query regardless of the labels have already been loaded, and I think in this method the labels will all need to be loaded anyway. By using sort_by we stand to gain a little if the labels are already preloaded, and we wouldn't be adding overhead if they are not.

Other than that (which is pretty minor) - looks good to merge to me

pleary (approved):
thanks for the change - looks good to me

PR #3051, opened 2021-05-14
Show who made and updates conservation statuses

kueda:
Since recent statuses are getting a bit weird (e.g. IUCN equiv and obscuration not really matching the cited source), I think we should show this info publicly.

Depends on https://github.com/inaturalist/iNaturalistAPI/pull/256


cmcheatle:
Is this retrospective, or only for newly added/edited items?

kueda:
@cmcheatle we've been recording who added a new conservation status since about [8 March 2020](https://github.com/inaturalist/inaturalist/commit/cdebe73f7c1663d1b86291c386dfdb862590cab7), and who updated an existing conservation status since about [16 February 2021](https://github.com/inaturalist/inaturalist/commit/2f73c81c1d1aff4237764ede4c77d893ced1f94e) (I'm saying "about" b/c these are the dates when the changes to the code were committed, not when they were actually deployed to the live website). The change in this pull request will show that information, so it will be retrospective within those limitations.

PR #3086, opened 2021-06-18
Update it.yml

dartar:
Removing redundant indefinite article that produced "una una osservazione" in the rendered HTML when substituting for %{noun}. 

Note that the use of a feminine indefinite article "una" preceding %{x} in the strings starting at line 4522 on creates an issue when the value is masculine (e.g. "una commento") and there's no easy way to fix this given how these messages are structured. It looks like the previous translator attempted to address this with (what I believe should read) "un/una' on line 4524-4525.

### DIFF
--- a/config/locales/it.yml
+++ b/config/locales/it.yml
@@ -4520,7 +4520,7 @@ it:
   user_added_a_flag_html: "%{user} ha aggiunto un contrassegno"
   user_added_a_project_invitation_html: "%{user} ha aggiunto in invito di progetto"
   user_added_a_x: "%{user} ha aggiunto una %{x}"
-  user_added_a_x_to_noun_by_html: "%{user} ha aggiunto una %{x} a una %{noun} di %{by}"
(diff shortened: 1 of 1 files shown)


kueda:
Hey Dario! Changes to the translation files will just get overwritten the next time we sync with Crowdin, so I just made this change there: https://crowdin.com/translate/inaturalistweb/38/en-it#22276. If you want to make any changes to translations in the future, please look on Crowdin first. Ideally source strings originate in the github repo, where Crowdin retrieves them, volunteers translate them on Crowdin, and then they flow back to the repo.

Also, FWIW, this and the other notification-related strings are some of the hardest ones to translate properly. We've had plans to overhaul translations for a long time, which would hopefully involve reducing the number of variables in these strings, but we've been pretty stymied by just keeping up with growth. Maybe we'll get there some day.

dartar:
Totally understand, thanks for the pointer!

PR #3107, opened 2021-06-30
Add shift/cancel buttons to time shifter

albullington:


app/assets/stylesheets/observations/uploader.scss
```diff
@@ -1119,3 +1119,11 @@ $tickmark: 6px;
...
   padding-left: 5px;
 }
+
+.time-shifter-buttons {
+  margin-top: 15px;
+
+  button {
+    margin-right: 15px
+  }
+}
```
kueda, line 1129: Instead of this stuff, let's add the [`btn-toolbar`](https://getbootstrap.com/docs/3.4/components/#btn-groups-toolbar) class to the `time-shifter-buttons` div to get conventional bootstrap spacing.

app/webpack/observations/uploader/components/time_shifter.jsx
```diff
@@ -141,6 +129,14 @@ class TimeShifter extends SelectionBasedComponent {
...
           {[-24, -12, 0, 12, 24].map( tick => <span className="tick" key={tick.toString( )}>{tick}</span> )}
         </div>
+        <div className="time-shifter-buttons">
+          <button className="btn btn-sm btn-primary" type="button" onClick={this.handleValueChange}>
+            {I18n.t( "shift" )}
+          </button>
+          <button className="btn btn-sm btn-default" type="button" onClick={this.resetShifter}>
+            {I18n.t( "cancel" )}
+          </button>
+        </div>
```
kueda, line 139: Can we hide these when the offset is zero? Right now if it's zero, these buttons do nothing.

config/locales/en.yml
```diff
@@ -4165,6 +4165,7 @@ en:
   settings: Settings
   share: Share
   share_colon: "Share:"
+  shift: Shift
```
kueda, line 4168: Let's change this to

```
# Button label to offset the date/time of observations by a set number of hours
offset_time_verb: Offset
```

That way we're using the same "Offset" verb we're using in the accordion label, translators know it's a verb and not a noun, and they have some understanding of what's being offset.

kueda (changes requested):

albullington:
Made these three changes.

kueda (approved):

Issue #3110, opened 2021-07-02
Dynamic Life List: Sorting obs by "Date Added" actually sorts of "Date Observed"


tiwane:
Go to https://www.inaturalist.org/lifelists/tiwane?view=tree&details_view=observations&tree_mode=full_taxonomy&taxon_id=1 which should show all my animal observations sorted by "Date Added", oldest to newest. However, the first observation is [this one](https://www.inaturalist.org/observations/58517), which was added in 2012. The first animal observation I *added* is [this one](https://www.inaturalist.org/observations/667), which appears third. 

We should either change the label to "Date observed" or actually sort by "Date added" - I'd prefer the latter since the default is to sort by "Date added", newest to oldest.

fredunit:
Can I claim this problem? and do you mean change the capitalized Added to added?

kueda:
@fredunit, go for it. I'd just use whatever version of "date added" we already have translated. Holler if you have trouble getting a working dev env.

fredunit:
So, this mean the function is not correct, and I need to review the logic behind and change it, am I right?

kueda:
The changes in https://github.com/inaturalist/inaturalist/pull/3111/files don't really change the "logic" of anything, they just change the capitalization of three strings. The problem described in this issue is that the sort order on Tony's life list is not bey "date added" as the interface claims, so the cause of the incorrect sorting needs to be investigated. Maybe that means the API request needs a different parameter, maybe that means the API is broken somehow.

fredunit:
So, which @tiwane is the image first submitted should be showed at first one, because the oldest to newest should be base on the date submitted, am I correct_
![problem](https://user-images.githubusercontent.com/52747821/125159439-ea350d00-e1a9-11eb-8d12-a0c3209d7973.jpg)

kueda:
Yes, if the sort order says "Date Added" then the results should be sorted in that order. The important result to achieve is that the UI text should describe what is actually happening.

elysiumplain:
Very old issue - looking at UI and ref https://github.com/adam-rankn/inaturalist/commit/9c6cb743d528783d2793e2448263ba307b8093bb it seems resolved. Want to close, @tiwane ?

PR #3111, opened 2021-07-03
change dynamic life list

fredunit:
Change the label from "Date Added" to "Date added".


kueda:
@fredunit this just changes the capitalization of the text, which isn't what #3110 is about (it's about changing the sort, or changing the label so it actually describes what sort is being applied). Was that your intent here or do you have more commits to add to this PR?

Issue #2547, opened 2020-01-03
Require place names to start with non-numeric characters


carrieseltzer:
When places are created, they should be required to have a name that begins with a non-numeric character. 

If places are named as numbers, then things can get wacky if there is an iNat place with the same number as its `place_id`. This came up in the [Forum](https://forum.inaturalist.org/t/places-and-constituencies/9103) with a place named only "[805](https://www.inaturalist.org/places/unnamed-gadm-805)" (apparently an error from [GADM](https://gadm.org/maps/KEN/baringo_2.html)), which interfered with the `place_id = 805 ` which is [Comal county, Texas](https://www.inaturalist.org/observations?place_id=805).

todtb:
Looks like `Place` already validates on this (`app/models/place.rb:55`). Is it possible this record was created prior to that implementation? Or perhaps there is some external import into the DB that bypasses AR?

kueda:
```sql
# select count(*) from places where name ~ '^[0-9]';
 count 
-------
    31
(1 row)
```

```sql
inaturalist_production=# select id, name, created_at from places where name ~ '^[0-9]';
   id   |                           name                            |         created_at         
--------+-----------------------------------------------------------+----------------------------
 151754 | 5MR Pat Deacon                                            | 2020-05-02 03:17:07.838341
 151502 | 63rd St. Beach Dunes                                      | 2020-04-29 15:28:01.184487
 150370 | 5MR - Jen Grant - Saanich BC                              | 2020-04-06 22:29:59.260996
 152259 | 222 Hamilton Avenue                                       | 2020-05-15 01:08:27.015976
 163518 | 2584, Vidyasagar Sarani                                   | 2020-11-22 20:00:24.803725
 164514 | 50MR - Presqu'ile Point                                   | 2021-01-11 06:08:36.905968
 164515 | 50MR - Presqu'ile Point                                   | 2021-01-11 06:08:36.917145
 162413 | 02_MARIAJOSÉ_BANANOLY2                                    | 2020-10-22 16:57:57.226231
 143639 | 674 Nerissa Rd, Port Shoreham property                    | 2019-10-06 05:17:07.830784
  99127 | 10 Ramadan 1                                              | 2016-03-09 07:50:41.672538
 120469 | 5859 5th Line Property                                    | 2017-05-21 13:39:25.8535
 102211 | 25 de Diciembre                                           | 2016-03-09 12:00:28.61149
[...]

PR #3118, opened 2021-07-07
Selectively validate on place bbox update

todtb:
Was thinking something like this for #2547 @kueda. 

I was assuming the skip validation was primarily around lat/lon, but let me know if you can think of another place where it would be conditional.

I played around with this in console and specs are green, but honestly must admit I don't have a great strategy here for manual testing.

### DIFF
--- a/app/models/place.rb
+++ b/app/models/place.rb
@@ -593,7 +593,7 @@ class Place < ActiveRecord::Base
         h[:swlng] = geom.points.map(&:x).select{|x| x > 0 || x < -180}.min
         h[:nelng] = geom.points.map(&:x).select{|x| x < 0 || x > 180}.max
       else
-        h[:longitude] = geom.centroid.x,
+        h[:longitude] = geom.centroid.x
         h[:swlng] = geom.envelope.lower_corner.x
         h[:nelng] = geom.envelope.upper_corner.x
       end
--- a/app/models/place.rb
+++ b/app/models/place.rb
@@ -41,13 +41,15 @@ class Place < ActiveRecord::Base
 
   validates_presence_of :latitude, :longitude
   validates_numericality_of :latitude,
-    :allow_blank => true, 
-    :less_than_or_equal_to => 90, 
-    :greater_than_or_equal_to => -90
+                            allow_blank: true,
+                            less_than_or_equal_to: 90,
+                            greater_than_or_equal_to: -90,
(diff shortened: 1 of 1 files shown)


app/models/place.rb
```diff
@@ -587,24 +585,19 @@ def append_geom(geom, other_attrs = {})
...
-      self.nelng = geom.envelope.upper_corner.x
+
+  def points_from_geom(geom)
+    { latitude: geom.centroid.y, swlat: geom.envelope.lower_corner.y, nelat: geom.envelope.upper_corner.y }.tap do |h|
+      if geom.spans_dateline?
+        h[:longitude] = geom.envelope.centroid.x + 180*(geom.envelope.centroid.x > 0 ? -1 : 1)
+        h[:swlng] = geom.points.map(&:x).select{|x| x > 0 || x < -180}.min
+        h[:nelng] = geom.points.map(&:x).select{|x| x < 0 || x > 180}.max
+      else
+        h[:longitude] = geom.centroid.x,
```
kueda, line 596: Trailing comma is causing `h[:longitude]` to be an array, which I think is the cause of all the spec failures. Behold the marvels of the Ruby interpreter: it has wonderful little gems like `tap` and truly horrific liabilities like this.

kueda (changes requested):
I suspect this will be fine without that one typo. And yeah, this one's hard to click test, so I think we just need to rely on the specs. In a way, it's encouraging that so many failed with that trailing comma.

app/models/place.rb
```diff
@@ -587,24 +585,19 @@ def append_geom(geom, other_attrs = {})
...
+        h[:longitude] = geom.envelope.centroid.x + 180*(geom.envelope.centroid.x > 0 ? -1 : 1)
+        h[:swlng] = geom.points.map(&:x).select{|x| x > 0 || x < -180}.min
+        h[:nelng] = geom.points.map(&:x).select{|x| x < 0 || x > 180}.max
+      else
+        h[:longitude] = geom.centroid.x
+        h[:swlng] = geom.envelope.lower_corner.x
+        h[:nelng] = geom.envelope.upper_corner.x
+      end
     end
-    calculate_bbox_area
```
todtb, line 606: This is handled by a before_save callback now that we aren't skipping.

kueda (approved):

Issue #2867, opened 2020-11-17
Suggestions filters should persist when changing observation modal tabs


kueda:
The observation modal (detail view on Identify, Compare tool on obs detail) has a "Suggestions" tab that shows suggested taxa for identifications, and allows you to filter those suggestions in a few ways. However, if you view the suggestions, change a filter, navigate to another tab like Annotations, then navigate back to Suggestions, your filters will be reset to the defaults. The people demand that they not do that. The filters *should* still reset when navigating between observations, and I don't think they need to persist when you close the modal.

todtb:
Looks like the "Sort By" and "Source" filters are working as expected, but not the "Taxon" and "Place" filters. I can start looking into why/how to fix. I would think it might also make sense not to fire the query each time the tab is changed, but rather when the tab is first selected and subsequently when changing filters.

kueda:
> I would think it might also make sense not to fire the query each time the tab is changed, but rather when the tab is first selected and subsequently when changing filters.

Sounds good. Also need to fire it after changing observation, of course.

PR #3123, opened 2021-07-10
Persist obs identification modal suggestion filters

todtb:
#2867 

Many of these Redux idioms are still fairly foreign to me, but this tested well in an admittedly small dataset. Subsequently this just covers the filter persistence, and not the query issue as that looks to be a bit more work/grokking of the patterns. Would love to hear any feedback.

Also, there is an existing edge case/bug where unknowns show the last observation's taxa ancestry. 
To reproduce select an obs adjacent to an "Unknown". When filtering on ranked taxon level for the observation (eg Vascular Plants for a observation of a sagebrush), clicking on prev/next from the modal to the adjacent "Unknown" observation will not be able to populate a recommended taxon level for this (the only reasonable one here is State of matter life), but rather defaults back to the last (Vascular Plants). Might be possible to build out a state of matter life object in `app/webpack/observations/identify/components/taxon_chooser_popover.jsx:35`

### DIFF
--- a/app/webpack/observations/identify/ducks/suggestions.js
+++ b/app/webpack/observations/identify/ducks/suggestions.js
@@ -78,44 +78,59 @@ export default function reducer(
       }
       break;
     case UPDATE_WITH_OBSERVATION: {
+      const { observation } = action;
+      const isFeaturedObs = observation.uuid === state.query.featured_observation_uuid
+        || observation.id === state.query.featured_observation_id;
       newState.query = {
         source: state.query.source,
         order_by: state.query.order_by
       };
-      const { observation } = action;
-      if ( observation.taxon ) {
-        let indexOfTaxonInAncestors = observation.taxon.ancestor_ids
-          .indexOf( observation.taxon.id );
-        if ( indexOfTaxonInAncestors < 0 ) {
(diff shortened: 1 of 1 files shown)


kueda:
This is certainly solving the problem #2867 describes. Did you look into preventing the `UPDATE_WITH_OBSERVATION` action from getting fired at all when switching between tabs? Theoretically, there is no reason to update the suggestions at all when switching between tabs, only when moving between observations. Maybe the `currentObservationReducer` needs to remember what the previous observation was and `fetchDataForTab` should only fetch data for a tab once per obs, or maybe `UPDATE_WITH_OBSERVATION` in the suggestions reducer should just return a copy of the old state if `featured_observation_id` didn't change.

Just thinking out loud here. What you've got here works, so I can merge it if you think it's the best solution.

todtb:
I looked into it briefly, but didn't want to hold up getting the fix out while trying to optimize/reduce the number of queries sent for the observations. 

While I'm certainly not a redux expert, I can look into it a little more. I make no claims that it's the best solution 😄  However, this PR here should at least prevent the filters from resetting and satiate the peoples' "demand that they not do that".

Issue #3120, opened 2021-07-07
Adding a new conservation status incorrectly marks existing statuses as updated


tiwane:
From https://forum.inaturalist.org/t/adding-a-new-conservation-status-incorrectly-marks-existing-statuses-as-updated/23465 I can replicate.

1. Find a taxon with an existing conservation status.
2. Add another conservation status to it and save.
3. On the Status screen of the taxon, all conservations statuses show and "updated by" date and attribution to you and to the current time and date, even if existing statuses were not touched.

todtb:
I can start looking into this

PR #3132, opened 2021-07-16
Filter blank values from conservation status params

todtb:
https://github.com/inaturalist/inaturalist/issues/3120

Ultimately this comes from a combination of normalizing the geoprivacy field and the params passing blank strings. Since `before_save` still lets the timestamps get touched and the record is "dirty" it shows as updated. Place_id could do something similar after it gets cast from `""` to `nil` on the db end. Nullifying the two offending params is the quickest fix here I think.

### DIFF
--- a/spec/controllers/taxa_controller_spec.rb
+++ b/spec/controllers/taxa_controller_spec.rb
@@ -314,7 +314,8 @@ describe TaxaController do
             "0" => {
               "id" => cs.id,
               "status" => cs.status,
-              "authority" => cs.authority
+              "authority" => cs.authority,
+              "geoprivacy" => cs.geoprivacy
             }
           }
         }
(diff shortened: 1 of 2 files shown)


spec/controllers/taxa_controller_spec.rb
```diff
@@ -314,7 +314,8 @@
             "0" => {
               "id" => cs.id,
               "status" => cs.status,
-              "authority" => cs.authority
```
todtb, line 317: Without geoprivacy it was changing from default (obscured), to ""/nil (open). Might be missing something with the way the params are parsed, but I think this makes sense?

kueda:
Bonus: I learned about `%i[foo bar]`. Thanks!

Issue #3127, opened 2021-07-14
Show site ids on /announcements


carrieseltzer:
With more network members using announcements, it would be useful for staff to be able to see at a glance which announcements are on which sites (since we see them all). Can we add a column after `Placement` that displays the site IDs associated with each announcement on /announcements? 

This will also help make it more visible to network members when they're looking at (and therefore shouldn't edit) an announcement displayed on multiple sites.

<img width="1012" alt="Screen Shot 2021-07-14 at 9 43 28 AM" src="https://user-images.githubusercontent.com/7501183/125632868-63b26323-762e-4e66-a165-3d7232bc36c3.png">

todtb:
Coming right up

todtb:
Hey @carrieseltzer, got a PR (https://github.com/inaturalist/inaturalist/pull/3134) but wasn't sure if you wanted this column visible for ALL admins (including site specific), or just top level admins. For now it's the latter, but a very simple fix if you were thinking of making it visible for all.

carrieseltzer:
Thanks @todtb! It should be visible for all site admins, not only iNat staff.

kueda:
Only staff (`is_admin?`) and site admins (`is_site_admin_of?( @site )`) can even see this page, so I think it's ok to show the "Sites" column to everyone who can see the page, even if the site admins just see the same thing for all their announcements (if they see something different then that would be a good indication that something's wrong, but it's also theoretically possible for one user to be a site admin for several sites).

carrieseltzer:
Sometimes staff (mostly me I guess) make announcements visible on multiple sites, so in those cases it would help the site admins see that an announcement isn't only on their site, so any edits they make would impact all selected sites.

PR #3134, opened 2021-07-16
Add site ids to announcement index

todtb:
#3127

### DIFF
--- a/app/views/announcements/index.html.haml
+++ b/app/views/announcements/index.html.haml
@@ -21,6 +21,7 @@
         %thead
           %tr
             %th= t :placement
+            %th= t :sites, default: "Sites"
             %th= t :start
             %th= t :end
             %th= t :body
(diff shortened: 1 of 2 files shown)


app/views/announcements/index.html.haml
```diff
@@ -21,13 +21,17 @@
...
+              %th= t :sites, default: "Sites"
             %th= t :start
             %th= t :end
             %th= t :body
         %tbody
           - @announcements.each do |announcement|
             %tr
               %td= h announcement.placement
+              - if current_user_site_ids.blank?
+                %td= announcement.site_ids.empty? ? t(:all) : announcement.site_ids.join(", ")
```
kueda, line 34: I think we should show the site names instead of the site IDs here.

kueda (changes requested):
I think it would suffice to just show the "Sites" column to anyone who can view the page, including staff and site admins. If site admins generally just see the same site over and over again, that's ok. Theoretically, some users might be site admins for numerous sites, too (though I don't think that's happened in reality yet).

kueda (approved):

Issue #3147, opened 2021-07-26
On a taxon's flags page, change format to look more like /flags


tiwane:
From https://forum.inaturalist.org/t/remove-links-on-taxon-flags-page/3216

The links on pages like this (eg https://www.inaturalist.org/taxa/890330/flags) are redundant. One proposed solution would be to make it look more like the layout at /flags

<img width="1144" alt="image" src="https://user-images.githubusercontent.com/15896181/127063236-b8333360-f5d8-4cae-bb10-883398be11d7.png">

Pros: easier to read, consistent with /flags, fewer redundant and confusing links (eg resolve/unresolve doesn't do either of those things).

Cons: people may not know they can comment or resolve/unresolve the flag by clicking on Details.

todtb:
I'll take a look

todtb:
Hey @tiwane, fitting this in stylistically is kind of an open question since the previous page included actions (hide, unhide) that do not fit the structure of the table.

Here is one proposed compromise below.

I think it's fairly clear to be an action and not a line item. It's also possible to break out the actions into a separate table, but I feel that loses some of the "timeline" feel we have now.

The other matter of redundant links should be more straight forward, just let me know which links up top need to go.

Let me know your thoughts!

![image](https://user-images.githubusercontent.com/19367605/127404252-360bfb85-4251-4d38-a848-803247a6015a.png)

tiwane:
I think this looks good. "Actions" seems like a weird column heading but we've been using at /flags for years so I guess it's working fine. 

>The other matter of redundant links should be more straight forward, just let me know which links up top need to go.

That was referring to these links, and I think your solution takes care of them. 

<img width="344" alt="image" src="https://user-images.githubusercontent.com/15896181/128240960-fb754b1d-9f80-4efc-8525-145f85979735.png">

tiwane:
@todtb I think it looks great but as I was testing on staging I think it would be good if we grayed out resolved flags to make them immediately distinguishable from unresolved flags.

todtb:
Hey @tiwane, that shouldn't be too difficult, but do you have an example page that you were thinking of stylistically?

PR #3151, opened 2021-07-28
Update resource flag index page

todtb:
Further discussion in #3147


pleary:
@todtb there's one very small thing I noticed that might be nice to fix before releasing. I noticed that with the new flag view helper, there is no longer a space in between icons and text - for example between the "Content" name and the pencil icon, or between the (i) icon and "Details". Otherwise this looks great

todtb:
> I noticed that with the new flag view helper, there is no longer a space in between icons and text - for example between the "Content" name and the pencil icon, or between the (i) icon and "Details".

Eagle eyes, nice catch!

Issue #3150, opened 2021-07-28
Can't create complex if complex's name = name of one of its parent's children


tiwane:
From: https://forum.inaturalist.org/t/species-with-same-name-as-complex-prevents-creation-of-complex-at-the-correct-location/23564

1. Create a new species complex taxon with the name `Amaranthus deflexus` and parent `Amaranthus`.
2. See this warning:

<img width="560" alt="image" src="https://user-images.githubusercontent.com/15896181/127400117-61e44087-2e1f-48e5-886d-654f41d7e2b1.png">

Most (all?) complexes will have the same name as one of its parent's children, so this shouldn't be a blocking error when creating a new complex.

todtb:
I can take a look at this @tiwane. I'm also assuming that we should only be able to create a single sibling complex with that duplicated name. For example, once the 'Amaranthus deflexus' complex is created (alongside 'Amaranthus deflexus' species), no other complexes named 'Amaranthus deflexus' should be allowed as a child of 'Amaranthus'. Sound right?

PR #3152, opened 2021-08-01
Fix complex taxon unique validation

todtb:
#3150 

Allow a complex to share name with sibling species, but do not allow multiple sibling complexes with the same name.


pleary:
Thanks @todtd - this is looking good to me. Are you set up to run the rspec tests for this repo? I think it would be good to have a couple tests on the taxon model for these new validations before we merge the code it. Is that something you feel comfortable working on?

pleary:
Thanks for the specs! This is all looking good so I've merged it into main.

Issue #2472, opened 2019-11-08
Allow users to flag places for curation


tiwane:
The forum (and sometimes help@inat) are currently the places where people ask for places to be edited and it would best to keep this onsite.

PR #3153, opened 2021-08-02
2472 Add flags to places

todtb:
#2472 

Kind of from a while back, but if this is still on the table @tiwane, @kueda, this adds in the ability to track flagged places from within iNat proper

Screen Shot 2021-08-02 at 5 37 40 PM

image
(subject to change in https://github.com/inaturalist/inaturalist/pull/3151)

image


pleary:
This is looking good to me. @tiwane assuming this is something you'd still like to have on the site, does this seem releasable to you, or is there anything you'd want to have changed?

PR #3158, opened 2021-08-11
Minor fix to bin/setup for Ubuntu

seanclifford:
On Ubuntu, I wasn't able to get past the createdb check when setting up a development environment, even with the createdb command available in the shell.

I found when run in `irb`:
`system "command -v createdb"`
returned `nil`.

Changing this to:
`system "command -v createdb > /dev/null"`
returned `true`.

Testing:
`system "command -v invalidcommand > /dev/null"`
returned `false`.

NOTE: I haven't tested on macOS, as I don't have access to a mac.

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -27,7 +27,7 @@ Dir.chdir APP_ROOT do
   end
 
   puts "\n== Creating Template Database =="
-  if system "command -v createdb"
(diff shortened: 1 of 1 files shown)


kueda:
I checked on a Mac and it works just fine. Thanks!

Issue #3131, opened 2021-07-15
Date picker in the Uploader is not localized


kueda:
**Describe the bug**
When you edit a date in the Uploader, either in the side menu or an obs card, the month and day names are not localized to the current locale.

**To Reproduce**
1. Set your language to French
2. Go to https://www.inaturalist.org/observations/upload and make a new observation
3. Click the date field in the side bar or in an obs card. The month and day names are not localized (e.g. the first weekday is labelled "Mo" instead of "Lu")

**Expected behavior**
Month and day names should be in the current locale.

**Screenshots**
![Screen Shot 2021-07-15 at 1 00 12 PM](https://user-images.githubusercontent.com/22691/125850196-04180a41-e443-48d1-9b4d-374b87f72814.png)

**Context**
 - OS: Mac OS
 - Browser Brave

**Additional context**
This might just require setting the moment locale correctly... or we might need to update to the more current https://github.com/arqex/react-datetime

kueda:
Closed by https://github.com/inaturalist/inaturalist/pull/3161

PR #3161, opened 2021-08-11
Show localized days of week and month in calendar (#3131)

albullington:
I tested this locally in French & Indonesian on the uploader, project creation form, and identify filter.

Moment.locale sets the correct locale for calendar month. The related PR in `react-bootstrap-datetimepicker` is needed to update the locale for days of the week.


kueda (approved):
Looks good to me. Github Actions were chocking on the v2 lock file you added so I replaced it with an older one and the specs seem to be passing.

PR #3155, opened 2021-08-04
Add shoulda matchers (a-f)

todtb:
Hey @pleary, was writing some of those specs [over here](https://github.com/inaturalist/inaturalist/pull/3152) and was wondering if you've ever considered 'shoulda-matchers'.

Here's a few samples of the built in matchers, but the full list is fairly extensive: https://github.com/thoughtbot/shoulda-matchers#matchers

If you like the idea, I can add more spec cases.

PR #3162, opened 2021-08-12
Add shoulda matchers "g-m"

todtb:
Related: https://github.com/inaturalist/inaturalist/pull/3155


kueda (changes requested):
This looks good to me, but could you remove the tests on all the Goal* models? Those are vestigial and should be removed, so testing them is just going to increase the time it takes to run the test suite without any benefit.

PR #3163, opened 2021-08-12
Remove deleted user profile pics from s3 within a month

kueda:
Prior to this we were only removing the files when a user asked to be forgotten. Patrick, I think this should work, but it's a bit hard to test so I'd appreciate another pair of eyes.


pleary:
Looks good to me. I tested locally with my own S3 account and icons were indeed deleted

PR #3166, opened 2021-08-12
Add shoulda matchers o p

todtb:
Related #3155

### DIFF
--- a/spec/models/observation_field_spec.rb
+++ b/spec/models/observation_field_spec.rb
@@ -1,17 +1,30 @@
 require File.dirname(__FILE__) + '/../spec_helper.rb'
 
 describe ObservationField do
+  it { is_expected.to belong_to :user }
+  it { is_expected.to have_many(:observation_field_values).dependent :destroy }
+  it { is_expected.to have_many(:observations).through :observation_field_values }
+  it { is_expected.to have_many(:project_observation_fields).dependent :destroy }
+  it { is_expected.to have_many(:projects).through :project_observation_fields }
+  it { is_expected.to have_many(:comments).dependent :destroy }
+
+  it { is_expected.to validate_uniqueness_of(:name).case_insensitive }
+  # TODO: `validate_presence_of :name` contradicts `allow_blank` of length validation
+  xit { is_expected.to validate_presence_of :name }
+  xit { is_expected.to validate_length_of(:name).is_at_most(255).allow_blank }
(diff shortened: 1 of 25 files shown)


kueda:
Argh, sorry @todtb, I "shoulda" merged this before the ProjectInvitiation PR. Could you resolve the conflict here?

spec/models/observation_field_spec.rb
```diff
@@ -1,17 +1,30 @@
...
 describe ObservationField do
+  it { is_expected.to belong_to :user }
+  it { is_expected.to have_many(:observation_field_values).dependent :destroy }
+  it { is_expected.to have_many(:observations).through :observation_field_values }
+  it { is_expected.to have_many(:project_observation_fields).dependent :destroy }
+  it { is_expected.to have_many(:projects).through :project_observation_fields }
+  it { is_expected.to have_many(:comments).dependent :destroy }
+
+  it { is_expected.to validate_uniqueness_of(:name).case_insensitive }
+  # TODO: `validate_presence_of :name` contradicts `allow_blank` of length validation
```
todtb, line 12: I noticed in `observation_field.rb` the seeming contradiction:
```
validates_presence_of :name
validates_length_of :name, :maximum => 255, :allow_blank => true
```
As a user, my instinct is that `allow_blank` here is no longer applicable?

todtb:
> Argh, sorry @todtb, I "shoulda" merged this before the ProjectInvitiation PR. Could you resolve the conflict here?

All set! I left a comment that you might have a real quick answer to, but not a blocker since the specs are marked pending.

spec/models/observation_field_spec.rb
kueda: Yeah, an `ObservationField` needs a name, so I suspect we can get rid of `allow_blank`

spec/models/project_spec.rb
```diff
@@ -1,6 +1,41 @@
 require File.dirname(__FILE__) + '/../spec_helper.rb'
 
 describe Project do
+  xit { is_expected.to belong_to :user }
+  xit { is_expected.to belong_to(:place).inverse_of :projects }
```
todtb, line 5: With the 5.x switch to requiring `belongs_to` by default and then globally setting it to false, I've marked a few of these pending. I think the best solution would be to rewrite the association:

```ruby
belongs_to :user, optional: false, inverse_of: :projects
```
and remove the separate presence validation, but I'll leave those changes to another PR (hopefully after factories merges).

PR #3169, opened 2021-08-13
Add jwt secret entries to config.yml.example

seanclifford:
To make setting up a new development environment smoother, as the jwt secret fallbacks in [lib/json_web_token.rb](/inaturalist/inaturalist/blob/main/lib/json_web_token.rb) don't match the jwt secret values in [iNaturalistAPI/config_example.js](https://github.com/inaturalist/iNaturalistAPI/blob/main/config_example.js).
Without this change, authenticated API requests fail out of the box on pages such as the Account Settings page.

These jwt secret values I've added here match the jwt secret values already in `iNaturalistAPI/config_example.js`

Instead of this change, another option could be to strip the jwt secret values from `iNaturalistAPI/config_example.js`, and rely on the fallbacks matching in both codebases for development environments.

### DIFF
--- a/config/config.yml.example
+++ b/config/config.yml.example
@@ -126,6 +126,11 @@ base: &base
     # same code running at CONFIG.tile_servers.elasticsearch
     node_api_url: http://localhost:4000/v1
 
+    # The secrets used to sign the JWT for authenticated requests to the
+    # iNaturalistAPI. These need to be same as in the API's config.js.
+    jwt_secret: jwtSecret
+    jwt_application_secret: jwtApplicationSecret
+
     # # Path to the CA .crt file
     # ca_file: "/path/to/certs/ca-bundle.crt"
     # # Path to the certificates directory


seanclifford:
@kueda Here's another dev setup bump in the road I found when setting up my dev environment.
Is just opening these PRs as I have been an acceptable way to contribute these small fixes?

todtb (approved):

todtb:
@seanclifford I see you're using the containerized workflow, how is that working out for you with a new setup?

config/config.yml.example
```diff
@@ -126,6 +126,11 @@ base: &base
     # same code running at CONFIG.tile_servers.elasticsearch
     node_api_url: http://localhost:4000/v1
 
+    # The secrets used to sign the JWT for authenticated requests to the
+    # iNaturalistAPI. These need to be same as in the API's config.js.
+    jwt_secret: jwtSecret
+    jwt_application_secret: jwtApplicationSecret
```
todtb, line 132: I noticed a few places in the config where a fallback is configured:

```ruby
CONFIG.jwt_secret || "secret"
...
CONFIG.jwt_application_secret || "application_secret"
```
But the fallback in js is slightly different.

Maybe the example configs and the runtime fallbacks could be made more consistent overall, but I like where it's headed.

kueda:
Many thanks, @seanclifford, I was just actually setting up a different machine with @todtb's Docker instructions and hit this very problem yesterday. Brian, setting up the containers was pretty smooth and a lot easier than doing things manually. IMO, I find `make services-api` a bit weird b/c you need to do some non-Docker setup to get it to work (check out the API repo, set up the config files). Unless the API can be fully containerized for developers who exclusively want to work on the Rails app, I sort of feel like it should be treated as a separate, non-containerized application that you need to run in parallel, but that depends on the same services (Postgres, elasticsearch, etc.) that the Rails app does. But then again, I know practically zilch about Docker, so maybe that's standard practice.

kueda:
Ah, and yes, PR's are a good way of making changes like this. Usually we'd like to see an issue first so we can consider whether we actually want the work done before you start putting time into it, but for small changes like this that don't cost you much time, I think a PR is fine.

seanclifford:
@todtb I found it very easy to setup. Thank you for putting it together. Just a [couple of tweaks to config files](https://github.com/seanclifford/inat-dev-setup/blob/master/setup_inat.sh#L5) that I think would have been required in some form if I set all the services up manually.
The only thing that I found that might be an issue in future is the tasks in [es.rake](https://github.com/inaturalist/inaturalist/blob/main/lib/tasks/es.rake) that check for elasticsearch binaries before running. Luckily, the one I needed to create all the indexes when I was setting up  - `es:rebuild` - didn't run this check.

On developing the API - I'm also not super familiar with docker, having only used it to host service dependencies before, but I imagine it would just require a `make services-api` after making changes to the API code?

@kueda Thanks for the merge and the feedback on contributing. I'll make use of issues and potentially the forum before making any non trivial changes.

config/config.yml.example
seanclifford: Ah yes I hadn't noticed the fallbacks were different on each side for the application secret at least. 
It would also be nice to make the example configs match the fallbacks too.
Would this change be worthy of a new issue?

kueda: IMO, this is equally trivial, so if you want to put it in a PR we can just review it there.

todtb:
> I find `make services-api` a bit weird b/c you need to do some non-Docker setup to get it to work (check out the API repo, set up the config files). 

I think that could be a nice next step @kueda, and could certainly be done, but with a few pre-reqs. I think it would require hosting the docker image on either a public or iNat hosted image repository. Perhaps there would need to be builds associated with either release tags or branches? I know there's some cool things that can be done with the `build:` tag in `docker-compose`. So there's a little overhead, all doable, but some infrastructure required. The short term benefit of checking out the repo individually would be that the volume maps directly to the local repo, so you can just check out whatever iNatAPI branch you need.

> I sort of feel like it should be treated as a separate, non-containerized application that you need to run in parallel, but that depends on the same services (Postgres, elasticsearch, etc.) that the Rails app does.

`make services` alone should allow that if the ENV variables are set to point to the docker containers. Also if you wanted to manually start just iNat API, running `docker compose up` can be used.

@seanclifford and @kueda, thanks for the feedback. I probably missed a few things in the configs setup since I was migrating to containerization, rather than starting from scratch, so the fresh eyes are super helpful. I can look into some of the tweaks, or if you've got some good first hand experience from the setup I can provide any comment I can on PRs.

PR #3174, opened 2021-08-14
Alter default jwt secrets in config.yml.example

seanclifford:
To match tidy of jwt secret defaults in iNaturalistAPI. It makes the example config values match the fallback values in code.

This PR is the pair of  https://github.com/inaturalist/iNaturalistAPI/pull/268 and a follow up to https://github.com/inaturalist/inaturalist/pull/3169


seanclifford:
Thanks @kueda 
Not sure if you've noticed, but inaturalist/iNaturalistAPI#268 is the other side to this - still to be merged.

PR #3175, opened 2021-08-14
Address slow spec

todtb:
Hey @kueda, @pleary. 

After adding in some of the simple validation specs to hopefully capture intent and prevent regressions, I wanted to look into profiling and speeding up some of the test runs. I think a large portion is DB hits (with other calls such as DelayedJob, ES, etc adding some as well). 

While it would be possible to rework some of the blueprints machinist defines (as many `make!` associations by default), it is no longer maintained, and might be iffy going into Rails 5, so I'm not sure if it's worth the extended effort. The leading contender would be FactoryBot (actively maintained) which has come a long way in terms of avoiding persistence where not necessary.

I've targeted these tests in particular for their long run time.

Lastly, while largely similar to `machinist`, I do believe that the continued community around `FactoryBot` has added some nice features, a few of which I've tried to demo here (particularly in `factories/taxa.rb`). That combined with the push towards Rails 5 and `machinist` no longer being maintained made me think it might make a good target for gradual migration.

Let me know what you think though.

---
#### Output

Build vs create
```
Run options:
  include {:locations=>{"./spec/models/observation_spec.rb"=>[1584]}}
  exclude {:disabled=>true}
.

Finished in 3 minutes 11.8 seconds (files took 34.43 seconds to load)
1 example, 0 failures
```
->
```
Run options:
  include {:locations=>{"./spec/models/observation_spec.rb"=>[1584]}}
  exclude {:disabled=>true}
.

Finished in 11.08 seconds (files took 28.49 seconds to load)
```

GuideTaxon stub setup:
This test was targeted again for run time, though most of the gains here are from the one expectation and stubbing out the expensive setup rather than any particular fixture/factory differences
```
Top 10 slowest examples (244.1 seconds, 14.7% of total time):
	GuideTaxon creation should be impossible to create more than 500 guide taxa per guide 
		114.19 seconds ./spec/models/guide_taxon_spec.rb:36
```
->
```
Top 3 slowest examples (0.40952 seconds, 2.6% of total time):
  GuideTaxon creation #within_count_limit should not be able to create more than 500 guide taxa per guide should not be valid
    0.2256 seconds ./spec/models/guide_taxon_spec.rb:44
  GuideTaxon creation #within_count_limit should be able to create up to 500 guide taxa per guide should be valid
    0.09294 seconds ./spec/models/guide_taxon_spec.rb:56
  GuideTaxon creation #within_count_limit should not be able to create more than 500 guide taxa per guide should have error
    0.09098 seconds ./spec/models/guide_taxon_spec.rb:46
```

For project observation:
A mix of both here again. The test was doing more work than it needed since the limit was changed from 30 to 15. Still, the results are better than the expected roughly 50%.
```
Run options:
  include {:locations=>{"./spec/models/project_observation_spec.rb"=>[189]}}
  exclude {:disabled=>true}
.

Finished in 1 minute 15.1 seconds (files took 21.4 seconds to load)
1 example, 0 failures
```
->
```
Run options:
  include {:locations=>{"./spec/models/project_observation_spec.rb"=>[189]}}
  exclude {:disabled=>true}
.

Finished in 23.6 seconds (files took 21.85 seconds to load)
1 example, 0 failures
```


kueda:
3 mins vs 11 seconds sounds amazing. FWIW, we're about to merge in a Rails 5 upgrade and machinist still works, but certainly not opposed to switching if it means that much of a speed boost. Since this PR is going to change a lot of tests I'm going to hold off on merging it until after we've got Rails 5 working. That will probably mean some adjustments here to accommodate Rails 5.

todtb:
> 3 mins vs 11 seconds sounds amazing. FWIW, we're about to merge in a Rails 5 upgrade and machinist still works, but certainly not opposed to switching if it means that much of a speed boost. Since this PR is going to change a lot of tests I'm going to hold off on merging it until after we've got Rails 5 working. That will probably mean some adjustments here to accommodate Rails 5.

Oh thanks, good to know, exciting! Also I think the fairest comparison would be the last where we would expect 37s (roughly 1:15 / 2) vs 23s. The 3m vs 11s example could theoretically be achieved (and perhaps only marginally slower) after reworking the machinist blueprints and specs. Most of it comes from `make` vs `make!` or factory bot's `build` vs `create`. However the bigger question would be would it be worth the effort of updating the blueprints and going through the specs when we know it is no longer maintained, or gradually moving over to the library that is maintained and targeting the slowest examples with the big gains shown here. I think after the big gains the speed improvement will be a lot less noticeable, but the library is current and has some pretty nice add on features as well.

todtb:
All set @kueda on adding in factories. More updates to come soon on slow specs, but this is ready for review.

kueda:
I think this is looking good and I agree it would be better to shift to a maintained dependency, and FactoryBot looks fine. My only quibble is that the `Color` specs should probably be deleted rather than updates, but that's a problem for another day. @pleary, do you want to take a look at this? Fairly significant change to our test system, but probably for the best.

todtb:
> `Color` specs should probably be deleted rather than updates

I only see the method called one place from `app/controllers/guides_controller.rb:408`, but if that endpoint is no longer used, I can just clean it up in this PR.

kueda:
It's used there, but guides are not something we're actively maintaining any more, so I would just remove the Color model and any functionality that refers to it... but that's even further off topic for this PR, so I wouldn't worry about it for now.

pleary:
Seems like a good change to me. We've been upgrading to not fall too far behind, and this seems like a good change for the same reason. Anecdotally the guide spec for me locally dropped from about 5 minutes to about 1 minute, which is a great improvement. I'll all for merging this in whenever and using FactoryBot going forward

Issue #3129, opened 2021-07-14
Remove ProjectInvitation


kueda:
This was a model that recorded a request that an observation be added to a project. This was before collection projects (which don't record explicit inclusion in a project), and it seems like it was something we abandoned around that time, or maybe when we redid the obs detail page. Either way, there doesn't seem to be a way to actually create these things anymore, so we should just remove the remaining infrastructure:

- [x] Check to see if there are any remaining legitimate uses of this (we're not using it in the mobile apps, are we?!)
- [x] Remove `ProjectInvitationsController` and all associated views and routes
- [x] Remove any links to actions in that controller, including the one on `ProjectsController#by_user`
- [x] Remove related translation keys
- [x] Remove the `ProjectInvitation` model
- [x] Add a migration to remove the `project_invitations` table

todtb:
I think it makes sense to do this in a (at least) 3-step process:
- Remove routes, controllers, views, links, etc.
- Remove model, associations, i18n keys
- Drop table

I'm testing a build of the first step here: #3137

PR #3180, opened 2021-08-19
Drop project invitations table

todtb:
#3129 

In order to preserve the best schema dump match to `structure.sql`, I've omitted that here.

### DIFF
--- a/db/migrate/20210819214533_drop_project_invitations.rb
+++ b/db/migrate/20210819214533_drop_project_invitations.rb
@@ -1,0 +1,16 @@
+class DropProjectInvitations < ActiveRecord::Migration[5.2]
+  def up
+    drop_table :project_invitations if table_exists? :project_invitations
+  end
+
+  def down
+    create_table :project_invitations do |t|
+      t.integer :project_id
+      t.integer :user_id
+      t.integer :observation_id
+
+      t.timestamps
+    end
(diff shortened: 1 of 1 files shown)


kueda:
@todtb the schema issues seem to stem from some additional schemas and extensions added in the Postgis image used in the Docker container, right? Could we run a script to drop the topology and TIGER stuff in the when the database gets setup? Or maybe use a different Postgis image?

todtb:
> the structure.sql issues seem to stem from some additional schemas and extensions added in the Postgis image used in the Docker container, right?

@kueda, almost certainly. 

We could try an approach like is done for `kuromoji` in ES, or just a derivative dockerfile defined in the project. I think when I set up the DB locally (non-containerized), I used `postgres (PostgreSQL) 12.4` which is a bit ahead of the current docker-compose. Does that sound like the right version? Any other extensions that might be missing?

kueda:
Probably best to target what we have [configured for CI](https://github.com/inaturalist/inaturalist/blob/main/.github/workflows/CI.yml), i.e. Postgres 12 and Postgis 2.5. Not sure if the minor version matters too much for Postgres. The only important extension other than postgis is `uuid-ossp` (though we don't seem to specify that in the CI config, so maybe it's not absolutely necessary).

Issue #3145, opened 2021-07-26
ObservationsExportFlowTask should periodically check FlowTask still exists


pleary:
I've occasionally seen ObservationsExportFlowTask DelayedJobs running where the associated FlowTask doesn't exist. This shows up in the DelayedJobs list with arguments `{:query=>nil}`, which makes me think these vestigial jobs are trying to export everything. We should have the query loop part of obs exports periodically check the FlowTask still exists while iterating in case the job is deleted while the export is running, and stop processing if that's the case

PR #3181, opened 2021-08-19
Stop exporting tasks that don't exist

kueda:
Is this what you had in mind for #3145, @pleary?


pleary:
yep looks good to me. I just wanted to make sure these didn't run forever and that deletion check should do that

PR #3184, opened 2021-08-20
More slow specs

todtb:
More factories added (using existing machinist blueprints as... blueprints), more work on the slowest examples:

`spec/models/project_observation_spec.rb`
`9 minutes 28 seconds` -> `4 minutes 38.2`

`spec/models/observation_spec.rb:58`
`4 minutes 43.9 seconds` -> `2 minutes 2.7 seconds`

### DIFF
--- a/spec/factories/observations.rb
+++ b/spec/factories/observations.rb
@@ -6,5 +6,35 @@ FactoryBot.define do
     description { Faker::Lorem.sentence }
     uri { Faker::Internet.url }
     observed_on_string { 'yesterday at noon' }
+    observed_on { Date.yesterday }
+    time_observed_at { Date.yesterday.noon }
+    time_zone { 'UTC' }
+    zic_time_zone { 'UTC' }
+  end
+
+  trait :without_times do
+    observed_on { nil }
+    observed_on_string { nil }
+    time_observed_at { nil }
+    time_zone { nil }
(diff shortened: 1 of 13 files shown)


spec/factories/observations.rb
```diff
@@ -1,7 +1,40 @@
...
+    time_zone { nil }
+    zic_time_zone { nil }
+  end
+
+  trait :research_grade do
+    taxon { build :taxon, :as_species }
+    latitude { 1 }
+    longitude { 1 }
+    quality_grade { Observation::RESEARCH_GRADE }
+    identifications { [association(:identification)] }
```
kueda, line 28: I don't think this will actually make a Research Grade observation, right? Seems like this is just going to make *an* identification associated with the observation, not an identification matching the identification the observer chose. And no media.

I don't think this needs to block the merge, but something to keep an eye on.  We might want to name this something to indicate that it's only making something that's sort of an RG obs, maybe only suitable for unit testing, and `make_research_grade_observation` should be used for more integrative tests.

todtb: Ahh, I see, absolutely something to correct early on! Particularly before it's used too much and giving a false sense of intent. I tried to address here: https://github.com/inaturalist/inaturalist/pull/3189 . Hopefully this can stub out (or create) something that qualifies as research grade.

PR #3205, opened 2021-09-02
Fix paragraphs not being applied in journal posts

### LINKED ISSUE #3204: Server-side markdown parsing removes linebreaks from paragraphs between lists
**Describe the bug**
As [reported in the Forum](https://forum.inaturalist.org/t/journal-posts-have-lost-formatting/25772/35), paragraphs entered between lists lose their linebreaks.

**To Reproduce**
1. Go to to your journal and make a new draft post
1. Enter this the following text:
    ```markdown
    List 1
    1. point 1
    2. point 2

    In between Paragraph 1

    In between Paragraph 2 

    In between Paragraph 3 

    List 2
    1. point 1
    1. point 2

    Extra paragraph 1

    Extra paragraph 2

    Extra paragraph 3
    ```
1. Preview that and you'll see the "In between" paragraphs are all on one line, when they should have linebreaks between them: 
Screen Shot 2021-09-01 at 11 16 40 AM

Note that this does *not* happen in our JS markdown parsing, e.g. for comments and IDs.

**Expected behavior**
It should look like Github markdown parsing:

List 1
1. point 1
2. point 2

In between Paragraph 1

In between Paragraph 2 

In between Paragraph 3 

List 2
1. point 1
1. point 2

Extra paragraph 1

Extra paragraph 2

Extra paragraph 3

**Context**
Since this only seems to be happening in server-side markdown parsing, presumably this affects everyone.

**Additional context**
This is probably a problem with [`ApplicationHelper#formatted_user_text`](https://github.com/inaturalist/inaturalist/blob/main/app/helpers/application_helper.rb#L323)

seanclifford:
Fixes #3204

Paragraphs were not being applied when in between 2 sections of tables, lists or preformatted text.
This was because the regex used was greedy, and would include everything between the first and last of one of those tags, rather than stop matching after the first end tag it finds.

Fix was to apply the non-greedy `?` modifier to the `*`s.

Issue was also visible on messages as well as journal posts.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -361,7 +361,7 @@ module ApplicationHelper
 
   def simple_format_with_structure( text, options )
     new_text = ""
-    chunks = text.split( /(<table.*table>|<ul.*ul>|<ol.*ol>|<pre.*pre>)/m )
+    chunks = text.split( /(<table.*?table>|<ul.*?ul>|<ol.*?ol>|<pre.*?pre>)/m )
     chunks.each do |chunk|
       if chunk =~ /<(table|ul|ol)>/
         html = Nokogiri::HTML::DocumentFragment.parse( chunk )


kueda:
Thanks, @seanclifford!

PR #3207, opened 2021-09-03
Identify: photo brightness shortcuts

albullington:
Adds increase/decrease brightness keyboard shortcuts to the identify modal.


kueda (commented):
This works great and I have no problem merging as is. However, the choice to make a separate reducer for brightnesses kind of bothers me. I don't imagine that part of state will be re-used elsewhere so maybe the observations reducer would be a better place for it? Similarly, the `ObservationModal` component probably doesn't need to know about the brightnesses for *all* relevant observations and photos, just the ones it's showing for the current observation.

What do you think, @albullington?

albullington:
Definitely agree on the separate reducer being an awkward choice. I originally put brightnesses in the current observation reducer, then broke it out when I realized the brightnesses state needed to be maintained when navigating between observations (not just photos), and I never looked for a better place to put that code. I updated the code so brightnesses is now in the observations reducer.

I also agree that the `ObservationModal` component doesn't need to know about all the brightnesses, but I didn't want to make any big changes from how the brightnesses code was working when state was local. We could leave as is, or we could nest all the brightnesses their observation ids and only pass the brightnesses for the current observation id into the `ObservationModal` component.

albullington:
I ended up trying option three here -- continuing to store all brightness keys the way they were before, but filtering the keys that get passed into the observation modal so the modal only knows about the keys relevant to that observation. That seemed a little more straightforward / less fussy than nesting.

PR #3208, opened 2021-09-03
Nested open geoprivacy conservation status

### LINKED ISSUE #3185: Make it so that obscuring-conservation-status on place ancestor doesn't override open-conservation-status on place descendant
This  is our current situation. In A there’s an obscuring-conservation-status on California and an open-conservation-satus on Nevada, but nothing on the US or Oregon. Only obs in California are actually obscured. In B there’s an open-conservation-status on the US but still only obs in California are actually obscured. In C there’s an obscuring-conservation-status on the US which acts to obscure all obs, even those in Nevada. 
Screen Shot 2021-08-20 at 4 24 50 PM
Instead we'd like the following behavior. A and B are the same, but in C, the open-conservation-status on Nevada would take precedence over the obscuring-conservation-status on the US.
Screen Shot 2021-08-20 at 4 27 36 PM
This would deal with the longstanding ‘rabbit is endangered globally but a big invasive species in australia’ issue
but is also necessary to create a simpler policy for deciding what gets obscured that doesn't require cracking down on existing Natureserve G and N ranked statuses that are currently complicating the destiny of descendant S ranked statuses.

kueda:
Allows conservation statuses associated with country- and state-level places to override global.

Pending questions:

- [x] This allows country-level and state-level statuses to override a global private status; I assume that's ok?
- [x] Should we consider continents? Currently if continent is obscured but country is open, the obs will be obscured b/c continents are just treated like every other place-based status, i.e. they're not overridedn by countries and states

Closes #3185


kueda:
Scott says overriding a global private status is ok (only a handful of statuses like this), added support for continents.

Issue #3183, opened 2021-08-20
Move from machinist to factory_bot for test data


kueda:
As proposed and begun by @todtb in https://github.com/inaturalist/inaturalist/pull/3175, machinist isn't maintained, factory_bot is, and factory_bot seems to enable some significant performance improvements to our test suite, so let's switch.

PR #3210, opened 2021-09-06
Use observation stubs where possible in specs

todtb:
Target slower specs in the "o-m" range. Adds in more factories to support.

Related: #3183

### DIFF
--- a/app/models/observation_field_value.rb
+++ b/app/models/observation_field_value.rb
@@ -42,8 +42,6 @@ class ObservationFieldValue < ApplicationRecord
     record.user_id != subscribable.user_id
   }
 
-  attr_accessor :updater_user_id
-
   include Shared::TouchesObservationModule
   include ActsAsUUIDable
   
@@ -96,12 +94,11 @@ class ObservationFieldValue < ApplicationRecord
   end
 
   def set_user
-    if updater_user_id
-      self.user_id ||= updater_user_id
-      self.updater_id = updater_user_id
+    if updater
+      self.user ||= updater
     elsif observation
-      self.user_id ||= observation.user_id
-      self.updater_id ||= user_id
+      self.user ||= observation.user
+      self.updater ||= user
     end
     true
   end
--- a/spec/factories/place_geometries.rb
+++ b/spec/factories/place_geometries.rb
@@ -1,0 +1,6 @@
+FactoryBot.define do
+  factory :place_geometry do
+    place
+    source
+  end
+end
--- a/spec/factories/place_geometries.rb
+++ b/spec/factories/place_geometries.rb
@@ -2,5 +2,11 @@ FactoryBot.define do
   factory :place_geometry do
     place
     source
+
+    transient { ewkt { "MULTIPOLYGON(((0 0,0 1,1 1,1 0,0 0)))" } }
+    after :build, :stub do |place_geometry, eval|
+      georuby_geom = GeoRuby::SimpleFeatures::Geometry.from_ewkt(eval.ewkt)
+      place_geometry.geom = RGeo::WKRep::WKBParser.new.parse(georuby_geom.as_wkb)
+    end
   end
 end
--- a/spec/models/observation_spec.rb
+++ b/spec/models/observation_spec.rb
@@ -2527,7 +2527,7 @@ describe Observation do
 
     before { allow(observation).to receive(:set_geom_from_latlon) }
 
-    it "sets geom on save" do
+    it "gets called on save" do
       observation.run_callbacks :save
 
       expect(observation).to have_received :set_geom_from_latlon
--- a/spec/models/observation_spec.rb
+++ b/spec/models/observation_spec.rb
@@ -1,6 +1,8 @@
 # encoding: UTF-8
 require File.dirname(__FILE__) + '/../spec_helper.rb'
 
+include ElasticStub
+
 describe Observation do
   it { is_expected.to belong_to :user }
   it { is_expected.to belong_to(:community_taxon).class_name 'Taxon' }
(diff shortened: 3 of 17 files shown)


app/models/observation_field_value.rb
```diff
@@ -96,12 +94,11 @@ def strip_value
...
-    if updater_user_id
-      self.user_id ||= updater_user_id
-      self.updater_id = updater_user_id
+    if updater
+      self.user ||= updater
     elsif observation
-      self.user_id ||= observation.user_id
-      self.updater_id ||= user_id
+      self.user ||= observation.user
+      self.updater ||= user
```
todtb, line 101: Noticed this when working on OFV, probably pre-dating `has_updater` inclusion. Seems like a fairly easy clean up. This and the associated controller params are the only non-spec changes here.

spec/models/observation_spec.rb
```diff
@@ -1532,61 +1534,59 @@
   end
 
   describe "species_guess parsing" do
-    before(:each) do
-      @observation = Observation.make!
-    end
-  
+    stub_elastic_index! Observation, Taxon
```
todtb, line 1537: Used where ES is not required in a larger spec where it is used (spec/models/observation_spec.rb:54)

spec/models/observation_spec.rb
```diff
@@ -2425,67 +2456,110 @@ def stub_api_response_for_observation( o )
...
-  describe "geom" do
-    it "should be set with coords" do
-      o = Observation.make!(:latitude => 1, :longitude => 1)
-      expect(o.geom).not_to be_blank
-    end
-  
-    it "should not be set without coords" do
-      o = Observation.make!
-      expect(o.geom).to be_blank
+  describe "#set_geom_from_latlon" do
```
todtb, line 2459: The coverage of the method is maintained below, however added this as each `it` block was inherently testing that the callback existed. Added this in since the stubs below now call the method directly rather than working through the entire callback chain.

It borders the line of testing the framework, however does cover potential regressions on the existence (and placement) of the callbacks.

spec/models/observation_spec.rb
```diff
@@ -3572,151 +3667,193 @@ def stub_api_response_for_observation( o )
...
+      context "and no editing preferred" do
+        let(:observer_preference) { User::PREFERRED_OBSERVATION_FIELDS_BY_OBSERVER }
+        let(:observer) { build_stubbed :user, preferred_observation_fields_by: observer_preference }
+
+        it { is_expected.to be true}
+      end
     end
   end
 
   describe "mappable" do
```
todtb, line 3725: This describe is a good example of what I've tried to break out. When testing the behavior of the method `calculate_mappable` we can use stubbed objects, however when introducing interactions such as "with a photo" or "should not be mappable if community disagrees with taxon", the specs continue to create fully persisted objects to maintain the integrative approach (if it's not already covered in request/controller specs).

spec/support/elastic_stub.rb
```diff
@@ -0,0 +1,11 @@
+module ElasticStub
```
todtb, line 1: Used for stubbing persistence yet still relying on ES, right now just in `describe Observation, "probably_captive?"`, however I think this could be a useful stub in the future.

kueda: Just so I understand this, this is stubbing observations in the ES index by instantiating them and indexing them, but not committing them to the database, allowing you to test indexing without the overhead of ActiveRecord callbacks or writing the database, right? Could you explain what this is doing in a comment in the code? Because I guarantee I'm going to encounter this in a few years and forget what this does. I agree this seems like a cool pattern for these kinds of tests.

app/models/observation_field_value.rb
kueda: Good catch.

spec/factories/place_geometries.rb
```diff
@@ -0,0 +1,6 @@
+FactoryBot.define do
+  factory :place_geometry do
+    place
+    source
```
kueda, line 4: I'm surprised this works without a `geom` value, but I guess we can deal with that when it comes up.

spec/models/observation_spec.rb
kueda: Works for me, though I might rename the test to `it "gets called on save"` since we're no longer testing whether `geom` is actually getting set here.

spec/models/observation_spec.rb
kueda: Seems prudent from a test performance perspective, and from the perspective of continuous integration. Definitely not how I think when I'm trying to do test-driven development, though, when I would never bother to test the framework, as you put it, but rather want to focus on testing my assumptions about the behavior of the code I'm about to write.

kueda (changes requested):
All looking good to me, I'd just appreciate having that comment to help future folks know what's going on.

spec/models/observation_spec.rb
todtb: Yeah makes sense. Especially because the actual testing of the `#set_geom_from_latlon` is happening below in  `geom` and `private_geom` blocks.

spec/support/elastic_stub.rb
todtb: Yeah absolutely. Classic case of it making sense when you're already deep in it, but a total mystery a few months later 😄

spec/factories/place_geometries.rb
todtb: Yeah, I see that validation now. I think it's only used right now in the context of creating a place (`after(:build) { |place| build :place_geometry, place: place }` ), but only built, not persisted. I'll see what we can put in place for full creation.

Issue #958, opened 2016-04-07
Style header/subnav for the following views consistently with profile/journal


loarie:
Here's how the header/subnav looks on the profile
<img width="1215" alt="screen shot 2016-04-07 at 11 19 09 am" src="https://cloud.githubusercontent.com/assets/598026/14362104/f2e88170-fcb2-11e5-910e-4a1e2904544d.png">
and here is the existing header/subnav on:
<a href="http://www.inaturalist.org/observations/loarie">observations/user</a>
<img width="1009" alt="screen shot 2016-04-07 at 11 19 24 am" src="https://cloud.githubusercontent.com/assets/598026/14362106/f2eb6516-fcb2-11e5-84ed-a012e73a0c73.png">
<a href="http://www.inaturalist.org/calendar/loarie">calendar/user</a>
<img width="1019" alt="screen shot 2016-04-07 at 11 19 38 am" src="https://cloud.githubusercontent.com/assets/598026/14362108/f2ee0c94-fcb2-11e5-9597-3a31470e9931.png">
<a href="http://www.inaturalist.org/faves/loarie">faves/user</a>
<img width="1093" alt="screen shot 2016-04-07 at 11 19 57 am" src="https://cloud.githubusercontent.com/assets/598026/14362107/f2ebc9a2-fcb2-11e5-8280-361d781c1e4f.png">
<a href="http://www.inaturalist.org/lists/loarie">lists/user</a>
<img width="911" alt="screen shot 2016-04-07 at 11 20 22 am" src="https://cloud.githubusercontent.com/assets/598026/14362105/f2eb1746-fcb2-11e5-8a62-56e6a5a1361d.png">
<a href="http://www.inaturalist.org/identifications/loarie">identifications/user</a>
<img width="970" alt="screen shot 2016-04-07 at 11 20 43 am" src="https://cloud.githubusercontent.com/assets/598026/14362109/f2ef6652-fcb2-11e5-970b-c8417363a0e7.png">
<a href="http://www.inaturalist.org/projects/user/loarie">projects/user</a>
<img width="1025" alt="screen shot 2016-04-07 at 11 20 50 am" src="https://cloud.githubusercontent.com/assets/598026/14362110/f2fc2c70-fcb2-11e5-877d-080648f12944.png">

seanclifford:
I'm planning to look into this - it's been bugging me for a while. Projects and Lists have been completed since this issue was created.
Left to do:
- [ ] Edit Observations
[...]

PR #3213, opened 2021-09-08
Bootstrapify identifications by login

seanclifford:
Made "Identifications by login" page use the bootstrap subnav header as part of issue #958.

I started by just changing the header, but found the default layout not wide enough to hold the bootstrap subnav header without wrapping, so figured I'd try applying bootstrap styles to the whole page, while trying to keep the original design.
I ended up having to apply a few css hacks that I'm not too happy with, but thought I'd put it out there and see what you think...

Summary of changes:
- Applied bootstrap layout, and used bootstrap grid column styles
- Updated `by_login.html` and `_identification_with_observation.html` to use Haml
- Removed some unused dead code around activity streams because it was referencing `_identification_with_observation`
- Fixed bootstrap subnav so the IDs tab was active when displaying identifications by login
- Moved css into a separate file
- Applied a number of css workarounds to make the page look OK

I tested that pagination, Agree button, Withdraw button and small page sizes work OK.

Here's how it looks
image


app/assets/stylesheets/identifications_by_login.css
```diff
@@ -0,0 +1,43 @@
...
+.bootstrap .readable {font-size:125%;}
+.bootstrap .observations.mini .icon {top:8px;}
+.bootstrap .needs_id_notice { 
+  background-color: #FFEE91;
+  color: #857430;
+}
+.bootstrap .research_grade_notice { 
+  background-color: #DCEAA3;
+  color: #529214;
+}
```
seanclifford, line 34: These are the extra css workarounds/hacks I added.

app/views/identifications/_identification_with_observation.html.erb
```diff
@@ -1,19 +0,0 @@
-<%- identification ||= identification_with_observation || @identification -%>
-<div class="mini <%= 'compact' if request.format.mobile? %> observations observationcol column span-8">
```
seanclifford, line 2: I left out the compact style in the haml. I'm guessed that it's not required with bootstrap columns resizing?

app/assets/stylesheets/identifications_by_login.css
seanclifford, line 34: These are the extra css workarounds/hacks I added.

kueda (changes requested):
In addition to the comments, two requests:

- [ ] Move `app/assets/stylesheets/identifications_by_login.css` to `app/assets/stylesheets/identifications/by_login.scss`
- [ ] Make sure all files end with a newline

Also, not something you need to do here, but in general I prefer to `git mv` files when converting from ERB to HAML, as opposed to deleting the ERB versions and adding the HAML ones. Helps preserve the history.

app/assets/stylesheets/identifications_by_login.css
```diff
@@ -0,0 +1,43 @@
...
+  text-decoration: underline;  
+  font-size: smaller;  
+  cursor: pointer; 
+  padding:0; 
+  border:0; 
+  float:none; 
+  background-color: transparent;
+}
+
+.bootstrap .readable {font-size:125%;}
```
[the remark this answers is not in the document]
seanclifford: Removed. 
I'm not sure which alignment you're referring to though - I didn't spot an improvement after it was gone.

seanclifford:
Thank you for the review @kueda!
I think I've made all requested changes now.
On the `git mv` front - I'm happy to redo the commits and create a move commit before applying the haml changes if you'd like the history preserved. I'd have to recreate the branch and PR I imagine though.

app/assets/stylesheets/identifications_by_login.css
```diff
@@ -0,0 +1,43 @@
...
+.bootstrap .observations.mini .icon {top:8px;}
+.bootstrap .needs_id_notice { 
+  background-color: #FFEE91;
+  color: #857430;
+}
+.bootstrap .research_grade_notice { 
+  background-color: #DCEAA3;
```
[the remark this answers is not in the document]
seanclifford: I went with the white on green. I think it's an improvement!

app/assets/stylesheets/identifications/by_login.scss
```diff
@@ -0,0 +1,46 @@
...
+  margin-bottom: 0;
+  border: 0;
+}
+
+.identification {
+  padding: 10px;
+  background-color: #eee;
+
+  .identification_image { width: 75px; }
+  .identification_body { float: left; }
```
kueda, line 14: Sorry I missed this on the first pass: remove this and add `display: flex` to `.identification` to achieve the same effect while not wrapping at smaller window sizes.

kueda (changes requested):
Sorry, just that one more thing. No need to redo things with `git mv`, just something to keep in mind for similar work in the future.

app/assets/stylesheets/identifications/by_login.scss
seanclifford: Fixed, thanks. 
I put `display: flex` under the bootstrap section, as the bootstrap_bundle css of `display: block` seemed to be overriding it.

Issue #3212, opened 2021-09-07
Improvements to taxon_ranges


loarie:
Some improvements to accomodate process for manually updating taxon ranges from IUCN described here:
https://forum.inaturalist.org/t/updating-iucn-red-list-rangemaps/25817
1. Ensure no more than one taxon_range per taxon
2. Add 'guidelines' taxon_range new and edit similar taxon_names new and edit
3. add user_id, updated_at, and updater_id to taxon_range
4. add `Relation to IUCN`: `IUCN` / `Error in IUCN` / `Not in IUCN`
5. add ability to remove geometry from taxon range record from taxon_range/edit

loarie:
merged https://github.com/inaturalist/inaturalist/commit/d42ce28253c004b92759f707197bf01b862cb75b

PR #3215, opened 2021-09-08
Taxon range improvements

loarie:
made changes described here https://github.com/inaturalist/inaturalist/issues/3212


loarie:
@kueda I added taxon_range/show to this pull request and I think its ready for re-review. If you could also give a shot at adding a link from the taxon maps to taxon_range/show that would be awesome
Screen Shot 2021-09-08 at 7 03 48 PM

app/views/taxon_ranges/_form_fields.html.haml
```diff
@@ -0,0 +1,33 @@
...
+          = f.radio_button :iucn_relationship, 0, :skip_builder => true
+          = t(:this_is_an_iucn_redlist_range_map)
+      .radio
+        %label.inline
+          = f.radio_button :iucn_relationship, 1, :skip_builder => true
+          = t(:iucn_redlist_range_map_for_this_taxon_has_issues)
+      .radio
+        %label.inline
+          = f.radio_button :iucn_relationship, 2, :skip_builder => true
+          = t(:taxon_not_in_the_iucn_redlist)
```
kueda, line 24: If you use `BootstrapFormBuilder` instead of `DefaultFormBuilder` you can re-write this as `= f.radio_button :iucn_relationship, 2, label: t(:taxon_not_in_the_iucn_redlist)`, plus you'll get the default styles for the other form fields. The source choose will cause problems, though. You might have to hard code the width of that input for it to work properly.

kueda (changes requested):

config/locales/en.yml
```diff
@@ -2085,6 +2091,18 @@ en:
...
+    <ul>
+      <li><p>
+        <b>IUCN RedList range map for this taxon is unsutable:</b> for example, you’ve manually narrowed
+        it to reflect a taxonomic split not represented in the IUCN RedList, or the IUCN
+        RedList range map has major errors making its display problematic.</p></li>
+      <li><p>
+        <b>Taxon is not in the IUCN RedList:</b> for example, the taxon is lumped with another taxon in the IUCN
+        RedList so you’ve manually carved it off here, or if the taxon lacks an IUCN RedList assessment
+        and range map.</p></li>
+    </ul>
```
kueda, line 2105: Again, I would avoid embedding HTML in the translatable text because any changes in markup will require re-translation, and translators tend to make mistakes with the markup. I think it also helps to name view-specific keys like this with some consistent prefix so they stay together in the source file, so like

```yaml
taxon_range_guidelines_if_not_uploading: "If you are not uploading an IUCN RedList range map, this should be because:"
taxon_range_guidelines_iucn_unsuitable: IUCN RedList range map for this taxon is unsutable
```

etc. Also, FWIW, have a translation for bold text followed by a colon: `bold_label_colon_value_html`.

app/views/taxon_ranges/_form_fields.html.haml
```diff
@@ -0,0 +1,27 @@
...
+        = link_to f.object.range.url, f.object.range.url
+        = link_to_function t(:replace_parentesis), "$(this).parents('.field').after('#{f.file_field(:range)}'); $(this).parents('.field').hide()", :class => "small"
+        = f.check_box :range_delete, label: t(:delete_range), label_after: true
+    - else
+      = f.file_field :range
+    = f.form_field :iucn_relationship, :label => false do
+      = label_tag :is_this_an_iucn_redlist_range_map_and_if_not_why, t(:is_this_an_iucn_redlist_range_map_and_if_not_why)
+      = f.radio_button :iucn_relationship, TaxonRange::IUCN_REDLIST_MAP, label: t(:this_is_an_iucn_redlist_range_map)
+      = f.radio_button :iucn_relationship, TaxonRange::IUCN_REDLIST_MAP_HAS_ISSUES, label: t(:iucn_redlist_range_map_for_this_taxon_has_issues)
+      = f.radio_button :iucn_relationship, TaxonRange::NOT_ON_IUCN_REDLIST, label: t(:taxon_not_in_the_iucn_redlist)
```
kueda, line 15: Since `iucn_relationship` can be null, I think we need a fourth option here for that, like "other" or "unknown" or something. In addition to "unknown" can't a range be for a taxon that *is* on the Red List but the range didn't come from the IUCN?

app/views/taxon_ranges/show.html.haml
```diff
@@ -0,0 +1,52 @@
+- content_for :title do
+  Taxon Range for
+  = render "taxa/taxon.txt.erb", taxon: @taxon_range.taxon
```
kueda, line 3: Needs to be globalized.

app/views/taxon_ranges/show.html.haml
```diff
@@ -0,0 +1,52 @@
...
+  .row.cf#top
+    .col-sm-3.col-xs-12.pull-right
+      .panel.panel-info
+        .panel-heading
+          %h3.panel-title
+            =t :about_taxon_ranges
+        .panel-body
+          %p
+            =raw t :about_taxon_ranges_desc, url: "/pages/taxon_ranges"
+          %img.media-object{ src: "https://static.inaturalist.org/wiki_page_attachments/2817-original.png", class: "img-responsive" }
```
kueda, line 21: I don't think this needs to block merge, but this is a bad habit: images that the app uses should be checked into the repo and served via the asset pipeline.

config/locales/en.yml
```diff
@@ -230,6 +230,10 @@ en:
   # should work here
   about_place: "About %{place}"
   about_places: About Places
+  about_taxon_ranges: About Taxon Ranges
+  about_taxon_ranges_desc: |
+    Taxon Ranges convey the global distribution of a taxon. Taxon ranges are
+    typically sourced from the IUCN RedList. <a href='%{url}'>Learn more</a>
```
kueda, line 236: Translation keys that contain HTML should end in `_html`. That way you don't have to use `raw` when you use them in the templates.

config/locales/en.yml
```diff
@@ -2167,6 +2185,7 @@ en:
   iphone: iPhone
   is_not_following_anyone: is not following anyone.
   is_the_evidence_provided_enough_to_confirm_this_is_taxon: Is the evidence provided enough to confirm this is %{taxon}?
+  is_this_an_iucn_redlist_range_map_and_if_not_why: Is this an IUCN RedList range map, and if not why?
```
kueda, line 2188: "RedList" should be "Red List"

config/locales/en.yml
```diff
@@ -2178,6 +2197,7 @@ en:
     site curator. If you think this is a mistake, please
     <a href="mailto:%{help_email}" class="contact">contact us</a>.
     <a href="%{manage_flags_path}">Manage flags</a>
+  iucn_redlist_range_map_for_this_taxon_has_issues: IUCN RedList range map for this taxon has issues
```
kueda, line 2200: "RedList" should be "Red List"

app/models/taxon_range.rb
```diff
@@ -1,9 +1,14 @@
...
   belongs_to :taxon
   belongs_to :source
+  belongs_to :user
+  has_updater
   has_many :listed_taxa, :dependent => :nullify
-  
+
+  validates_uniqueness_of :taxon_id, :message => "taxon range already exists"
+  validates_presence_of :taxon
+  validate :species_common_name_cannot_match_taxon_name
```
kueda, line 10: This... isn't actually what this validation is doing, right?

app/models/taxon_range.rb
```diff
@@ -1,9 +1,14 @@
 class TaxonRange < ApplicationRecord
-  
   belongs_to :taxon
   belongs_to :source
+  belongs_to :user
+  has_updater
   has_many :listed_taxa, :dependent => :nullify
-  
+
+  validates_uniqueness_of :taxon_id, :message => "taxon range already exists"
```
kueda, line 8: This results in an error like "Taxon taxon range already exists." I would leave the message blank, or change it to "already has a range" so the message reads "Taxon already has a range" instead. Actually, this text should be globalized, but I'm not actually sure how to do that properly off the top of my head.

config/locales/en.yml
```diff
@@ -2658,6 +2678,7 @@ en:
   multiple_select_option: -- multiple --
   multiple_taxa_for_taxon_change: "Multiple taxa (input for swap/merge, output for split)"
   multiple_taxon_ids_separated_by_commas: 'Multiple taxon IDs separated by commas, e.g. 123,654'
+  must_be_an_accepted_value: must be an accepted value
```
kueda, line 2681: Definitely good to translate these errors, but this translation will not be used when you do `errors.add(:iucn_relationship, :must_be_an_accepted_value)`. For that to work it needs to exist at `activerecord.errors.models.taxon_range.attributes.iucn_relationship.must_be_an_accepted_value`

config/locales/en.yml
```diff
@@ -3742,6 +3764,10 @@ en:
   please_fix_your_csv_and_try_again: Please fix your CSV and try again.
   please_hold_on_while_the_file: Please hold on while the file is generated. It can take a couple minutes if you have a lot of observations.
   please_hold_on_while_the_file_is_generated: Please hold on while the file is generated.  It can take a couple minutes for large lists.
+  please_provide_details: |
+    Please provide details on how the map was derived in description and link to any relevant flags
+    where taxonomic splits or why the IUCN RedList range was determined to be problematic.
+    <a class="readmore" href="%{url}">Learn more about taxon ranges</a>.
```
kueda, line 3770: Proposed re-write:

> Please use the description to explain how the map was derived and link to any relevant flags, taxonomic splits, or other discussions of why the IUCN Red List range required alteration.

config/locales/en.yml
```diff
@@ -4660,6 +4688,12 @@ en:
      planet. Most check lists are works in progress, but if there's one
      that is comprehensive, i.e. it lists <em>all</em> the representatives of
      its taxon in a place, the Identotron will show a notice.
+  the_iucn_redlist_is_the: |
+    The <a href="https://www.iucnredlist.org/">IUCN RedList</a> is the typical source of taxon ranges.
+    Please indicate whether you are adding an IUCN RedList range and if not how what you are
+    adding relates to the IUCN RedList. This will help inform processes
+    of automatically update ranges sourced to the IUCN RedList. If you are adding an IUCN RedList
+    range please properly indicate the IUCN RedList version (e.g. 2021-2) in the source.
```
kueda, line 4696: "RedList" should be "Red List"

config/locales/en.yml
```diff
@@ -4725,6 +4759,7 @@ en:
   this_is_a_more_convenient_way_to_find_a_taxon_id: |
     This is a more convenient way to find a taxon ID. It's the taxon ID
     that actually gets saved. This search only works for active taxa.
+  this_is_an_iucn_redlist_range_map: This is an IUCN RedList range map
```
kueda, line 4762: "RedList" should be "Red List"

app/assets/stylesheets/taxon_ranges/taxon.css
```diff
@@ -0,0 +1 @@
+input.ui-chooser-input { max-width:400px; }
```
kueda, line 1: Since this has nothing to do with the taxon the range is associated with, how about renaming this file to `form.css`. Also needs a newline at the end.

kueda (changes requested):

loarie:
@kueda I think this is ready to merge if you're happy with it

app/views/taxon_ranges/_form_fields.html.haml
loarie: My vision was that we should work towards every taxon range having one of these 3 values:
IUCN_REDLIST_MAP
IUCN_REDLIST_MAP_HAS_ISSUES
NOT_ON_IUCN_REDLIST
So while we display unknown for the many existing records where this is unknown, we want to encourage curators to choose one of those 3 if they are going to update that attribute.
For the range is on the Red List but the range didn't come from the IUCN case it would be IUCN_REDLIST_MAP_HAS_ISSUES because there's still some reason why someone felt the IUCN range was unsuitable to display and therefore opted for an alternative.

kueda: Ah, that wasn't clear to me. Could we change that text to read, "This is not an IUCN Red List range map because that map has problems" or something? Maybe the last could be "This is not an IUCN Red List range map because this taxon is not on the IUCN Red List"?

loarie: I changed it to that but I'm going with "unsuitable" rather than "has problems" to make it more subjective to iNat's purposes

kueda (approved):

PR #3224, opened 2021-09-13
Atlas improvements

loarie:
### 1. Made it so that atlases for taxa coarser than species don't roll up listed taxa for taxon descendants
Atlases roll up listed taxa for taxon descendants. The idea is that if a species has no listed_taxa for place X but it has sub-species that do, place X should still be calculated as an atlas presence_place.

While atlases were mostly conceived for species, they are our only means of reassigning input_taxon IDs to unambiguous output_taxa as part of a taxon_split which sometimes needs to be done for taxa coarser than species (e.g. families etc.). This method of rolling up listed taxa for taxon descendants for taxa with many many descendants (e.g. large families) isn't scaling well and times out. 

This change makes it so that atlases for taxa coarser than species only calculate presence places based on listed_taxa associated with the taxon itself, ignoring listed_taxa associated with the taxon's descendants.

### 2. Made atlases show verifiable observations only
Previously taxon maps only showed verifiable observations. This was appropriate for atlases which are meant to convey biogeography complicated by observations of captive observations. When the React maps were extended to allow users to toggle verifiable observations, observations without media, and captive/cultivated observations, the jquery maps that persist on atlases and taxon changes changed to display all observations including casual observations.

This change stops short of adding the 3 options in the Reach maps, but makes it so that the maps on atlases show Verifiable observations _instead of_ all observations. To accomodate this, I added a parameter to the `setup_map_tag_attrs` application helper.

~~3. Made it so that atlases only flag observations with <5k accuracy~~
The utility of using atlases to mark out of atlas observations is being diminished by observations with really large accuracies. I'd like to exclude these from the existing atlas marking functionality. But I forgot that the acc_below parameter will also exclude acc=nil observations which we don't want to exclude. so I removed this piece until we can add new API params that act as acc_below OR acc=nil

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -727,7 +727,7 @@ module ApplicationHelper
       "taxon-range-layer-label" => I18n.t("maps.overlays.range"),
       "taxon-places-layer-label" => I18n.t("maps.overlays.checklist_places"),
       "taxon-places-layer-hover" => I18n.t("maps.overlays.checklist_places_description"),
-      "taxon-observations-layer-label" => I18n.t("maps.overlays.observations"),
+      "taxon-observations-layer-label" => I18n.t( options[:taxon_observations_layer_label] ? options[:taxon_observations_layer_label] : "maps.overlays.observations" ),
       "all-layer-label" => I18n.t("maps.overlays.all_observations"),
       "all-layer-description" => I18n.t("maps.overlays.every_publicly_visible_observation"),
       "gbif-layer-label" => I18n.t("maps.overlays.gbif_network"),
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -727,7 +727,7 @@ module ApplicationHelper
       "taxon-range-layer-label" => I18n.t("maps.overlays.range"),
       "taxon-places-layer-label" => I18n.t("maps.overlays.checklist_places"),
       "taxon-places-layer-hover" => I18n.t("maps.overlays.checklist_places_description"),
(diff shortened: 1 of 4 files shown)


app/helpers/application_helper.rb
```diff
@@ -727,7 +727,7 @@ def setup_map_tag_attrs(options = {})
       "taxon-range-layer-label" => I18n.t("maps.overlays.range"),
       "taxon-places-layer-label" => I18n.t("maps.overlays.checklist_places"),
       "taxon-places-layer-hover" => I18n.t("maps.overlays.checklist_places_description"),
-      "taxon-observations-layer-label" => I18n.t("maps.overlays.observations"),
+      "taxon-observations-layer-label" => I18n.t( options[:taxon_observations_layer_label] ? options[:taxon_observations_layer_label] : "maps.overlays.observations" ),
```
kueda, line 730: Can you make the entire content of the label optional instead of the translation key? Like

```ruby
"taxon-observations-layer-label" => options[:taxon_observations_layer_label].blank? ? I18n.t( "maps.overlays.observations" ) : options[:taxon_observations_layer_label],
```

Having stuff other than a string literal inside the `t( )` makes it much harder to analyze the code for things like unused translations.

app/views/atlases/show.html.haml
```diff
@@ -41,11 +41,12 @@
         map_attributes = setup_map_tag_attrs(
           min_zoom: 1,
           show_all_layer: false,
+          taxon_observations_layer_label: "verifiable_observations",
           taxon_layers: [ {
             taxon: @atlas.taxon,
             places: false,
             ranges: true,
-            observations: true,
+            observations: {verifiable: true},
```
kueda, line 49: Please add spaces inside the curly brackets.

kueda (changes requested):

kueda (approved):

PR #3226, opened 2021-09-14
Analyze taxon split

loarie:
tool on draft taxon splits to estimate the destinations of input taxon IDs after the split is committed (to specific output taxa or to the common ancestor) to help determine whether the split is necessary, optimize atlases, and minimize ID coarsening from splits.

See [this doc]( https://docs.google.com/document/d/1yuUq-AtagW_oaV978zd8VzL-LKT9BBLN_HGnJ2v6Cb4/edit?usp=sharing) for use cases of this tool while making complex ancestry changes.


app/controllers/taxon_changes_controller.rb
```diff
@@ -357,6 +357,37 @@ def group
       format.html { render layout: "bootstrap" }
     end
   end
+
+  def analysis_with_url( row )
```
kueda, line 361: This should be private. Also could use a more descriptive name. Maybe `identification_analysis_with_url` or `analyze_ids_row`

app/controllers/taxon_changes_controller.rb
```diff
@@ -357,6 +357,37 @@ def group
...
+      observation_params = {
+        id: row[:id_obs].join(",")
+      }
+      url = observations_path( observation_params )
+    end
+    return { taxon_id: row[:taxon_id], id_count: row[:id_count], url: url }
+  end
+
+  def analyze_ids
+    return false unless @taxon_change.type == "TaxonSplit"
```
kueda, line 374: In addition to returning false this needs to render a 422 response.

app/controllers/taxon_changes_controller.rb
```diff
@@ -357,6 +357,37 @@ def group
...
+    else
+      observation_params = {
+        id: row[:id_obs].join(",")
+      }
+      url = observations_path( observation_params )
+    end
+    return { taxon_id: row[:taxon_id], id_count: row[:id_count], url: url }
+  end
+
+  def analyze_ids
```
kueda, line 373: Since this is kind of a performance issue, it should require auth, and maybe require a curator.

kueda (changes requested):

app/controllers/taxon_changes_controller.rb
```diff
@@ -358,34 +358,35 @@ def group
     end
   end
 
-  def analysis_with_url( row )
-    if row[:id_count] == 0
-      url = nil
+  def analyze_ids
+    unless logged_in? && current_user.is_curator?
+      response = { json: I18n.t( :only_curators_can_access_that_page ), status: :unprocessable_entity }
```
kueda, line 363: Without a `respond_to` block and a `return` here, you're still running the risk of a non-curator executing the rest of this method. You could also just add this action to `before_action :curator_required` above and omit this entirely, though I guess you'd have to change `curator_required` to respond with JSON.

app/controllers/taxon_changes_controller.rb
```diff
@@ -399,6 +400,14 @@ def load_taxon_change
       ).first
   end
 
+  def stub_taxon_split
+    input_taxon = Taxon.where( id: params[:inputTaxonId].to_i ).first
+    output_taxa = Taxon.where( id: params[:outputIds].map{|i| i.to_i} )
+    @taxon_change = TaxonSplit.new
+    @taxon_change.taxon = input_taxon
+    @taxon_change.new_taxa << output_taxa
```
kueda, line 408: Doesn't using `<<` actually save the `@taxon_change` here and save the `TaxonChangeTaxa` records? I would have thought you would need to do something like `output_taxa.each { |t| @taxon_change.taxon_change_taxa.build( taxon: t )` to make a record without saving it.

app/views/taxon_changes/_form.html.haml
```diff
@@ -0,0 +1,154 @@
...
+  taxon_change_type = taxon_change.class.name
+
+= form_for taxon_change, as: :taxon_change,
+    url: taxon_change.new_record? ? taxon_changes_path : taxon_change_path( taxon_change ),
+    builder: DefaultFormBuilder,
+    html: { multipart: true } do |f|
+  = error_messages_for :taxon_change
+  
+  .column.span-24
+    -if taxon_change.committed?
```
kueda, line 27: Space after the `-`, please, here and everywhere else in HAML.

app/views/taxon_changes/_form.html.haml
```diff
@@ -0,0 +1,154 @@
...
+    -else
+      = f.hidden_field :type
+
+    %fieldset#sourcing{ class: ( taxon_change.committed? ? "last column span-24" : "last column span-16" ) }
+      %legend
+        =t :sourcing
+        - helptip_for "source" do
+          =t :please_add_a_citation_for_this_change
+      .stacked
+        = render partial: 'sources/nested_form_fields', locals: { f: f }
```
kueda, line 55: Please use double quotes, here and everywhere else.

app/views/taxon_changes/edit.html.haml
```diff
@@ -0,0 +1,13 @@
...
+  - @title = capture do
+    =t :editing
+    =t @taxon_change.class.name.underscore
+    = surround '(', ')' do
+      = @taxon_change.committed? ? t(:committed_on, date: l( @taxon_change.committed_on ) ) : t(:draft)
+  = strip_tags @title
+= render partial: "form_head"
+.last.column.span-24#pageheader
+  %span.breadcrumbs
+    = link_to t( :back_to_x, :noun => t( @taxon_change.class.name.underscore ) ), taxon_change_path( @taxon_change ), class: 'back crumb'
```
kueda, line 11: I'd just use `t( "back" )` or add `t( "back_to_taxon_change" ) or something. This is the kind of thing that drives translators crazy with gender agreement and such.

kueda (changes requested):

app/controllers/taxon_changes_controller.rb
loarie: doesn't curator action already apply since its not excepted?
```before_action :curator_required, :except => [:index, :show, :commit_for_user, :commit_records, :group]```

Comment (kueda) [reply app/controllers/taxon_changes_controller.rb:363]: Ah, good point!

Comment (kueda) [reply app/controllers/taxon_changes_controller.rb:363]: But that means you can just remove this, right? A non-curator can't get this far.

File: app/controllers/taxon_changes_controller.rb [hunk 9]
```diff
@@ -357,6 +357,41 @@ def group
...
+      analysis_data = {
+        analysis_header: analysis_header,
+        analysis_table: analysis_table
+      }
+      response = { json: analysis_data, status: :ok }
+    else
+      response = { json: I18n.t( :only_curators_can_access_that_page ), status: :unprocessable_entity }
+    end
+    respond_to do |format|
+      puts response[:json].to_json
```
kueda, line 391: Pls. remove the debugging code.

kueda (changes requested):

kueda (approved):

PR #3232, opened 2021-09-19
Bootstrapify Calendar, including bootstrap subnav

seanclifford:
Implemented as part of issue #958 to standardize the subnav component used on the site.

Style is mostly the same as before with bootstrap grid classes used.
Resizes reasonably nicely now, and the year header and footer dropdown are always centered too.
Additionally moved the custom page CSS to a an `scss` file.

Note: As previously discussed, I made separate commits for the file renames to `haml`, but I believe history of the renamed files will not be "tracked" if these commits are squashed on merge.


app/assets/stylesheets/calendars/index.scss
```diff
@@ -0,0 +1,38 @@
...
+  display:inline-table;
+
+  caption {
+    text-align: center;
+  }
+  th {
+    text-align: center;
+  }
+  td {
+    font-size:12px;
```
kueda, line 31: Can we get

1. A space after `:`
2. `line-height: 30px` so it's not quite so vertically smushed

app/views/calendars/index.html.haml
```diff
@@ -0,0 +1,41 @@
...
+        = render :partial => 'calendar', :locals => { :observations => @observations_by_month[i+1], :year => @year, :month => i+1, :month_name => t("date_format.month.#{month}") }
+
+  .row.yearnav
+    .col-xs-4
+      %h2.left
+        = link_to "&lsaquo; #{@year - 1}".html_safe, :year => @year - 1
+    .col-xs-4
+      %h4.center
+        %form( method="get" )
+          = select_tag :year, options_for_select(1950..Date.today.year, :selected => @year), "data-autosubmit" => true
```
kueda, line 37: I think you'll get a slightly more bootstrap-y form with something like

```haml
%form{ method: "get", class: "form-inline" }
  = select_tag :year, options_for_select( 1950..Date.today.year, selected: @year ), "data-autosubmit" => true, class: "form-control"
```

kueda (changes requested):
Mostly looking good! Not sure what to do about squashing vs history. I like both, but maybe history is more important.

app/assets/stylesheets/calendars/index.scss
seanclifford: Ah, that looks much better, thanks! 
After changing the line height I noticed the columns were also horizontally smushed around the Thu,Fri,Sat columns. I've made them even widths by adding `table-layout: fixed;`

app/views/calendars/index.html.haml
seanclifford: Thanks, that looks better.
In case you haven't noticed, I'm working out bootstrap as I go along!

seanclifford:
On squashing vs history - I'm personally not too bothered either way... 
Another option would be that I could reauthor the commits down to 2 commits - one for renames, and the rest with the changes on a new branch and PR... But that is a bit of a faff.

kueda (approved):

PR #3239, opened 2021-09-22
don't reassess obs on global non-obscuring status creation/deletion

loarie:
when global non-obscuring conservation statuses are created or destroyed, no need to reassess observation geoprivacies

### DIFF
--- a/app/models/conservation_status.rb
+++ b/app/models/conservation_status.rb
@@ -125,7 +125,7 @@ class ConservationStatus < ApplicationRecord
   def update_observation_geoprivacies
     return true if skip_update_observation_geoprivacies
     # Don't reassess obs for newly created or destroyed non-obscuring global statuses
-    return true if ( saved_change_to_id? || destroyed? ) && geoprivacy == Observation::OPEN && place_id.nil?
+    return true if ( saved_change_to_id? || destroyed? ) && ( geoprivacy.nil? || geoprivacy == Observation::OPEN ) && place_id.nil?
     Observation.delay(priority: USER_INTEGRITY_PRIORITY,
       queue: "throttled",
       unique_hash: {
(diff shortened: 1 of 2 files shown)


loarie:
@kueda ok to merge this in?

app/models/conservation_status.rb
```diff
@@ -124,6 +124,8 @@ def normalize_geoprivacy
 
   def update_observation_geoprivacies
     return true if skip_update_observation_geoprivacies
+    # Don't reassess obs for newly created or destroyed non-obscuring global statuses
+    return true if ( saved_change_to_id? || destroyed? ) && geoprivacy == Observation::OPEN && place_id.nil?
```
kueda, line 128: On staging there are 55991 ConservationStatus records where the `geoprivacy` column is `null` not `open`, so I think you're going to want to check for that here.

spec/models/conservation_status_spec.rb
```diff
@@ -57,6 +57,26 @@
...
+      cs = ConservationStatus.make!( taxon: species )
+      jobs = Delayed::Job.where("created_at >= ?", stamp)
+      expect(jobs.select{|j| j.handler =~ /reassess_coordinates_for_observations_of/m}).not_to be_blank
+    end
+
+    it "should not reassess observations of taxon if open and global" do
+      species = Taxon.make!( rank: Taxon::SPECIES )
+      Delayed::Job.delete_all
+      stamp = Time.now
+      cs = ConservationStatus.make!( taxon: species, geoprivacy: "Open" )
```
kueda, line 74: Use `Observation::OPEN` here instead of the string literal. Testing against `"Open"` is testing against a value that doesn't exist in reality. You should probably also test the situation where geoprivacy is nil.

kueda (changes requested):

kueda (approved):

PR #3246, opened 2021-09-25
26675 updown keyboardshortcut dialog

dooug:
Adding scroll up/down to the Identify Modal  Keyboard shortcut popup.  

Untested. Just thought I'd take a stab at it. 
https://forum.inaturalist.org/t/add-up-down-arrows-in-identify-modal-keyboard-shortcut-popup/26675

### DIFF
--- a/app/webpack/observations/identify/components/observation_modal.jsx
+++ b/app/webpack/observations/identify/components/observation_modal.jsx
@@ -392,6 +392,8 @@ class ObservationModal extends React.Component {
       { keys: ["z"], label: I18n.t( "zoom_photo" ) },
       { keys: ["&larr;"], label: I18n.t( "previous_observation" ) },
       { keys: ["&rarr;"], label: I18n.t( "next_observation" ) },
+      { keys: ["&uarr;"], label: I18n.t( "scroll_up" ) },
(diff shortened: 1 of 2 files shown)


kueda:
Effort appreciated, @dooug, but please don't work on issues that aren't issues in this github repo. In this case, there is no issue 26675 (that's the identifier for a post in the Forum). We ask that proposals for changes to functionality start in the Forum (which you did), and iNat staff will translate them to issues on Github if/when we've decided we want those changes to happen and we can specify what that means to a developer in text. Otherwise, developers may end up spending time working on things that we won't merge.

This is one of those cases, because I don't think there's any need for us to document the default behavior of the browser like the up and down arrows scrolling the primary scrollable component of the visible modal. We should document keyboard shortcuts that are unique to a specific page that we built, but I think it's the user's responsibility to understand the keyboard shortcuts the browser makes available on all pages. Making the up and down arrows work in the observation modal was more like fixing something we broke rather than adding a feature that needs documentation.

FWIW, we have more guidance on contributing code over at https://github.com/inaturalist/inaturalist/blob/main/CONTRIBUTING.md. If you want to work on some other existing issues and need some help getting a development env set up, please let me know and/or message me with an email address I can use to invite to a Slack channel we have set up for this purpose.

PR #3253, opened 2021-09-30
Bootstrapify observations/by_login subnav header

seanclifford:
Changes completed as part of issue #958.

I've done as little as possible for this one in terms of bootstrap styling.
This was because when I applied bootstrap css across the whole page, the text in the columns looked very squished due to the larger font size - even when keeping the existing 100% width. 
I think in order to integrate bootstrap styling to the page successfully it would need a level of redesign that I think is a bit out of scope of the issue I'm looking at.

Having said that, I'm open to suggestions on making further styling changes to the page other than just the subnav header.

Summary of changes:
- Converted `observations/by_login.html.erb` to Haml
- Renamed `observations/by_login.js.erb` to just be `.js`
- Moved most inline javascript from the page to the js file
- Changed subnav to use the bootstrap subnav. Only included bootstrap styles for the subnav section, not the whole page
- Fixed the subnav to not 'activate' the Observations tab when Edit Observations was active


app/views/observations/by_login.html.haml
```diff
@@ -0,0 +1,155 @@
...
+    %ul
+      - for o in @proj_obs_errors_obs
+        %li.clear.stacked
+          = image_and_content observation_image(o, :style => "max-width: 48px; margin: 0 auto; display:block;") do
+            = link_to o.to_plain_s(:user => false), o
+            %div
+              = raw @proj_obs_errors[:errors][o.id].join("<br/>")
+
+.bootstrap
+  .container
```
kueda, line 58: Could you make this `.container-fluid` instead? I know this screws up alignment if you're navigating between pages in the subnav, but I think it looks pretty broken when the observations are fluid but the header is not.

kueda (changes requested):
I think this mostly looks good. As you say, bootstrapifying the content is a heavier lift, as I discovered for myself in https://github.com/inaturalist/inaturalist/pull/2796. This whole page should probably be replaced with some kind of power user editing interface, but that's a whole other thing.

app/views/observations/by_login.html.haml
seanclifford: Yeah, I was going for consistency between pages - but after seeing it like this now also prefer how it looks when viewing the page by itself. (Commit added)

kueda (approved):

PR #3256, opened 2021-09-30
bump up place admin_levels to make room for region

pleary:


app/assets/javascripts/atlases/show.js.erb
```diff
@@ -163,7 +163,7 @@ atlasJS.appendPlaceMetadata = function( panel, place ) {
     panel.append( addPlaceLink );
     panel.append( "<br>" );
   }
-  if( !existing || ( existing.type != "exploded"  && existing.place.admin_level < 2 ) ) {
+  if( !existing || ( existing.type != "exploded"  && existing.place.admin_level < 20 ) ) {
```
kueda, line 166: Since this is an ERB can we just refer to `Place::COUNTY_LEVEL`? Or are we trying to avoid model references in the JS templates?

app/models/observation.rb
```diff
@@ -2660,7 +2660,7 @@ def intersecting_places
   end
 
   # Actually referring to Place::STATE_LEVEL seems to cause trouble here
-  [1, 2].each do |admin_level|
+  [10, 20].each do |admin_level|
```
kueda, line 2663: I think the problem here was either a circular reference problem, or maybe loading the Place model triggered a lot of other stuff like loading preferences that then caused problems. Could we avoid that by moving these constance into a standalone module or class?

app/webpack/observations/identify/ducks/suggestions.js
```diff
@@ -122,7 +122,7 @@ export default function reducer(
         if ( observation.places ) {
           const place = _
             .sortBy( observation.places, p => p.bbox_area )
-            .find( p => p.admin_level !== null && p.admin_level < 3 );
+            .find( p => p.admin_level !== null && p.admin_level < 30 );
```
kueda, line 125: Would be nice to find a better place for these constants in JS, though clearly I didn't do that when I wrote this stuff originally. Maybe a `.js.erb` that gets included in the application bundle? Something to consider in the future, maybe.

db/migrate/20210930182050_modify_place_admin_levels.rb
```diff
@@ -0,0 +1,45 @@
...
+    batch_size = 1000
+    batch_count = ( max_id / batch_size )
+    for batch_index in ( 0..batch_count ) do
+      start_id = batch_index * batch_size
+      puts start_id
+      User.where( "id > ? AND id <= ?", start_id, start_id + batch_size )
+        .update_all( "lat_lon_acc_admin_level = lat_lon_acc_admin_level / 10" )
+    end
+  end
+end
```
kueda, line 45: This looks pretty chunky. Any estimate on how long it will take?

kueda (approved):
Looks good to me. Most of my comments are about trying to keep the constants in one place, but I don't think that needs to block merging.

app/assets/javascripts/atlases/show.js.erb
pleary: Yeah I think we probably could use the model constant here. I'll try that out

app/models/observation.rb
pleary: I didn't try any other options here because of the comment, and I wasn't sure exactly what issues it was causing. I can look into it more, but maybe after merging and deploying in this code so we can isolate this specific change

app/webpack/observations/identify/ducks/suggestions.js
pleary: We often define constants like SITE, CURRENT_USER, LIFE_TAXON, etc in views when rendering React components. So maybe we could extract those as well as other things we use more than once like these admin_level contstants into a JS erb file

db/migrate/20210930182050_modify_place_admin_levels.rb
pleary: The place one is pretty fast. Users is a bit longer, which is why I'm using a smaller batch size of 1000 so the update queries don't take too long. I don't have an exact time - I'm hoping to get that when running on staging. But I suspect it'll be on the order of a few minutes - less than 15 minutes total

kueda: Ah, so you're expecting to do this with some downtime?

pleary: I wasn't planning on doing this with downtime, but we can if you think that'll be preferred. There will be a short period of time when the admin_levels in the DB are updated, but the code the apps are using don't reflect the latest admin_levels since we'll migrate before restarting the apps. That may result in a few small issues for things like the dashboards of users w/ no activity, but I figured that downtime would be more disruptive. Now that I think of it, it's probably best to update the users admin_level first since that'll take a little longer and that change is least likely to result in problems, so I'll change the order in this migration

kueda: Hm, well I guess if things don't get locked up and Capistrano isn't going to hang during a 15 min deploy then this sounds ok to me.

PR #3291, opened 2021-10-28
Fix taxon photo term filter

### LINKED ISSUE #3250: Taxon Photo browser: not all photos will load if  taxon has no observations with "No Evidence fo Flowering" annotation but user is filtering for that
From https://forum.inaturalist.org/t/view-all-photos-still-not-working/26778 I can replicate. Might be a more general issue and not just related to `No Evidence of Flowering` annotation

From pisum on the forum:

1. open a taxon that has been annotated with “No Evidence of Flowering”. (i used Nyssa.) look for just photos from observations with No Evidence of Flowering.
2. now in the same window, search for Wikstroemia uva-ursi in the upper-right corner of the screen, and pull up that taxon.
3. look at the photos for that taxa (my screenshot below, with network monitor). note that although the screen shows that it is looking for phenology=any, it’s doing it while looking only among observations that have a phenology annotation (shown the parameters of the network request as &term_id=12). normally, you would expect that it would search any observation, regardless of whether it has an observation or not.

image

4. if you now pick any phenology annotation (ex. flowering), and then change it back to any, you’ll see that now you get everything back:

image

seanclifford:
Fix taxon photo term filter so both term_value_id AND term_id need to be relevant for the taxon for both of them to be included. 

Fixes #3250

Beforehand, if the term_id was valid, but the term_value_id not used, then only the term_id was included in the filter which would filter out all observations that didn't include some value for that term_id. This clashed with what the UI was showing as it indicatied that no filtering was occurring.

I also moved the setObservationParams function call to the bottom of the function (outside the `if ( taxon )` block) as it looked like a more correct location for it.

Note: I've reproduced and tested the fix locally, but as my local dev environment is lacking much data it might be worth putting it through its paces in a test environment.

### DIFF
--- a/app/webpack/taxa/photos/ducks/photos.js
+++ b/app/webpack/taxa/photos/ducks/photos.js
@@ -404,23 +404,19 @@ export function hydrateFromUrlParams( params ) {
     if ( params.quality_grade ) {
       newObservationParams.quality_grade = params.quality_grade;
     }
-    if ( taxon ) {
+    if ( taxon && params.term_id && params.term_value_id ) {
       const controlledAttrIDs = _.keys( taxon.fieldValues ).map( k => parseInt( k, 0 ) );
       const controlledValueIDs = _.flatten( _.values( taxon.fieldValues ) )
         .map( v => v.controlled_value.id );
-      _.forEach( params, ( value, key ) => {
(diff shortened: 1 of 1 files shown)


kueda:
Thanks!

Issue #2332, opened 2019-07-16
On collection project observations tab, show link to Explore at end of results


tiwane:
From https://forum.inaturalist.org/t/display-all-observations-on-project-pages/4951

When you click on the "[X number of] Observations" tab, you are only shown about 200 observations, and there is no link at the bottom of the page to either load another page, or to view all observations in explore. I think pagination is the preferred solution from the community, if possible.

kueda:
Ok, so what we want to do is when you've loaded all the results on this tab (200, I guess), show a link to `/observations?project_id=${project.slug}&verifiable=any&place_id=any` with the text "View more"

bianavarga:
Working on this.

PR #3303, opened 2021-11-06
Add View More links to bottom of Project observations pages

bianavarga:
Issue: https://github.com/inaturalist/inaturalist/issues/2332

Added View More link buttons to the bottoms of both the Grid and List views of Project observations.

Links will only be visible when there are more observations than are shown on the page and the infinite scroller has reached its limit.

Screen Shot 2021-11-05 at 11 15 11 PM
Screen Shot 2021-11-05 at 11 12 37 PM

### DIFF
--- a/app/webpack/projects/show/components/view_more_footer.jsx
+++ b/app/webpack/projects/show/components/view_more_footer.jsx
@@ -1,0 +1,35 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Grid, Row, Col } from "react-bootstrap";
+
+/**
+ * Renders a button-styled link with the text "View More"
+ * if the prop showViewMoreLink is true.
+ *
+ * Otherwise, renders nothing.
+ *
+ * Spacing around the link is appropriate for use as a footer.
+ *
+ */
+
+const ViewMoreFooter = ( { showViewMoreLink, viewMoreUrl } ) => {
+    return (showViewMoreLink
+        ?
+        <div className="ViewMoreFooter">
+            <Grid>
+                <Row>
+                    <Col xs={12} className="text-center">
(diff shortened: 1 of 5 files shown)


app/webpack/projects/show/components/view_more_footer.jsx
```diff
@@ -0,0 +1,35 @@
...
+ */
+
+const ViewMoreFooter = ( { showViewMoreLink, viewMoreUrl } ) => {
+    return (showViewMoreLink
+        ?
+        <div className="ViewMoreFooter">
+            <Grid>
+                <Row>
+                    <Col xs={12}>
+                        <a href={viewMoreUrl} className={"btn btn-primary btn-inat btn-green pull-left"}>{ I18n.t( "view_more" ) }</a>
```
kueda, line 22: Can you remove the `pull-left` here and add `className="text-center"` to the <Col> so the button is centered?

kueda (changes requested):
Other than the un-centered button I think this is looking good.

app/webpack/projects/show/components/view_more_footer.jsx
bianavarga: Done! Thx for the review.

bianavarga: Actually, I'm not sure that's sufficient from playing around with CSS on other buttons on the site. I'll figure it out & get back to you later this week.

kueda: Oops, forgot some backticks. Yeah, `className="text-center"` should be on a surrounding element.

bianavarga: Fixed for real & tested.

kueda (approved):

PR #3324, opened 2021-11-21
Replace react datetimefield component with newer fork

seanclifford:
Related issue #3276 

I've attempted to replace the [inaturalist fork](https://github.com/inaturalist/react-bootstrap-datetimepicker) of the old react-bootstrap-datetimepicker with an up-to-date fork - called [react-datetime](https://github.com/arqex/react-datetime).

This has some good locale support for the referenced issue, and enough flexibility to remove many of the work arounds need for the old component, such as on the uploader where the input field was hidden and a new one added below, just so the button can be put on the left instead of the right.

This is the first time I've made any substantial changes in react - so watch out for newbie mistakes, I may have made some!

Although I've got it working mostly nicely, the one thing that I haven't been able to get to work yet is pretty annoying - and means I don't recommend merging it yet as it is.
The problem is that when editing the text, the component tries to be too smart and replaces the text with what it thinks is valid given the format - this often results in deletions of text being undone, and makes editing text in the textbox very annoying.


kueda (changes requested):
This definitely fixes #3276, and there's a net reduction in lines of code, both great. Are you running eslint? I'm seeing a number of code style issues that `app/webpack/.eslintrc` should alert you to. If you don't have eslint integrated into your editor, you can run it manually like this:

```bash
./node_modules/.bin/eslint app/webpack/observations/uploader/components/date_time_wrapper.js
```

Probably not worth doing on all your changes, but certainly on that file where it's all new code.

Regarding the editing issue, that is annoying. For me it seems like the delete button is completely disabled, so I can't make any edits. Is it possible you could use the `renderInput` prop to render a custom input that has its own onChange listener registered, and stop event propagation if you detect a change due to a keyboard event?

@tiwane, what's your sense for how important manually editing the date text is in the Uploader? I almost never do it, personally, but I feel like if we were to ditch it we'd need to provide a mechanism to delete the date or just delete the time.

seanclifford:
Thanks for taking a look. I'd missed linting - something I should be doing for sure! I'll rectify that.

Yes a custom input/onChange might just solve it. I'd just need to make sure any edits sync back to the calendar.
I'll have another go at making it work - as I would personally be annoyed by the current behavior.

kueda:
@seanclifford this isn't ready for another review, correct? Looks like your last commit was just for eslint stuff. Just want to make sure I'm not leaving you hanging.

seanclifford:
Nope, not leaving me hanging - I've been a bit busy recently - I feel like I'm leaving this issue hanging!
I did get the datetime control working nicely in isolation a couple of weeks ago, but when I tried to implement that in iNat, it didn't work too well. Not sure why yet.
I do plan to come back to this... though let me know if the issue needs fixing sooner rather than later, I'm OK with dropping it if you'd like it seen to by someone else too.

seanclifford:
I took another look at this today, and think it's beyond what I can currently achieve in react. 
Here's the control in isolation, working reasonably nicely - that I couldn't get working in the application: https://codesandbox.io/s/react-datetime-value-disconnect-dnejw
I'll keep the branch around on my fork in case anyone wants to take another look at swapping out the control in future. 
I managed to resolve the issue on a new branch with minimal changes and without touching the control. I'll create another PR for that instead.

PR #3325, opened 2021-11-22
Truncate Projects by_login title and description with css

seanclifford:
Instead of truncating at the server.

This allows the full width to be used to display text.

No accepted issue #, but see this forum post for an example of the issue:
https://forum.inaturalist.org/t/projects-titles-are-cut-off-in-user-profile/20749

I used `-webkit-line-clamp` etc. for the 2 line truncation as it was being used elsewhere in the codebase.

### DIFF
--- a/app/assets/stylesheets/projects.css
+++ b/app/assets/stylesheets/projects.css
@@ -25,3 +25,15 @@
 	margin-bottom: 20px;
 	margin-top: 20px;	
 }
+.truncate {
+	white-space: nowrap;
+	overflow: hidden;
+	text-overflow: ellipsis;
+}
+.truncate-2 {
+	display: -webkit-box;
+	-webkit-line-clamp: 2;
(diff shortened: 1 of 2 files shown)


kueda:
Thanks again, @seanclifford, but in the future, please try and restrict yourself to the issues on Github. There are a lot of things that pop up in the Forum that we don't actually want to do, so I don't want folks spending too much time on stuff that we won't merge. I know sometimes things seem like obvious bugs that need fixing (like this case), but sometimes it's not so clear.

seanclifford:
I hear you @kueda. 
I did know I was sidestepping things which ran the risk of this not being merged. But as it only took me a short amount of time, and I found the issue myself before checking the forums, I thought I'd see if this PR would be accepted.
I don't plan to make a habit of it!
Thanks for the merge 😄

Issue #3098, opened 2021-06-24
Add comment field to new flag form


kueda:
- [ ] On the [new flag page](https://www.inaturalist.org/taxa/47655/flags/new), add a textarea labeled "Comment" that will create the flag and add a new comment by the user on that flag; indicate that the comment is to add supporting rationale for the proposed change like links
- [ ] Indicate that the "Reason" is required, since there's now an optional field in the form
- [ ] Remove the "Add Another Flag" link (why is that even there)

fredunit:
What is flag? is it flag variable?

kueda:
A [Flag](https://github.com/inaturalist/inaturalist/blob/main/lib/acts_as_flaggable/flag.rb) is a record indicating that something should change, usually that the site curators or sometimes site admins should do something about it. So if you think a taxon name is outdated and should change, you can add a Flag to that Taxon to indicate that a curator should update the taxon record somehow. The Flag model has a `flag` attribute that is a string describing what's wrong and what should change. However, the flag attribute is short, and sometimes curators need more information than people can provide there to actually enact the requested change, so flaggers are encouraged to leave a comment on the Flag with that additional information. The current flag form doesn't make that easy, though: you have to create your Flag record, then you get redirected back to the taxon page, then you have to go back the detail page of the flag you just made and add a comment. The proposed change here is to simplify this process by adding a comment field to the flag form so people can provide all the information a curator requires in a single form.

todtb:
Hey @fredunit, were you thinking of taking a look at this? 

I was just working on a few other changes to Flags, and was considering picking this up if you weren't already working it.

tiwane:
@todtb (or anyone) if you do have time to look at this, that would be awesome.

bianavarga:
I'll grab this 😄

bianavarga:
I think this will either a) require more refactoring of existing code than I feel comfortable doing without first consulting with someone b) require some nugget of rails knowledge that I have yet to acquire as I figure my way around.

The confusion stems from the fact that flags already have both a `comment` field, which seems to be used to provide details about how/why the flag was resolved, and a `comments` field, which holds the array of comments related to the flag.

However, I think this is causing some trouble related to assumptions that Rails makes about singular & plural words.
For instance, this HAML, which I believe should create a text area field for editing the body of the first comment:

```
= f.fields_for( @flag.comments.first ) do |f_comments|
[...]

PR #3329, opened 2021-11-24
Add comment field to new flag form

bianavarga:
Issue: https://github.com/inaturalist/inaturalist/issues/3098

Screen Shot 2021-11-29 at 12 06 03 AM

This is what happens when you try to submit the form with no reason.

Screen Shot 2021-11-29 at 12 06 59 AM

Note: My IDE did some trailing space deletion that I hope we can all live with.

### DIFF
--- a/app/controllers/flags_controller.rb
+++ b/app/controllers/flags_controller.rb
@@ -113,7 +113,7 @@ class FlagsController < ApplicationController
     @flags = @flags.paginate(per_page: 50, page: params[:page])
     render :global_index, layout: "bootstrap"
   end
-  
+
   def show
     @object = @flag.flagged_object
     @object = @object.becomes(Photo) if @object.is_a?(Photo)
@@ -123,7 +123,7 @@ class FlagsController < ApplicationController
       format.html { render layout: "bootstrap" }
     end
   end
-  
+
   def new
     @flag = Flag.new(params[:flag])
     @object = @model.find(params[@object_key])
@@ -135,7 +135,7 @@ class FlagsController < ApplicationController
       return
     end
   end
-  
+
   def create
     create_options = params[:flag]
     create_options[:user_id] = current_user.id
@@ -145,16 +145,26 @@ class FlagsController < ApplicationController
       redirect_to root_path
     end
 
-    if @flag = Flag.where(create_options).where(resolved: true).first
+    @create_comment = false
+    if @flag = Flag.where( create_options.except( "initial_comment_body" ) ).where( resolved: true ).first
       @flag.resolved = false
     else
       @flag = @object.flags.build(create_options)
+      @create_comment = create_options.key?( :initial_comment_body ) && create_options[:initial_comment_body].length > 0
+      if @create_comment
+        @comment_attributes = { parent: @flag, user: current_user, body: create_options[:initial_comment_body] }
+        @flag.comments.build( @comment_attributes )
+      end
     end
     if @flag.flag == "other" && !params[:flag_explanation].blank?
       @flag.flag = params[:flag_explanation]
     end
+
     if @flag.save
-      flash[:notice] = t(:flag_saved_thanks_html, url: url_for( @flag ) )
+      flash[:notice] = t( :flag_saved_thanks_html, url: url_for( @flag ) )
+      if @create_comment && Comment.where( @comment_attributes ).length == 0
+        flash[:notice] << " " + t( :unable_to_save_comment )
+      end
     else
       flash[:error] = t(:we_had_a_problem_flagging_that_item, :flag_error => @flag.errors.full_messages.to_sentence.downcase)
     end
@@ -184,7 +194,7 @@ class FlagsController < ApplicationController
 
 
   end
-  
+
   def update
     if resolver_id = params[:flag].delete("resolver_id")
       params[:flag]["resolver"] = User.find_by_id(resolver_id)
@@ -202,7 +212,7 @@ class FlagsController < ApplicationController
       if @object.is_a?(Project)
         Project.refresh_es_index
       end
-      format.html do 
+      format.html do
         flash[:notice] = msg
         redirect_back_or_default(@flag)
       end
@@ -214,9 +224,9 @@ class FlagsController < ApplicationController
         end
       end
     end
-    
+
   end
-  
+
   def destroy
     unless @flag.deletable_by?( current_user )
       msg = t(:you_dont_have_permission_to_do_that)
@@ -249,7 +259,7 @@ class FlagsController < ApplicationController
   end
 
   private
-  
+
   def load_flag
     render_404 unless @flag = Flag.where(id: params[:id] || params[:flag_id]).includes(:user, :resolver).first
   end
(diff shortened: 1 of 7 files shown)


config/locales/en.yml
```diff
@@ -173,7 +173,7 @@ en:
   x_taxa_curated_html:
     one: "<span class='count'>%{count}</span> taxon curated"
     other: "<span class='count'>%{count}</span> taxa curated"
-  x_taxa_matching_current_filters:
```
bianavarga, line 176: All changes in this file are whitespace removals except where marked.

config/locales/en.yml
```diff
@@ -1851,6 +1851,7 @@ en:
   # Label for the date/time when a flag was created
   flag_created: Flag Created
   flag_description: Flag this content to let the curators know something is wrong with it.
+  flag_comment_description: Add a comment to the flag to add supporting rationale for the change, such as links.
```
bianavarga, line 1854: My change! Does the community help translate these strings to every other language or should I do that with Google translate?

lib/acts_as_flaggable/flag.rb
```diff
@@ -34,11 +34,12 @@ def set_uuid
   }
 
   blockable_by lambda {|flag| flag.flaggable.try(:user_id) }, on: :create
-  
+
   # NOTE: Flags belong to a user
   belongs_to :user, inverse_of: :flags
   belongs_to :resolver, :class_name => 'User', :foreign_key => 'resolver_id'
   has_many :comments, :as => :parent, :dependent => :destroy
+  attr_accessor :initial_comment_body
```
bianavarga, line 42: This is the only code change in this file.

app/controllers/flags_controller.rb
```diff
@@ -113,7 +113,7 @@ def index
     @flags = @flags.paginate(per_page: 50, page: params[:page])
     render :global_index, layout: "bootstrap"
   end
-
```
bianavarga, line 116: Non-whitespace edits in this file between lines 148 and 159.

config/locales/en.yml
kueda: Yeah, translation is done by other volunteers at https://crowdin.com/project/inaturalistweb. Strings from the main branch get synced to Crowdin, so this new string will be available for translation when we merge this.

kueda: Let's change this to "Add an optional comment to this flag that further explains why you are adding it and provides links to supporting evidence."

app/views/flags/new.html.haml
```diff
@@ -8,10 +11,14 @@
   %p= t :flag_description
 = form_for( [@object,  @flag], html: { class: "clear readable buttonrow" } ) do |f|
   = f.label :reason, raw( t(:flag_reason, :record => @model.name.downcase) ), class: "inter"
-  = f.text_field :flag, class: "inter text", maxlength: 255, style: "max-width: 300px;"
+  = f.text_field :flag, required: true, class: "inter text flag-reason-input", maxlength: 255
   %label &rdquo;
+  .description
+    %p{ :class => "comment-description" }= t :flag_comment_description
+  = f.label raw( t(:comment) ), class: "inter"
```
kueda, line 18: I don't think you need the `raw()` here.

app/views/flags/new.html.haml
```diff
@@ -8,10 +11,14 @@
...
+  = f.text_area :initial_comment_body, class: "pull-left initial-comment-body", maxlength: 5000
```
kueda, line 19: Use the `Comment::MAX_LENGTH` constant instead of a hard-coded number here.

kueda (changes requested):
Does what it says on the tin! Just needs a few small changes and we should be good to go. FWIW, I think we should probably just scrap the flag form in favor of a new, bootstrapified version, but that's beyond the scope of the issue. I think the way you have it now is consistent with the ancient styling of that page, which is good enough for now.

config/locales/en.yml
```diff
@@ -1851,6 +1851,7 @@ en:
   # Label for the date/time when a flag was created
   flag_created: Flag Created
   flag_description: Flag this content to let the curators know something is wrong with it.
+  flag_comment_description: Add an optional comment to this flag that further explains why you are adding it and provides links to supporting evidence.
```
bianavarga, line 1854: New!

config/locales/en.yml
```diff
@@ -5069,6 +5070,7 @@ en:
   type_taxon_name: "Type taxon name"
   uicn_red_list_status: "IUCN Red List status:"
   umbrella_projects: Umbrella Projects
+  unable_to_save_comment_for_flag: Unfortunately, we were unable to save the comment.
```
bianavarga, line 5073: New!

config/locales/en.yml
```diff
@@ -5069,6 +5070,7 @@ en:
   type_taxon_name: "Type taxon name"
   uicn_red_list_status: "IUCN Red List status:"
   umbrella_projects: Umbrella Projects
+  unable_to_save_comment: Unfortunately, we were unable to save the comment.
```
bianavarga, line 5073: New!

app/controllers/flags_controller.rb
```diff
@@ -145,16 +145,26 @@ def create
...
+      end
     end
     if @flag.flag == "other" && !params[:flag_explanation].blank?
       @flag.flag = params[:flag_explanation]
     end
+
     if @flag.save
-      flash[:notice] = t(:flag_saved_thanks_html, url: url_for( @flag ) )
+      flash[:notice] = t( :flag_saved_thanks_html, url: url_for( @flag ) )
+      if @create_comment && Comment.where( @comment_attributes ).length == 0
```
pleary, line 165: Minor suggestion - this query works fine, but I wonder if it would be more readable and succinct to use the model you just created - i.e. `if @create_comment && !@flag.comments.any?`. Not necessary, just offering it as a possibility

pleary (changes requested):
It's all looking pretty good @bianavarga. I added one change request to make sure the saving behavior works as intended with the comment is invalid. If you commit that change I'll go ahead and approve this request. Thanks!

app/controllers/flags_controller.rb
```diff
@@ -145,16 +145,26 @@ def create
...
+      if @create_comment
+        @comment_attributes = { parent: @flag, user: current_user, body: create_options[:initial_comment_body] }
+        @flag.comments.build( @comment_attributes )
+      end
     end
     if @flag.flag == "other" && !params[:flag_explanation].blank?
       @flag.flag = params[:flag_explanation]
     end
+
     if @flag.save
```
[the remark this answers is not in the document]
bianavarga: Just curious, how did you test with an invalid comment?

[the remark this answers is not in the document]
pleary: I made the text extremely long to fail the length validation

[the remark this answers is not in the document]
bianavarga: Interesting -- I would have expected the text field to cut it off.

app/controllers/flags_controller.rb
bianavarga: Your suggested edit doesn't get the trick done, so I've left it as before.

pleary:
Looks good to me so I'll merge this in. Thanks!

PR #3332, opened 2021-12-06
Dwca changes

pleary:
* Create any extensions that use obs search within a single round of searching
* Cache taxonomy for faster ranked ancestor lookups
* More preloading


kueda:
Looks great to me, Patrick. Much more work than I thought! Also, I learned about `Hash#dig`.

Issue #3304, opened 2021-11-08
Add "observation created using" application field on Identify modal


abhasinat:
There was a request that @tiwane highlighted for identifiers to be able to see what application created the observation in the identify modal. Identifiers typically have different responses for observations made by new users on Android vs iOS vs web and can give better instructions to correct errors when they can figure out what device an observer is using.

![Identify - obs made indicator](https://user-images.githubusercontent.com/44380775/140807656-3abeffe8-82a7-4a33-b221-e6ae630a900e.png)

My design suggestion is to include the icon and application name beneath the Location. Additionally, I would recommend adding about 5-10px of extra space beneath the Application name field to create some distinction between the data fields and the View/Follow buttons - right now the View/Follow buttons are difficult to scan as distinct from the data fields.

kueda:
One minor design complication here is that in this situation the app icon should be square, but the app icon from the API might not be, so `background-size: contain` in a fixed size container might be preferable here, and we'll just live with some of the weirder icons.

erebuseternal:
I'll take a look at this :)

kueda:
@abhasinat, Marcel raises a good point in https://github.com/inaturalist/inaturalist/pull/3336: if we use the app icon, it's going to add distracting bit of color and inconsistency to that top section of the observation modal. Is there an icon (preferably a free fontawesome one) that you think would work there aside from the dark bullet in your mock? Or do you want to go with that dark bullet, or stick with a square, colorful version of the app icon? Keep in mind we don't have much control over that app icon, and even the ones we use for our own apps seem to have white backgrounds, which looks a bit odd over the gray of the obs modal. We can force the icons to be grayscale with `filter: grayscale(1)`, but we can't force them to use the exact shade of gray used by the other icons.

abhasinat:
In my ideal world, I would like the Seek app icon and the iNaturalist app icon to show as their images, and everything else to show as one standardized symbol. Our apps should get priority and should be treated differently. Everything else can be represented by one standard icon– perhaps this icon from the Octicons pack in Fontastic? If standardization is easiest, I'm comfortable applying this treatment to the "This observation was created using" section on Observation detail, and everywhere else on the site as well.

<img width="134" alt="Screen Shot 2021-12-17 at 3 06 05 PM" src="https://user-images.githubusercontent.com/44380775/146617863-848ad797-464f-4c67-9c7e-199b791d531f.png">

kueda:
[...]

PR #3336, opened 2021-12-09
Added "Created Using" to Observation Modal

erebuseternal:
Issue: https://github.com/inaturalist/inaturalist/issues/3304

I decided against using the icon within the modal as I felt it would ruin the styling pattern that comes from the fa icons and the name is definitely enough to identify the application. But I can change that if the icon is a must. 

Screen Shot 2021-12-09 at 9 19 58 AM

### DIFF
--- a/app/assets/stylesheets/observations/identify/observation_modal.scss
+++ b/app/assets/stylesheets/observations/identify/observation_modal.scss
@@ -550,14 +550,14 @@ body > [role="dialog"] .ObservationModal {
           text-align: center;
           display: block;
           line-height: 16px;
-          &.fa-globe {
+          &.fa-globe, &.fa-cloud-upload {
             top: 1px;
           }
         }
         img.app-thumbnail {
           position: absolute;
           left: 0;
-          top: 0;
+          top: 1px;
           width: 15px;
           text-align: center;
           display: block;
--- a/app/views/observations/identify.html.haml
+++ b/app/views/observations/identify.html.haml
@@ -3,7 +3,7 @@
   = I18n.t :identify_title
 - content_for :extrajs do
   :javascript
-    var DISPLAYABLE_APP_IDS = #{ [OauthApplication.inaturalist_android_app&.id, OauthApplication.inaturalist_iphone_app&.id, OauthApplication.seek_app&.id] };
+    var OFFICIAL_APP_IDS = #{ [OauthApplication.inaturalist_android_app&.id, OauthApplication.inaturalist_iphone_app&.id, OauthApplication.seek_app&.id].compact.to_json };
     var RECENT_OBSERVATION_FIELDS = #{ logged_in? ? current_user.recent_observation_fields.to_json(
       only: [:id, :name, :description, :datatype, :allowed_values, :values_count]).html_safe : "[]" };
     if ( typeof( CURRENT_USER ) !== undefined ) {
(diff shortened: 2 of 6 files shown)


kueda (changes requested):
This works, but there are a few minor code style issues. Do you have eslint running in your editor? If not, you can run it with

```bash
./node_modules/.bin/eslint app/webpack/observations/identify/components/observation_modal.jsx
```

You can ignore the errors that are outside of what you changed, but please fix the ones with your new code. Waiting to hear what Abhas thinks about the icon.

erebuseternal:
Linting done :)

erebuseternal:
Updated based on: https://github.com/inaturalist/inaturalist/issues/3304#issuecomment-999038943

### Case 1: Core App
Screen Shot 2021-12-22 at 11 23 24 PM

### Case 2: 
Screen Shot 2021-12-22 at 11 23 16 PM

app/views/observations/identify.html.haml
```diff
@@ -3,6 +3,7 @@
   = I18n.t :identify_title
 - content_for :extrajs do
   :javascript
+    var DISPLAYABLE_APP_IDS = #{ [OauthApplication.inaturalist_android_app&.id, OauthApplication.inaturalist_iphone_app&.id, OauthApplication.seek_app&.id] };
```
kueda, line 6: This can potentially make a JS array that looks like `[nil,nil,nil]` which is not valid JS. I think you should probably stick `.compact.to_json` onto the end.

app/assets/stylesheets/observations/identify/observation_modal.scss
```diff
@@ -554,6 +554,15 @@ body > [role="dialog"] .ObservationModal {
             top: 1px;
           }
         }
+        img.app-thumbnail {
+          position: absolute;
+          left: 0;
+          top: 0;
```
kueda, line 560: Can you add 1px here, and do the same for the `fa-cloud-upload` version too? That looks a bit more vertically aligned to me.

app/webpack/observations/identify/components/observation_modal.jsx
```diff
@@ -738,6 +738,24 @@ class ObservationModal extends React.Component {
                                 I18n.t( `places_name.${_.snakeCase( country.name )}`, { defaultValue: country.name } ) || I18n.t( "somewhere_on_earth" )
                               ) : I18n.t( "somewhere_on_earth" ) }
                             </li>
+                            { observation.application && observation.application.name && (
+                              <li>
+                                <a href={observation.application.url}>
+                                  { DISPLAYABLE_APP_IDS.includes( observation.application.id ) ? (
```
kueda, line 744: I think you should make this a prop (and name it something more descriptive like `officialAppIds`, since you're displaying something regardless of whether it's in this array) and set the default value to an empty array. You can still set a constant in the haml template like you're doing, but maybe store it in the `config` part of state in `webpack-entry.js` like we're doing with `PREFERRED_PLACE`, then pass it to this component as a prop from the container. The upshot of all that is that `ObservationModal` can be reused and not crash even when a developer fails to pass in this array of IDs, and a developer can supply the IDs in different ways in different contexts (maybe setting a global constant isn't possible in some situations).

kueda (changes requested):
Working well on the whole, just a few changes to do.

erebuseternal:
Not entirely sure if I'm setting the default value in the correct spot but otherwise I believe this addresses the last batch of comments.

kueda (approved):

PR #3340, opened 2021-12-18
first pass at email-suppressions

loarie:
one approach to making sure we don't send emails to people who have suppressed emails via sendgrid:

1) local EmailSuppressions table with columns for email and suppression_type to store remote sendgrid data locally
2) sync_suppressions tool to import data from SendGrid
3) when we check user.prefers_no_email, also check whether the user's email is represented in the EmailSuppressions table
4) allow database queries like this to export lists of people to email:

```
SELECT count(*)
FROM   users u
LEFT   JOIN preferences p ON p.owner_id = u.id AND p.owner_type = ‘User’ AND p.name = ‘no_email’
LEFT   JOIN email_suppressions s ON s.email = u.email AND s.suppression_type NOT IN (“news_from_inaturalist”, “transactional_emails”, “account_emails”)
WHERE  (p.value = ‘t’ OR p.id IS NULL)
AND (s.id IS NULL)
AND u.email IS NOT NULL
AND u.suspended_at IS NULL
AND u.site_id = 1
AND u.locale =  ‘en’;
```


loarie:
@kueda this is pretty bare bones but wanted to see if you think this is an ok approach before continuting

app/models/email_suppression.rb
```diff
@@ -0,0 +1,8 @@
+class EmailSuppression < ApplicationRecord
+
```
kueda, line 2: Can you please add a comment here saying that these are the names of Sendgrid suppression groups specified on Sendgrid? Otherwise it's kind of mysterious to someone who has no idea what's going on. Might also be worth noting that the IDs of the suppression groups that we actually use when sending emails are specified in `config/config.yml`.

app/models/user.rb
```diff
@@ -721,6 +721,17 @@ def get_lat_lon_from_ip
     self.lat_lon_acc_admin_level = lat_lon_acc_admin_level
   end
   
+  def suppressed_emails( suppressed_groups )
```
kueda, line 724: Put a question mark at the end of methods that return a binary. I would rename this `email_suppressed_in_group?` and allow it to take a single group name or an array.

db/migrate/20211216171216_create_email_suppressions.rb
```diff
@@ -0,0 +1,9 @@
+class CreateEmailSuppressions < ActiveRecord::Migration[5.2]
+  def change
+    create_table :email_suppressions do |t|
+      t.text :email
+      t.text :suppression_type
+      t.timestamps null: false
+    end
```
kueda, line 7: You should `add_index :email_suppressions, [:email, :suppression_type]` to support your most common query.

db/migrate/20211216171216_create_email_suppressions.rb
```diff
@@ -0,0 +1,9 @@
+class CreateEmailSuppressions < ActiveRecord::Migration[5.2]
+  def change
+    create_table :email_suppressions do |t|
+      t.text :email
+      t.text :suppression_type
+      t.timestamps null: false
```
kueda, line 6: I don't think you need `null: false` here.

tools/sync_suppressions.rb
```diff
@@ -0,0 +1,110 @@
...
+
+def get_supression_group
+  puts "getting group suppressions"
+  url_base = "https://api.sendgrid.com/v3/asm/suppressions"
+  records = get_records( url_base )
+
+  emails = Hash.new
+  records.map{ |a| a["group_name"] }.uniq.each do |group_name|
+  	group = records.select{ |a| a["group_name"] == group_name }.map{ |a| a["email"] }
+  	emails[group_name.downcase.gsub( " ", "_" )] = group
```
kueda, line 61: I think `group_name.parameterize.underscore` would do the trick here too.

app/models/email_suppression.rb
```diff
@@ -0,0 +1,8 @@
...
+  ACCOUNT_EMAILS = "account_emails"
+  DONATION_EMAILS = "donation_emails"
+  NEWS_EMAILS = "news_from_inaturalist"
+  TRANSACTIONAL_EMAILS = "transactional_emails"
+
```
kueda, line 7: You might consider adding a validation to ensure that `suppression_type` is one of the accepted values.

tools/sync_suppressions.rb
```diff
@@ -0,0 +1,110 @@
...
+  end
+  return emails
+end
+
+def save_emails( emails, suppression_type )
+  emails.each do |email|
+    unless es = EmailSuppression.where( email: email, suppression_type: suppression_type ).first
+      es = EmailSuppression.new( email: email, suppression_type: suppression_type )
+    end
+    es.save
```
kueda, line 71: `EmailSuppression.find_or_create_by( email: email, suppression_type: suppression_type )` would save you a few lines here. You might also want to log how many EmailSuppression records fail to save and output that at the end of your script.

tools/sync_suppressions.rb
```diff
@@ -0,0 +1,110 @@
...
+emails = get_supression_group
+emails.each do |suppression_type, emails|
+  save_emails( emails, suppression_type )
+end
+
+created_count = EmailSuppression.where("updated_at >= ? OR created_at >= ?", start, start).count
+
+to_destroy = EmailSuppression.where("updated_at < ? OR created_at < ?", start, start)
+destroy_count = to_destroy.count
+to_destroy.destroy_all
```
kueda, line 104: I think you could use `delete_all` here since there don't seem to be any `after_destroy` callbacks that need to run when removing an EmailSuppression instance.

kueda (changes requested):
Looks pretty good to me, just needs some small changes. Minor issues re: code style: please run `bundle exec rubocop -A tools/sync_sendgrid_suppressions.rb app/models/email_suppression.rb` to automatically get your new code to conform to the code style.

app/models/email_suppression.rb
```diff
@@ -0,0 +1,32 @@
...
+  DONATION_EMAILS = allowed_types[1]
+  NEWS_EMAILS = allowed_types[2]
+  TRANSACTIONAL_EMAILS = allowed_types[3]
+  BOUNCES = allowed_types[4]
+  BLOCKS = allowed_types[5]
+  INVALID_EMAILS = allowed_types[6]
+  SPAM_REPORTS = allowed_types[7]
+  UNSUBSCRIBES = allowed_types[8]
+
+  SUPRESSION_TYPES = allowed_types
```
kueda, line 31: IMO, it would be easier to maintain if you did something like

```ruby
SPAM_REPORTS = "spam_reports"
UNSUBSCRIBES = "unsubscribes"

SUPRESSION_TYPES = [
  SPAM_REPORTS,
  UNSUBSCRIBES
]
```

That way you're not going to worry about someone sticking something in the middle of the `allowed_types` array and messing with your indexes.

kueda (approved):
Nothing blocking, just that one thing that feels off.

PR #3342, opened 2021-12-28
Fix react-bootstrap-datetimepicker internationalization 

seanclifford:
Fixes issue #3276

Coming at this again after a failed attempt in #3324.

Fix was reasonably straight forward. Some react entry points didn't set the moment locale before the react-datetime control was rendered, so ended up rendering the days of the week in the wrong positions for the locale.

It might be possible to move this to some shared initialization? Not sure where though.

Fixes datetime controls in:
- project create
- uploader (just the left menu was broken)
- computer vision demo

Other areas that use the control already had moment locale initialized on webpack-entry:
- identify filters
- observation show (observation fields)

I also changed the react-bootstrap-datetimepicker git hash in `package.json` to point to the head of main rather than the head of a branch in case that branch is ever deleted. No functional change expected there.

### DIFF
--- a/app/webpack/computer_vision/webpack-entry.js
+++ b/app/webpack/computer_vision/webpack-entry.js
@@ -2,11 +2,14 @@ import "@babel/polyfill";
 import thunkMiddleware from "redux-thunk";
 import { createStore, compose, applyMiddleware, combineReducers } from "redux";
 import React from "react";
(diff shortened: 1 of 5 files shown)


kueda:
Looking good. Thanks again, @seanclifford!

PR #3347, opened 2022-01-05
first crack at app owner role

### LINKED ISSUE #3339: make creating an API client a privilege requiring 10 improving identifications a weeks for the last 4 weeks and and an account at least 2 months old
Currently, any account can create an API client https://www.inaturalist.org/oauth/applications
this makes it relatively easy for someone new to the iNaturalist to post large volumes of disruptive content
this change will hopefully skew use of the API clients towards established members of the community possibly less likely to cause disruptions via their use of the API client

loarie:
closes #3339 but uses an approach analogous to curators (e.g. app_owner role, email application to grant role, and conditions (account older than 2 months, at least 10 improving ids for 4 weeks...) to be able to submit application


loarie:
@kueda after talking with tiwane and carrieseltzer rather than treat https://github.com/inaturalist/inaturalist/issues/3339 using a `privilege` I have it hear using a `role` very analogous to how we deal with curators:

1) users without the role don't see a link to create an application, rather to apply for the app owner role
Screen Shot 2022-01-05 at 1 15 25 PM
If they don't have an account >2 months old and more than 10 IDs a week for the past 4 weeks, they can't submit an application:
Screen Shot 2022-01-05 at 1 16 54 PM
Alternatively, they can submit the application which sends an email to help:
Screen Shot 2022-01-05 at 1 08 50 PM
users/curate has tools to add that role:
Screen Shot 2022-01-05 at 1 16 15 PM
and remove that role:
Screen Shot 2022-01-05 at 1 16 21 PM
and users with that role can create applications biz as usual
Screen Shot 2022-01-05 at 1 15 45 PM

Any thoughts on this approach before going into the details? I still want to test the email sending part of this which I haven't done but its rubocop and spec compliant

config/locales/en.yml
```diff
@@ -635,13 +635,37 @@ en:
...
+    This content must be generated with real humans in the loop. Please read our
+    community guidelines prohibiting machine generated observations, identifications,
+    and comments for more details.
+  app_owner_application_about_p2_html:  |
+    In order to create third party apps, you must first submit an application to earn
+    the role of ‘app owner’. This application should describe the functionality your app
+    will have and how you plan to use it. In order to submit an application your account
+    must be at least 2 months old and must have made at least 10 improving identifications
+    a week for each of the last 4 consecutive weeks. We may ask followup questions about your
+    plan and we reserve the right to turn down applications for any reason.
```
kueda, line 652: Here's a re-write with American usage consistent with the rest of the default English text of the site (double quotes, terminal quote right of period, hyphenated adjectival phrases):

> In order to create third-party apps, you must first submit an application to become an "app owner." This application should describe the functionality your app will have and how you plan to use it. In order to submit an application your account must be at least 2 months old and must have made at least 10 improving identifications a week for each of the last 4 consecutive weeks. We may ask follow-up questions about your plan and we reserve the right to turn down applications for any reason.

app/views/oauth_applications/app_owner_application.html.haml
```diff
@@ -0,0 +1,21 @@
+- content_for :title do
+  =t :apply_for_permissions_to_create_inat_third_party_applications_title
```
kueda, line 2: This text needs to be added to en.yml, or just use `apply_to_create_third_party_applications_title`

app/controllers/oauth_applications_controller.rb
```diff
@@ -16,45 +24,96 @@ def create
...
+    elsif !@eligible
+      flash[:error] = t( :app_owner_application_inelligible_error )
+      @app_owner_application = params[:application] || {}
+      render :app_owner_application
+    elsif current_user.is_app_owner?
+      flash[:notice] = I18n.t( :user_already_app_owner )
+      redirect_to oauth_applications_url
+    else
+      Emailer.app_owner_application( current_user, params[:application] ).deliver_now
+      flash[:notice] = t( :app_owner_application_success )
```
kueda, line 74: This translation is missing.

app/controllers/oauth_applications_controller.rb
```diff
@@ -16,45 +24,96 @@ def create
...
+    if params[:application].to_s.size < MIN_EXPLANATION_LENGTH
+      flash[:error] = t( :app_owner_application_explanation_required )
+      @app_owner_application = params[:application] || {}
+      render :app_owner_application
+    elsif !@eligible
+      flash[:error] = t( :app_owner_application_inelligible_error )
+      @app_owner_application = params[:application] || {}
+      render :app_owner_application
+    elsif current_user.is_app_owner?
+      flash[:notice] = I18n.t( :user_already_app_owner )
```
kueda, line 70: This translation is missing.

app/views/users/curation.html.haml
```diff
@@ -164,6 +164,10 @@
             method: :post, data: { confirm: t(:are_you_sure_you_want_to_remove_spammer) },  |
             class: "btn btn-block btn-success"
         = link_to "Make taxon curator", new_taxon_curator_path( user_id: @display_user.id ), class: "btn btn-block btn-success"
+        - if @display_user.is_app_owner?
+          = link_to t(:remove_app_owner_status), remove_role_path(@display_user, :role => Role::APP_OWNER), :method => :delete, class: "btn btn-block btn-warning"
+        - else
+          = link_to t(:make_app_owner), add_role_path(@display_user, :role => Role::APP_OWNER), :method => :post, class: "btn btn-block btn-success"
```
kueda, line 170: I'm pretty sure we don't use this view any more. I think you want to update https://github.com/inaturalist/inaturalist/blob/main/app/views/admin/user_detail.html.haml

kueda (changes requested):
Couple thoughts:

1. As mentioned, I think the current application requirement is too harsh and would recommend easing it to 10 improving IDs in the last month if the intent is to restrict API client use active, helpful users and not just superusers. If we remove manual approval maybe it could be a bit more harsh.
2. Adding manual approval is a big downer for client authors (it would certainly prevent me from ever making an iNat client), and it creates more work for Tony. Is Tony ok with this?
3. If we have manual approval, what are the requirements? The applicant should understand what is and is not acceptable functionality before they apply, both for clarity and to save Tony time, otherwise Tony is going to have to exchange multiple emails trying to understand things like what will be posted, how frequently, how auth info will be stored, etc.
4. You are implicitly grandparenting in all existing clients, so you might want to address that somewhere, maybe on the application page, e.g. "As of 2022 we are requiring manual approval for users wishing to make API clients"
5. Check all your translated text; I noted a few instances where you didn't actually update en.yml, but you should review them all.
6. Try walking through the entire application process locally including applying, approving, and rejecting to make sure everything works.

kueda (changes requested):
Still missing some translation keys, and some text needs to be updated. Also, how does the applicant get notified that their application was approved?

PR #3361, opened 2022-01-16
Include scopes in JWT payload

nornagon:
Hi there! This is my first contribution to the iNat repository, so please let me know if there's anything I've missed.

The context for this change is that I'm working on an app ([iNaturalist Bingo](https://forum.inaturalist.org/t/inaturalist-bingo-a-monthly-observation-quest/29046)) that users can sign in to through iNaturalist. The app doesn't need write access to the user's iNaturalist account—it only needs the user's ID. At present, the only way to get a user ID on a cross-origin domain through the OAuth flow seems to be to request a JWT token and fetch `/v1/users/me` from the API server. This requires the _write_ scope, because of these two lines in users_controller.rb:

https://github.com/inaturalist/inaturalist/blob/fc8218f329abbb5462015e723a2756031d8834b1/app/controllers/users_controller.rb#L6-L7

i.e., an access_token with only the `login` scope cannot actually fetch any details about the logged-in user.

My proposal here is that we extend the JWT to include details about the granted scopes, which the API server can then read and apply.

Thoughts? Have I missed some way to implement this "Sign in with iNaturalist"-type flow more simply?

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -833,8 +833,13 @@ class UsersController < ApplicationController
     # not sure why current_user would be nil here, but sometimes it is
     return redirect_to login_path if !current_user
     payload = { user_id: current_user.id }
-    if doorkeeper_token && (a = doorkeeper_token.application)
-      payload[:oauth_application_id] = a.becomes( OauthApplication ).id
+    if doorkeeper_token
+      payload[:scopes] = doorkeeper_token.scopes
+      if a = doorkeeper_token.application
+        payload[:oauth_application_id] = a.becomes( OauthApplication ).id
+      end
+    else
+      payload[:scopes] = [:login, :write]
     end
     render json: { api_token: JsonWebToken.encode( payload ) }
   end


kueda:
>  At present, the only way to get a user ID on a cross-origin domain through the OAuth flow seems to be to request a JWT token and fetch /v1/users/me from the API server.

That's not strictly correct: you can use https://api.inaturalist.org/v1/search?q=nornagon&sources=users to find users (and user IDs) associated with a username, though there are going to be some edge cases where a username is a substring of many other usernames and our less-than-perfect search ranking won't return it in the first page of results.

Also, though undocumented and thus not really a part of the API, you *can* use https://api.inaturalist.org/v1/users/nornagon or https://www.inaturalist.org/users/nornagon.json to retrieve a user ID given a username.

On the other hand, if your client requires your user to sign in with iNat because you need to authenticate them (i.e. ensure that only the user with that username is using your client), you can request an access token with the `login` scope and then GET www.inaturalist.org/users/edit using that token and you'll get a response with the user ID.

To the larger issues of whether the concept of OAuth access token scopes should extend to our JSON Web Tokens, I guess I can imagine uses where that might be desirable or necessary, but since none are on the table at the moment I don't think that's something we want to build or support at the moment. Doing so would require us not only to write the bake the scopes into the JWT and duplicate some of Doorkeeper's OAuth scope recognition in iNaturalistAPI, but we'd also have to do the same in the Rails app, which will also accept JWT as a valid form of auth. Not a trivial amount of work, and it adds further complexity to our already-complex auth system that we then have to maintain in perpetuity.

Zooming back even further, please read https://github.com/inaturalist/inaturalist/blob/main/CONTRIBUTING.md. I realize this is a super technical issue that might seem weird for the Forum, but that's still the best place to propose new functionality. Maybe a Github issue would be ok for stuff like this, but in general we use Github to coordinate work that we on staff have decided to do. Proposals for other ideas should start in the Forum where they can be considered and discussed before spending time doing the the work.

nornagon:
Hi Ken-ichi! Thanks for the detailed response. I did file a feature request in the forum, here: https://forum.inaturalist.org/t/allow-api-authentication-without-write-access/29091

My goal with this change is to allow fetching the user's details—which I have none of to begin with, not even the username—through a "Sign in with iNaturalist" button. The desired experience is:

1. The user visits my app, & is unauthenticated.
2. The user clicks "Sign in with iNaturalist" or similar.
3. The user authenticates my app through the OAuth interstitial on inaturalist.org, and is then redirected back to my app.
4. My app can verify that the user is now logged in, & fetch basic user info (e.g. user name, user id).

I tried to `GET http://www.inaturalist.org/users/edit` but unfortunately this endpoint is not CORS-enabled.

kueda:
Hm, you're right about the CORS issue, but why do you need the user to authenticate at all? Is there some harm if I see your bingo page? If not you could just have people enter a username.

nornagon:
There's no particular harm right now, but I think it's a better experience to just click a button and be connected than to type in your username. Also, in future I can imagine wanting to rely more strongly on the authentication mechanism, e.g. to save preferences or coordinate a scoreboard. It seems handy to support!

CORS-enabling something like `/users/edit` might be a simpler way to go.

Issue #2589, opened 2020-02-21
Prevent curators from resolving flags on their non-taxa content


kueda:
If a curator leaves a nasty comment and that comment gets flagged, they should not be able to resolve that flag on their own comment. However, if you make a taxon and the taxon gets flagged, you should be able to resolve that flag since it's not about inappropriate behavior.

PR #3362, opened 2022-01-17
Only Taxon Flags can be Self Resolved

erebuseternal:
Resolves https://github.com/inaturalist/inaturalist/issues/2589

### DIFF
--- a/lib/acts_as_flaggable/flag.rb
+++ b/lib/acts_as_flaggable/flag.rb
@@ -201,7 +201,7 @@ class Flag < ActiveRecord::Base
 
   def resolvable_by?( user )
     if user.is_curator?
-      return true unless user.id === self.flaggable_user_id and flaggable_type != "Taxon"
+      return true unless (user.id === self.flaggable_user_id) && (flaggable_type != "Taxon")
     end
     false
   end
--- a/lib/acts_as_flaggable/flag.rb
+++ b/lib/acts_as_flaggable/flag.rb
@@ -199,4 +199,11 @@ class Flag < ActiveRecord::Base
     false
   end
 
+  def resolvable_by?( user )
+    if user.is_curator?
+      return true unless user.id === self.flaggable_user_id and flaggable_type != "Taxon"
+    end
+    false
+  end
+
 end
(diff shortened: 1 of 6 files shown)


lib/acts_as_flaggable/flag.rb
```diff
@@ -199,4 +199,11 @@ def deletable_by?( user )
     false
   end
 
+  def resolvable_by?( user )
+    if user.is_curator?
+      return true unless user.id === self.flaggable_user_id and flaggable_type != "Taxon"
```
kueda, line 204: Please use `&&` instead of `and`. I don't it doesn't make a difference here and in general we try to wrap assignments in parens which kind of skirts the difference, but it's helpful for consistency... and because every time I see `and` I have to spend a few seconds wondering whether it's going to cause an order of operations bug.

kueda (changes requested):
1. You've accomplished this in the view, but not in the model, leaving the possibility that someone who knows how to make HTTP requests outside of a browser could still resolve flags on their own content, so please address this at the model level as well with a validation.
2. Please test this in `flags_controller_api_spec.rb` by ensuring that a) a request to resolve a flag on your own content fails, and b) a request to resolve a flag on a taxon you created succeeds.

kueda:
You addressed the substance of my concerns, but I figured I'd note:

1. Stuff in spec/factories is for FactoryBot, which you use like `create :flag`, not `Flag.make!`. The latter is using machinist, which we're slowly transitioning away from as it isn't being maintained.
2. I asked you to address this in a model validation but you did it in the controller instead. That's fine as it addresses my concern about people making http requests outside of the browser, and I think it's reasonable to argue that permissions like that should be addressed in the controller and not the model, but FWIW, I usually err on the side of performing validations in the model so the problem is "solved" no matter how the model gets used. For example, right now the problem is solved for people using the FlagsController, but if we build some other way to resolve flags using a different controller, we'd have to re-implement this check.

erebuseternal:
Appreciate the feedback! As for (2) I figured that access to a flag is more about how you interact with a flag rather than what a flag is itself which is why I went for the controller rather than the model. But in the future I can stick with the model if that aligns better with the overall philosophy.

PR #3370, opened 2022-01-29
Exclude showing project user rules without users

### LINKED ISSUE #3318: Project pages go blank when you scroll down
https://forum.inaturalist.org/t/cant-leave-certain-projects/27941 Two project pages turn blank if you start scrolling down on them. Tried in both FF and Edge, with extensions on and off and could replicate. Here's a console shot from the user. Mine were the same.

image

Projects:

https://www.inaturalist.org/projects/discord-inaturalist-january-2020-weekendblitz
https://www.inaturalist.org/projects/discord-inaturalist-yearlisting-2020

seanclifford:
Fixes #3318

It's possible to have project user rules for deleted users.
When displaying project user rules, filter out the rules where the user is missing to avoid an error.

I thought about filtering this further up in `/app/webpack/projects/shared/models/project.js`, but as it was a shared, and the API can return rules without users by not passing `rule_details=true` in the request, I thought it best to fix it close to where the error was occurring.

### DIFF
--- a/app/webpack/projects/show/components/requirements.jsx
+++ b/app/webpack/projects/show/components/requirements.jsx
@@ -56,17 +56,19 @@ const Requirements = ( {
       </a>
     ) );
   let userRules;
-  if ( _.isEmpty( project.userRules ) ) {
+  const projectUserRules = _.filter( project.userRules, "user" );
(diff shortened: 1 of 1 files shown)


kueda:
Thanks again, @seanclifford!

PR #3377, opened 2022-02-10
Move photo metadata

pleary:
- Creates a new table `photo_metadata` for storing what's currently in `photos.metadata`
  - The main reason is to make the photos table bytewise smaller to speed up backup and recovery time
  - The metadata in the new table is stored compressed in a PostgreSQL bytea type column, ultimately taking about 45% of the space of current text column
  - Metadata gets YAML-dumped and Zlib-deflated before saving, and can be Zlib-inflated and YAML-loaded to fully restore the original data types, like `EXIFR::TIFF::Degrees`
- This PR also removes setting default values for the `native_*` fields on photos like `native_username` or `native_page_url`, in another effort to save space and prevent stale data
  - There are already methods to fallback to photo.user.login for example for `native_username`, so we shouldn't need to denormalize those values
- There is a method that can be used in delayed jobs to retroactively move metadata and strip `native_*` fields where applicable
  - Once all metadata has moved, there will need to be another commit to use the new `photo_metadata` table in place of `photos.metadata` so we can remove the `photos.metadata` column


kueda:
FWIW, this looks good to me. Out of curiosity, did you look into using a `jsonb` column instead of zipped yaml in a binary column? My guess is that it would take up less space than our current text column, but probably wouldn't save as much as zipping.

pleary:
I did try a `json` column, but not a `jsonb` column as it seems like json might end up needing less space, and we just store and retrieve metadata, never query it. I made 3 separate tables, one with a `text` column, one with a `json` column, and one with a `bytea` column. Normalized to the size of the text table the sizes were: text: 100%, json: 78%, and bytea: 48% with a test of about 5.3 million rows.

We also use get some pretty specific datatypes from EXIF parsing like `EXIFR::TIFF::Degrees` which get kind of mucked up when trying to convert to JSON and back. We can cast them as floats, but that'll change how we use them. I was also hitting encoding errors converting to JSON, which could be overcome, but but leaving the data as compressed YAML we can avoid those in this process.

I'm doing some timing of dump/restore right now, and it's actually not looking great for the compressed table. Saving disk space in the running DB might be costing time and space on dump/restore. If that hold up, I may want to yank the custom compressed serializer and stick with the regular YAML serializer but keep everything else

PR #3392, opened 2022-03-01
Updated patch for Rails, nokogiri, inaturalistjs

kueda:
Pretty sure this is safe, but just in case anyone wants to review.


pleary:
Looks good

PR #3398, opened 2022-03-04
Require email confirmation

kueda:
This is the first phase of email confirmation requirement in which new users must confirm their email, existing users can still sign in, but existing users can preemptively confirm their email addresses at users/edit. Later, we'll need to notify as many existing users as possible that they need to confirm their email address, then we will lock out anyone who doesn't.

The following API and nodejs PRs need to be merged before this gets merged:

* https://github.com/inaturalist/iNaturalistAPI/pull/303
* https://github.com/inaturalist/inaturalistjs/pull/58

The following work needs to be completed in the mobile clients before we can merge this:

* https://github.com/inaturalist/SeekReactNative/issues/418
* https://github.com/inaturalist/SeekReactNative/issues/417
* https://github.com/inaturalist/iNaturalistAndroid/issues/1188
* https://github.com/inaturalist/INaturalistIOS/issues/631


pleary (approved):
All looks good to me. I tested what I could locally and the main bits seemed to work well - dealing with legacy user confirmation, changing email confirmation, new user confirmation. The migration took a few minutes, which we knew would be the case. Not sure if the plan is to update production data a different way and mark the migration as having run, or do it during downtime, but it may causes issues to run that query as-is while users data is getting modified elsewhere.

kueda:
Many thanks. Regarding the migration, my plan was to run this prior to running the migration: 

```ruby
batch_size = 1000
num_batches = User.maximum( :id ) / batch_size
Parallel.each( num_batches.times, in_processes: 4, progress: "Updating users" ) do | i |
  start = i * batch_size
  stop  = i * batch_size + batch_size - 1
  User.connection.execute <<~SQL
    UPDATE users
    SET confirmation_token = NULL, confirmation_sent_at = NULL, confirmed_at = NULL
    WHERE
      (
        confirmation_token IS NOT NULL
        OR confirmation_sent_at IS NOT NULL
        OR confirmed_at IS NOT NULL
      )
      AND id BETWEEN #{start} AND #{stop}
  SQL
end
```
That took about an hour on staging and made the subsequent migration reasonably fast.

PR #3406, opened 2022-03-18
Small performance improvements

pleary:
A handful of smaller changes:

- Use "infinite scrolling" on species tabs of explore and projects pages. Results in a initial species_counts request that is smaller and a little faster
- Limit data fetched for top observer and IDer for taxon page
- Add indices to annotations and rules. Makes deleting annotations much faster; rules lookups are faster
- Increase query timeout on explore. Explore had a lower timeout than the API
- Cache results of API query from login page
- Change ListedTaxon query for observation bulk indexing to be more performant
- Added various preloading to reduce number of DB queries


kueda:
Works for me locally, not seeing any problems otherwise.

PR #3409, opened 2022-03-28
Makara

pleary:
- Adds makara gem which can be activated by updating database.yml
- Currently opt-in:
  - Delayed jobs are allow-listed in a plugin
  - Controller actions need to `prepend_around_action :enable_replica`
- Some overrides like assuming all DJ queries to need the primary, and allowing for context resetting in a non-Rack environment


app/controllers/application_controller.rb
```diff
@@ -291,6 +291,15 @@ def logstash_catchall
...
   end
 
+  def enable_replica
+    begin
+      ActiveRecord::Base.connection.enable_replica
+      yield
+    ensure
+      ActiveRecord::Base.connection.disable_replica
+    end
+  end
```
kueda, line 301: Just so I understand, is this turning on read/write splitting? I.e. if an action has this around it, Makara will perform SELECTs in the replica and INSERTs in the primary?

app/controllers/application_controller.rb
```diff
@@ -745,7 +754,9 @@ def user_viewed_updates_for( record, options = {} )
     return unless logged_in?
     updates = UpdateAction.where( resource: record )
     if options[:delay]
-      UpdateAction.delay( priority: USER_PRIORITY ).user_viewed_updates( updates, current_user.id )
+      ActiveRecord::Base.connection.without_sticking do
```
kueda, line 757: Again, just for my edification, this means that inside this block, if Makara directs a query to the primary, that does not mean that all subsequent queries get sent to the primary, which is the default "sticky" behavior, right?

config/initializers/makara.rb
```diff
@@ -0,0 +1,136 @@
...
+      SQL_SKIP_STICKINESS_MATCHERS = [] unless defined?( SQL_SKIP_STICKINESS_MATCHERS )
+      CUSTOM_SQL_SKIP_STICKINESS_MATCHERS  = SQL_SKIP_STICKINESS_MATCHERS + [/FROM "sessions"/i].map(&:freeze).freeze
+
+      def sql_primary_matchers
+        CUSTOM_SQL_PRIMARY_MATCHERS
+      end
+
+      def sql_skip_stickiness_matchers
+        CUSTOM_SQL_SKIP_STICKINESS_MATCHERS
+      end
```
kueda, line 49: This makes sure the module is defined, but what are these methods doing?

lib/elastic_model/acts_as_elastic_model.rb
```diff
@@ -259,7 +259,9 @@ def bulk_index(batch, options = { })
             refresh: options[:wait_for_index_refresh] ? "wait_for" : false
           })
           if batch && batch.length > 0 && batch.first.respond_to?(:last_indexed_at)
-            where(id: batch).update_all(last_indexed_at: Time.now)
+            ActiveRecord::Base.connection.without_sticking do
+              where(id: batch).update_all(last_indexed_at: Time.now)
+            end
```
kueda, line 264: Presumably the UPDATE gets run on the primary, but what are the subsequent queries that need to be "unstuck" and run on the replica?

kueda (commented):
I don't think you need my approval for anything here, I'm mostly reviewing so I can learn about how this works. On additional question: I take it chef will handle the changes Makara needs in database.yml?

app/controllers/application_controller.rb
pleary: Yes. This is a toggle I added for temporary use where we can opt-in a request to read/write splitting. If a controller/action does not call this with `prepend_around_action` then all its queries will run against the primary. If it does call this, it'll use read/write splitting that I plan to configure with "stickiness". Meaning all reads to to the replica until there's an INSERT or UPDATE or anything else that is known to need the primary. After a query goes to the primary (unless we temporarily ignore stickiness), all subsequent queries including SELECTS will go to the primary for the duration of the request, as well as for subsequent requests from the same user up until `primary_ttl` seconds (I plan to make that 5)

app/controllers/application_controller.rb
pleary: correct. There are some cases where we may need to write to the primary, but we know we won't ever read what was just written for the duration of the request, so there's no need to stick all subsequent queries to the primary

config/initializers/makara.rb
pleary: `sql_skip_stickiness_matchers` is defined in the Makara library, and I'm overriding it here. These are patterns where, if the current query matches any of the patterns, it'll automatically ignore stickiness. I wanted to extend the library defaults w/o repeating the default values, so I'm extending them with the CUSTOM_* constants, and overriding the methods to use the extended matchers. I can add some more comments in this file

lib/elastic_model/acts_as_elastic_model.rb
pleary: When we index an obs, we'll update it's last_indexed_at time. But we pretty much never need to turn around and query for that attribute in a time frame longer than expected replication lag. So this is making sure that for any request that indexes an observation, it's not stuck querying the primary for the duration of the request after doing the indexing, so we can spread around the post-indexing query load. Where we stick and don't will need some tweaking as we see how this works and how well load is split between the DBs

pleary:
Yes Chef will write database.yml, but for reference it'll look something like this:
```
production:
  username: USERNAME
  password: PASSWORD
  encoding: utf8
  database: DB_NAME
  sslmode: require
  adapter: makara_postgis
  makara:
    primary_ttl: 5
    sticky: true
    connections:
      - role: primary
        name: primary
        host: PRIMARY_HOST
        disable_blacklist: true
      - role: replica
        name: replica
        host: REPLICA_HOST_1
      ...
```

PR #3441, opened 2022-05-10
Clashes

loarie:
Added clash analysis to taxa/:id/edit when ancestry changes would narrow a parent or ancestors and on  taxon_changes/show when committing a swap would do the same.

I plan to post the following tutorial to the forum when this is deployed
https://docs.google.com/document/d/1AljPgP-ubihjzL11xkhkFy-NrfqIUwJCOx51ltArGfo/edit?usp=sharing

@kueda all changes to the Gemfile/Gemfile.lock here are to accommodate the M1 situation - not otherwise required by this branch.


Gemfile
```diff
@@ -18,7 +18,6 @@ gem "audited"
 gem "aws-sdk-cloudfront"
 gem "aws-sdk-s3"
 gem "aws-sdk-waf"
-gem "biodiversity"
```
kueda, line 21: This just kicks the can down the road. Either refactor https://github.com/inaturalist/inaturalist/blob/main/lib/ratatosk/lib/ratatosk/name_providers/eol.rb#L215 or remove the EOL name provider since I don't think we use it any more and don't have plans to bring it back. My preference is for the latter.

Gemfile
```diff
@@ -2,7 +2,7 @@
 
 source "https://rubygems.org"
 
-ruby "~> 2.6.0"
+ruby "> 2.6.0"
```
kueda, line 5: @pleary is this going to cause any problems for us in production?

Gemfile.lock
```diff
@@ -820,7 +818,7 @@ DEPENDENCIES
   yui-compressor
 
 RUBY VERSION
-   ruby 2.6.5p114
+   ruby 2.7.6p219
```
kueda, line 821: @pleary same here. It deployed to staging, interestingly. I was sort of expecting it to barf b/c it didn't know about that Ruby.

app/assets/javascripts/taxon_changes/show.js
```diff
@@ -60,4 +60,32 @@ $( document ).ready( function () {
     $( ".spinner" ).hide();
   } );
 
+  var potential_clashes_link = $( ".potential_clashes_link" )
+  var old_ancestors = potential_clashes_link.data( "input-ancestry" )
+  if( $.isNumeric( old_ancestors ) ){
+    old_ancestors = [ String( old_ancestors ) ];
+  }else{
```
kueda, line 67: I guess we don't have instructions for setting up eslint, but for my sanity please put spaces around `else`.

app/assets/stylesheets/taxa/form.css
```diff
@@ -36,3 +36,35 @@ fieldset.column {margin-bottom: 1em;}
...
+  color: #333;
+  border: 1px solid black;
+  padding: 20px;
+}
+
+.jqmOverlay { background-color: #000; }
+
+* html .jqmWindow {
+  position: absolute;
+  top: expression( ( document.documentElement.scrollTop || document.body.scrollTop ) + Math.round( 17 * ( document.documentElement.offsetHeight || document.body.clientHeight ) / 100 ) + 'px' );
```
kueda, line 57: Don't use expression() in CSS. This is an Internet Explorer-specific feature that apparently some other browsers picked up and then IE abandoned. Just write Javascript in the .js files.

app/assets/stylesheets/taxa/form.css
```diff
@@ -36,3 +36,35 @@ fieldset.column {margin-bottom: 1em;}
...
+  margin-bottom: 10px;
+  display:  none;
+}
+
+td.fig20 {
+  width: 20%
+}
+td.fig10 {
+  width: 10%
+}
```
kueda, line 70: Newline at the end of all files, please.

app/models/taxon.rb
```diff
@@ -2158,6 +2158,92 @@ def flagged_with( _flag, _options = {} )
     elastic_index!
   end
 
+  def clash_analysis( new_parent_id )
+    if INatAPIService.get( "/identifications", { current: true, exact_taxon_id: id, per_page: 0 } ).total_results.zero?
+      []
```
kueda, line 2163: Is this supposed to be `return []`?

config/locales/en.yml
```diff
@@ -596,6 +596,7 @@ en:
   alphabetical_order: Alphabetical Order
   already_have_an_account_log_in_html: Already have an account? <strong>Log in.</strong>
   amphibians: amphibians
+  analyze_clashes: Analyze Clashes
```
kueda, line 599: This and every other sentence fragment needs a comment explaining what it is. What's a clash? What do you mean by analyze?

config/locales/en.yml
```diff
@@ -5051,6 +5064,7 @@ en:
     others are just down the street! %{site_name} Places let you browse life
     from around the planet.  Start exploring!
   then_keybord_sequence: then
+  there_are_no_current_ids: There are no current IDs of any narrowed parents.
```
kueda, line 5067: Again, I don't know what a narrowed parent is, so you're going to have to explain it in a comment.

kueda (changes requested):
Just a review based on a preliminary read, have not tried running the code.

app/models/taxon.rb
loarie: not according to rubocop

kueda: I don't know what that means. As written this `if` statement does nothing, so why is it here?

loarie: I guess you're right, and rubocop no longer complaining with that return. Thanks

Gemfile
loarie: I removed the EOL name provider

app/assets/javascripts/taxon_changes/show.js
loarie: fixed

app/assets/stylesheets/taxa/form.css
loarie: I just removed it - are we developing for IE anymore?

app/assets/stylesheets/taxa/form.css
loarie: done

config/locales/en.yml
loarie: added comments

config/locales/en.yml
loarie: added comments

app/assets/stylesheets/taxa/form.css
kueda: I think it's worth supporting Edge, but not IE. It's certainly not worth using functionality that was originally made to only work in IE.

config/locales/en.yml
kueda: I still don't understand what this means. If you move genus Homo to family Canidae, then is family Homonidae "narrowed" because it has fewer descendants? Is there no better word for this, or a word that we already use? I feel like you're adding several new pieces of jargon that make our complicated taxonomic situation even harder to understand, and that's just in English.

loarie: yes - if you were to move Homo to Canidae you would be narrowing Homonidae. Existing IDs of Homonidae that were made thinking that the branch includes Homo would be reinterpreted by iNat as disagreements with IDs of Homo. These are clashes. If there are >10 clashes you should split the narrowed taxa (e.g. Homonidae). If you want to propose with different semantics or policy happy to consider

config/locales/en.yml
kueda: This is still pretty confusing to me. If you can't explain it succinctly, I think you need to link to your tutorial. I guarantee you this is going confuse translators, which means more time spent answering questions. I also suggest using a more specific term for a "clash" like "taxon change clash" or "identification clash" and then use that term consistently.

config/locales/en.yml
kueda: I don't think you need new meanings (i.e. semantics), I think you need more descriptive terms, but I don't have better suggestions. Personally I prefer to stick to the tree metaphor (e.g. prune, graft), so I might go with "thin" instead of "narrow" as in thinning the branches. I wish there were more tree-like terms for ancestor / descendant and parent / child.

I do think this is so complicated that you should probably link to your documentation in these comments.

kueda (changes requested):
* Your images can't be translated, so please remove the text from them
* I'd still like @pleary to sign off on the Gemfile / ruby version changes before we merge and deploy

Gemfile
pleary: I'll test it out on staging too. Ruby versions are fixed on our servers and to change them we'll need reinstall Passenger and recompile nginx. We also specify 2.6 in the CI.yml, so that's what CI will use. I think ideally we'd keep `~> 2.6.0` so developers don't use features only available or changed in 2.7 which don't work in CI or on our servers. I'm happy to make a push to upgrade to 2.7 in production sooner since it seems like that's a requirement for the ARM processors on M1 Macs. So I'll test out if this will break this one deploy, and it sounds like it won't, but I think I'd prefer moving this to something like an `M1-requirements` branch or something until we update production to 2.7, and make it `ruby "~> 2.7.0"` so we keep a fixed major version of Ruby

loarie:
@kueda I reverted the ruby version, removed text from the images, and changed the semantics from 'clashes' to 'unintended disagreements' and from 'narrowed' to 'thinned'

PR #3466, opened 2022-06-16
Dev setup changes

kueda:
* Site.default creates a new site if one doesn't exist
* welcome/index should function without a site
* System calls in bin/setup raise exceptions
* Hardcoded default postgres username and password in example database and
  Docker config
* Minor updates to some data loading scripts


pleary (approved):
This all looks good to me

PR #3473, opened 2022-06-24
Update file type error messages

### LINKED ISSUE #3472: Inconsistent/wrong file type error messages in web uploader
From https://forum.inaturalist.org/t/inconsistent-wrong-file-type-error-messages-in-web-uploader/21869

In the web uploader we don't always consistently or correctly show the proper error message if a file is not accepted. Large unsupported images will correctly show the file type error, small unsupported images will create ! cards, and unknown MIME types won’t give details of their error.

jwidness:
Resolves #3472


kueda (approved):
This looks good to me. Tested with a number of image and sound files and I think the problems in the issue have been addressed.

PR #3474, opened 2022-06-24
Commit job notice

loarie:
added simple notice to taxon_changes show when a commit_records delayed job is still working to alert users

Also fixed bugs with analyze_ids and analyze_unintended_disagreement tools


kueda (approved):
I just made some cosmetic changes. Should be good to merge when tests pass.

PR #3490, opened 2022-08-01
Remove gulp

pleary:
Removes the `gulp` package which is no longer actively maintained and requires a bunch of package versions with security warnings. Webpack can be triggered now with `npm run webpack` for a one-time run, and `npm run webpack-watch` to keep webpack listening for changes much like the `gulp` command was doing previously


kueda:
I guess it can't be part of this PR, but could you also update the parts of https://github.com/inaturalist/inaturalist/wiki/Development-Setup-Guide that refer to gulp, please?

pleary:
Since specs are passing, I'll go ahead and merge this in then I'll update the dev setup guide. I made a similar change to inaturalistjs https://github.com/inaturalist/inaturalistjs/pull/63 and I'll merge that one in as well

PR #3503, opened 2022-08-30
Adding circle/rectangle drawing btns to map

benkeen:
Hi @kueda, here's a first pass at adding a couple of buttons to the iNat map for drawing arbitrary rectangles and circles. Once I got into it doing the actual implementation I found there were a tremendous number of little UX scenarios that needed to be caught, including some oddities of the DrawingManager google library. So I ended up scaling back my original intentions somewhat.

Still, give it a look over when you get time. Absolutely anything can be changed. And I won't claim my Angular is the most beautiful thing in the world... :)

Also, disclaimer: I wasn't able to get my local dev env fully working (tiles, especially) so I tested all this with a pretty basic local setup. 

Screen Shot 2022-08-29 at 9 02 30 PM


app/views/observations/index.html.haml
```diff
@@ -147,9 +146,33 @@
                     = "{{ p.placeTypeLabel( ) }}"
                 %span{ "ng-show": "!nearbyPlaces.community || nearbyPlaces.community.length == 0" }
                   {{ shared.t("no_places_available") }}
-            %button.button.btn-default.reload{ "ng-click": "filterByBounds( )", "ng-show": "viewing('observations', 'map') && !hideRedoSearch" }
-              %span.glyphicon.glyphicon-repeat{ "aria-hidden": "true" }
-              {{ shared.t('redo_search_in_map') }}
+            %span.map-shape-searches
+              %div.shape-buttons
```
benkeen, line 150: I grouped the buttons in a div so they'd wrap together on small screens. But perhaps expanding it to include the Fit to Map button?

config/locales/en.yml
```diff
@@ -899,7 +899,7 @@ en:
   bold_label_colon_value_html: "<strong>%{label}:</strong> %{value}"
   # Text inserted into a textarea when the bold button is clicked but no text was selected
   bold_text: bold text
-  bounding_box: Bounding Box
+  bounding_box: Clear boundary
```
kueda, line 902: 1. It's almost always better to change the translation key as well as the value, for the non-obvious reason that the way our translations are set up right now, translators will be notified about new keys, but not about changes to values of existing keys. Beyond technicalities, I think it just makes things easier to reason about when the keys match their values, e.g. when you're wondering what key defines the text "Clear boundary" you're probably going to look for `clear_boundary` first.
2. I'd like to stick with a noun for this text so that the part of speech is closer to the proper nouns of place names, so here that would be "Custom Boundary" I think rather than "Clear Boundary." I get that a verb makes more sense for a button, but I think this would be the least intrusive approach that's most consistent with the existing design.

app/views/observations/index.html.haml
```diff
@@ -147,9 +146,33 @@
...
+                  %img{ src: asset_path( "draw-circle-icon" ), "aria-hidden": "true" }
+              %button{
+                "ng-click": "filterByBounds( )",
+                "ng-show": "viewing('observations', 'map')",
+                :class => "btn btn-default btn-fit-to-map {{ disableFitToMap ? 'is-active' : '' }}",
+                "tooltip-enable" => "!disableFitToMap",
+                "uib-tooltip" => "{{ shared.t('search_current_map_range') }}",
+                "tooltip-class": "search-current-map"
+              }
+                {{ shared.t("fit_to_map") }}
```
kueda, line 175: I'd like to stick with the current "Redo search in map". I know it's wide, but I don't think the meaning of "Fit to map" will be immediately obvious to most people at first.

app/views/observations/index.html.haml
```diff
@@ -147,9 +146,33 @@
...
+                  "tooltip-enable" => "currentActiveDrawing == null",
+                  "uib-tooltip" => "{{ shared.t('draw_circular_range') }}"
+                }
+                  %img{ src: asset_path( "draw-circle-icon" ), "aria-hidden": "true" }
+              %button{
+                "ng-click": "filterByBounds( )",
+                "ng-show": "viewing('observations', 'map')",
+                :class => "btn btn-default btn-fit-to-map {{ disableFitToMap ? 'is-active' : '' }}",
+                "tooltip-enable" => "!disableFitToMap",
+                "uib-tooltip" => "{{ shared.t('search_current_map_range') }}",
```
kueda, line 172: Let's stick with the word "boundary" instead of "range" for consistency.

app/views/observations/index.html.haml
```diff
@@ -147,9 +146,33 @@
...
+                  "tooltip-enable" => "currentActiveDrawing == null",
+                  "uib-tooltip" => "{{ shared.t('draw_rectangular_range') }}"
+                }
+                  %img{ src: asset_path( "draw-rectangle-icon.png" ), "aria-hidden": "true" }
+                %button{
+                  "ng-click": "drawRegion( 'circle' )",
+                  "ng-show": "viewing('observations', 'map')",
+                  :class => "btn btn-default btn-circle {{ currentActiveDrawing != null ? 'is-active' : '' }}",
+                  "tooltip-enable" => "currentActiveDrawing == null",
+                  "uib-tooltip" => "{{ shared.t('draw_circular_range') }}"
```
kueda, line 164: Let's stick with the word "boundary" instead of "range" for consistency.

app/views/observations/index.html.haml
```diff
@@ -147,9 +146,33 @@
...
-              %span.glyphicon.glyphicon-repeat{ "aria-hidden": "true" }
-              {{ shared.t('redo_search_in_map') }}
+            %span.map-shape-searches
+              %div.shape-buttons
+                %button{
+                  "ng-click": "drawRegion( 'rectangle' )",
+                  "ng-show": "viewing('observations', 'map')",
+                  :class => "btn btn-default btn-rect {{ currentActiveDrawing != null ? 'is-active' : '' }}",
+                  "tooltip-enable" => "currentActiveDrawing == null",
+                  "uib-tooltip" => "{{ shared.t('draw_rectangular_range') }}"
```
kueda, line 156: Let's stick with the word "boundary" instead of "range" for consistency.

app/assets/javascripts/ang/controllers/observation_search.js
```diff
@@ -1196,33 +1271,36 @@ function( ObservationsFactory, PlacesFactory, TaxaFactory, shared, $scope, $root
   $scope.preInitialize( );
 }]);
 
+var boundaryBoxStyle = {
```
kueda, line 1274: Maybe add a comment here explaining that Google requires this style info be added in code, assuming that's true.

kueda (changes requested):
This is awesome, I think people (including me) are going to be psyched about using it. Thanks for all this work, and for all the minor bug fixes and formatting fixes you added too. Good work on all the careful edge case handling.

Here's some feedback that didn't seem to work for inline comments:

Blocking bug:

- [x] If you search for a place (e.g. https://www.inaturalist.org/observations?place_id=14) then draw a shape on the map, the boundary of the place is not removed and cannot be removed without reloading the page.

Thoughts:

* I want there to be a way to clear a shape that's on the map. In the current setup where you don't interact with the map to change the search query it kind of makes sense that the map is not the place to clear a spatial filter, but introducing the ability to add that filter on the map makes me feel like I should be able to remove it on the map as well. That introduces some UI complexities with clearing place boundaries, though, so that might be out of scope here.
* When I redo search in map area and try to pan the map, I end up dragging the rectangle instead. Not sure there's a way around that, but it might help to zoom out a level after redoing the search in map area so the user can see that the bounding box is draggable. Maybe hiding that button when fully zoomed out would assist in that and remove one of your edge cases, too.

benkeen:
Thanks @kueda, great feedback! 

> I want there to be a way to clear a shape that's on the map. In the current setup where you don't interact with the map to change the search query it kind of makes sense that the map is not the place to clear a spatial filter, but introducing the ability to add that filter on the map makes me feel like I should be able to remove it on the map as well. That introduces some UI complexities with clearing place boundaries, though, so that might be out of scope here.

Yes, agreed. Having drawn on the map your focus is on that area and seeing an "x" icon to remove what you just drew feels intuitive. The "Clear Boundary" button is a long way off. Once you learn that's where you go to clear what you just drew, that's fine, but it's a bit of a leap if you haven't used it before. My concern with adding a clear icon was mainly _where does it go?_ Dead centre felt a bit ugly visually, but it would work for both shapes. An "x" would also obscure the contents of the map a little. Not a big deal but it still kinda rankled somehow. But let me take a look and see if I can find something that works.

> it might help to zoom out a level after redoing the search in map area so the user can see that the bounding box is draggable. 

Great idea. Let me try that out. 

> Maybe hiding that button when fully zoomed out would assist in that and remove one of your edge cases, too.

Agreed.

benkeen:
Actually one other question: I hunted around but couldn't find good icons for the rectangle + circle drawing icons, so ended up created them using an open source image editor, Inkscape. It might be worthwhile including the original source files in case down the road they need to be edited. Where would be a good location to add them? I can add them to app/assets/images but I wasn't 100% sure if they'd then be included with the other "live" assets when deploying.

kueda:
Regarding the icons, I think it's fine to leave them as PNGs, we can always make new ones if we need to. If you want to go above and beyond though, you could redo them as simple SVGs and style them with CSS. We use SVGs across the site so as long as you're not doing anything too fancy they should work just fine.

config/locales/en.yml
```diff
@@ -899,7 +899,6 @@ en:
   bold_label_colon_value_html: "<strong>%{label}:</strong> %{value}"
   # Text inserted into a textarea when the bold button is clicked but no text was selected
   bold_text: bold text
-  bounding_box: Bounding Box
```
benkeen, line 902: Let me know if I should remove this from all locale files (or if there's an automated way to do it).

benkeen:
Hey! Quick update: I've implemented everything except the option to remove the drawing right from the map. Give me a day or two to play around with that.

benkeen:
Hey @kueda, I didn't have much success with adding a close icon on the map I'm afraid. But try out the latest commit for an example. I only added it for the draw rectangle button - and the new "x" icon on the map is the only one that properly clears the drawn shape.

Position-wise, I found the new close icon had to be placed _outside_ the drawn shape, since when zooming out the shape could become very small so wouldn't have room for an icon if it were inside. But the big problem was that I couldn't find a way to add that icon _relative to the drawn shape_, just the map itself, so all movements had to be handled separately. e.g. try dragging the rectangle around the map - the "x" does reposition relative to the shape as you drag it, but it's ever so slightly off at times, presumably due to off-by-ones in the math. But the worst part was zooming in & out... that looks really jumpy and I couldn't find a way to smooth it out. I experimented with hiding/showing the icon during map interactions, but that felt pretty klunky. 

So yeah, not a triumph. Perhaps might be best to just omit this option for now?

config/locales/en.yml
kueda: en.yml is the only source file, so the next time we import new translations from Crowdin it will automatically get removed from all the other locale files.

app/views/observations/dwc.eml.erb
```diff
@@ -118,7 +118,7 @@
     <coverage>
       <% if @extent %>
       <geographicCoverage>
-        <geographicDescription>Bounding Box</geographicDescription>
+        <geographicDescription>Custom Boundary</geographicDescription>
```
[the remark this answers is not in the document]
kueda: @benkeen can you explain why you made this change?

kueda:
How about something like this?

```javascript
$scope.positionClearDrawingIcon = function ( shape, options ) {
  options = options || {};
  if ( !$scope.clearDrawingIcon ) {
    return;
  }

  var ne = shape.getBounds().getNorthEast();
  var sw = shape.getBounds().getSouthWest();

  var topLeft = {
    lat: ne.lat(),
    lng: sw.lng()
  };
  const point = latLng2Point(topLeft, $scope.map);
  const pos = point2LatLng({ x: point.x - 20, y: point.y - 20 }, $scope.map);

  if ( options.reappear ) {
    $scope.clearDrawingIcon.setVisible( false );
    $scope.clearDrawingIcon.setPosition( pos );
    setTimeout( function ( ) {
      $scope.clearDrawingIcon.setVisible( true );
    }, 500 );
  } else {
    $scope.clearDrawingIcon.setPosition( pos );
  }
};
```

That way when you're repositioning while panning you can maintain your former approach, which looks nice and smooth, but while zooming you can pass the `reappear` option and the button just disappears and comes back, so there's no jerk as the zoom animation is happening.

benkeen:
Yeah that's pretty good - the timeout really improves things actually. Thanks very much!

app/views/observations/dwc.eml.erb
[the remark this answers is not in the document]
benkeen: I think I was just conflating two things. I was avoiding the "bounding box" language since a circle is no longer a box, but this appears to be a different creature.

benkeen:
Hey @kueda, sorry for the wait. Try out the latest!

app/views/observations/dwc.eml.erb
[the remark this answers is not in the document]
kueda: Roger. Could you change this back to "Bounding Box" then?

config/locales/en.yml
```diff
@@ -1517,6 +1516,7 @@ en:
     private and obscured contributions to this project.
   curator_identification: Curator identification
   current_flags: Current Flags
+  custom_boundary: Custom boundary
```
kueda, line 1519: Could you make this "Custom Boundary"? Most text here is going to be in title-case.

kueda (commented):
This is looking really good. I got it on our staging server for the team to review so I might come back with some more feedback.

One thing that's bugging me is when I make a search boundary and then zoom in, I can no longer pan the map because it just drags the bounding box instead of panning the map. Not sure what the best solution to that is, though. Did you think about that at all? Like, you could disable drag when a bounding box is outside the viewport, or you could make some kind of dragging "handle" element similar to the button that removes the boundary, or maybe just the borders of the boundary are draggable.

benkeen:
> One thing that's bugging me is when I make a search boundary and then zoom in, I can no longer pan the map because it just drags the bounding box instead of panning the map. Not sure what the best solution to that is, though. Did you think about that at all? Like, you could disable drag when a bounding box is outside the viewport, or you could make some kind of dragging "handle" element similar to the button that removes the boundary, or maybe just the borders of the boundary are draggable.

Yeah, agreed. I thought the same when I first got it going - but after playing with it so much it got familiar. I'll be curious to hear what your team thinks, first impressions are key. In my local env I don't have many items in the DB, but I hoped that with live data when you dragged the boundary you'd still see the URL change and the various counts in the tabs change, so you'd see that the drag action you just performed did in fact _do_ stuff - and from that infer what's going on. But at first it's definitely not obvious, agreed. 

I think disabling drag might be a bit weird and maybe feel like it was broke... Feels like the behaviour should be consistent... Another possibility would be to add in a slightly opaque orange fill (10% or something) to the bounding box, so even when zoomed in you could visually see that you were within a bounding box. That might make a difference. Or even adding in some sort of faint grid within the bounding box, so when you drag you'd see the grid lines move. Though that might be a fiddly one from a technical point of view.

A "drag" handler icon would definitely alleviate the problem, good call. But I quite like how easy it is now, just dragging the shape itself. Feels intuitive (when not zoomed in). 

Sorry, no good answers here. :)

benkeen:
Just so I can keep track + log progress. TODO:

- [x] remove delete icon on map
- [x] make region draggable via icon, positioned center-underneath shape instead of by dragging shape itself
- [ ] check new icon is positioned properly for a circle shape
- [ ] create new drag icon & pass along screenshot for UX/UI for review
- [ ] fix mouse cursor when mousing over drawn shape. Should be same as map (hand to indicate the map is draggable)
- [ ] ensure requests are retriggered properly on resizing shape & manual drag
- [ ] add new button on grid to clear shape (verify with UX on this, since this functionality already existed with the "Clear Boundary [x]" button)
- [ ] allow drawing new shape when one is already on screen 
- [ ] no longer disable the drawing icons. This'll require verifying a ton of scenarios - will list when I get to this.

benkeen:
Here we go @kueda! Note: I squashed & force-pushed on the branch to clean things up. 

As always - feedback very welcome!

kueda (approved):

pleary (commented):
Here's a couple things I noticed. Probably not blocking, but imo the first few are worth fixing. Overall I think things are looking and working great. This will be a very nice improvement to observation searching. 

- The "Redo search in map" is still missing the circle icon - I think we planned on adding that back
- The font color for "Redo search in map" and the new "Reset" button is #eee and it used to be pure white #fff. I think it would be good to go back to white
- If you draw a shape, then draw another shape, then click the back button. You'll see results from the first drawn shape, but there is no boundary, and the drag icon from the second shape erroneously remains. Ideally the back button would restore the original boundary and drag icon. Looks like the boundaries get lost in production now too with "Redo search in map" and the back button. So this isn't entirely new - but the vestigial drag icon is new, and it might get more noticeable if people use boundaries more with this neat drawing tool
- This also seems to happen sometimes in production currently. If you draw a shape, then click on any other view (grid, list, species, identifiers, or observers), then click the back button. This time the search results don't include any boundary but the drag icon remains. Ideally this would maintain the boundary parameters, and restore the drawn boundary and drag icon
- Very minor, but after drawing a rectangular boundary, the "drag-to-reshape" circles in the corners of the boundary are opaque white and on top of the boundary, but the ones in the middle of the sides are transparent. This makes me think they are not all the same, but I think they actually are. Can we make them all the same opaque white?
- I don't think there's much to improve here, but I'll point out an interesting case I noted that other people may hit. When zoomed way out and using the circle draw tool near the poles or around the dateline, you can draw some weird invalid shapes that look line sine waves or inverted circles. After you end the drawing, you end up with no boundary at all and a search of the world as the shapes were invalid. Same if you start with a valid circle then drag it up to the North Pole - the shape will disappear. These are edge cases that again I don't think need to be addressed, just pointing out it can happen.

benkeen:
Great stuff, thanks @pleary!

> - The "Redo search in map" is still missing the circle icon - I think we planned on adding that back
> - The font color for "Redo search in map" and the new "Reset" button is #eee and it used to be pure white #fff. I think it would be good to go back to white

Yes, I kinda went rogue here... :) I left the icon out partly for consistency with the new Reset button, but mostly for space reasons. That row is getting pretty long and will be more likely to wrap now. And well noticed on the colour change. For that I wanted a visual indicator when the user is mousing over (the text + icons now go white). I swiped the existing styles for the grey "Filters" button for those styles.

Happy to make these changes back but could we maybe run it by Abhas to get his input? 

> If you draw a shape, then draw another shape, then click the back button. You'll see results from the first drawn shape, but there is no boundary, and the drag icon from the second shape erroneously remains. Ideally the back button would restore the original boundary and drag icon. Looks like the boundaries get lost in production now too with "Redo search in map" and the back button. So this isn't entirely new - but the vestigial drag icon is new, and it might get more noticeable if people use boundaries more with this neat drawing tool
> This also seems to happen sometimes in production currently. If you draw a shape, then click on any other view (grid, list, species, identifiers, or observers), then click the back button. This time the search results don't include any boundary but the drag icon remains. Ideally this would maintain the boundary parameters, and restore the drawn boundary and drag icon

Mmm thanks... I'd say this was a blocker. Let me look at it.

> Very minor, but after drawing a rectangular boundary, the "drag-to-reshape" circles in the corners of the boundary are opaque white and on top of the boundary, but the ones in the middle of the sides are transparent. This makes me think they are not all the same, but I think they actually are. Can we make them all the same opaque white?

I *think* this may not be doable but I'll take another look. I originally tried to remove the center point because I only wanted it to be draggable via the new icon, but Google's DrawingManager didn't provide that option, nor could I find a way to target it in the DOM. Think it may be the same problem with all those little circles. 

> I don't think there's much to improve here, but I'll point out an interesting case I noted that other people may hit. When zoomed way out and using the circle draw tool near the poles or around the dateline, you can draw some weird invalid shapes that look line sine waves or inverted circles. After you end the drawing, you end up with no boundary at all and a search of the world as the shapes were invalid. Same if you start with a valid circle then drag it up to the North Pole - the shape will disappear. These are edge cases that again I don't think need to be addressed, just pointing out it can happen.

Yeah, this wasn't pretty, I just couldn't think of a way to accommodate it. The library doesn't handle rendering a circle at the poles, so I had to catch it and simply remove it. We _could_ do a toast or something "Invalid region!", but I figured that was well out of scope.

abhasinat:
Here are my UI changes:

- I'd like the redo search button to have the refresh icon in it. Right now, I don't think adding the icon will affect the wrapping on most screen sizes. And it doesn't need consistency with the reset button imo, but good thought!
- I'd like the reset button's text to be white #ffffff, the darker text makes it look like the text may be see-through because it looks like the map's ocean color. I think changing the color of the orange on hover is enough to indicate a hover state.
- Right now, the move icon has a transparency applied to it. I'd prefer if it was at 100% transparency, and for the white arrows to be visible as white rather than subtracted, appearing see-through into the map. I'm pretty sure the SVG I shared had the arrows as white, but if you need a new file let me know!

benkeen:
Excellent, thanks @abhasinat - I'll update it tonight.

benkeen:
Hey @abhasinat, I applied all your changes - thanks! Good call on the white for the arrows, it's much clearer now. I did have to ditch the previous method of using inline SVG content for the marker due to the 2nd colour, but I cached the request to the drag icon image so it's not too bad. 

@pleary the back/forward btn behaviour is very solveable and I agree it's a bit worse. If there's no great rush, give me another night or two and I'll try to resolve it. There's already an window.onpopstate hooked up in the code, so should just be a matter of some code shuffling to check it's being reset properly.

app/assets/javascripts/ang/controllers/observation_search.js
```diff
@@ -90,6 +90,16 @@ function( ObservationsFactory, PlacesFactory, TaxaFactory, shared, $scope, $root
   $scope.dragRectDeltaXInPx = null;
   $scope.dragRectDeltaYInPx = null;
   $scope.dragCircleDeltaInKm = null;
+  $scope.dragIconMarker = new google.maps.Marker({
+    position: { lat: 0, lng: 0 },
+    icon: {
+      anchor: { x: 12, y: 12 },
+      url: "assets/drag-icon.png",
```
kueda, line 97: Hard-coding the path to a file in the asset pipeline won't work in production. We usually handle this by

1. Renaming the file `observation_search.js.erb`. This will cause the file to get run through the ERB interpreter when the assets get compiled.
2. Using the asset helper to insert the correct URL
    ```javascript
    url: "<%= asset_path( "drag-icon.png" ) %>",
    ```

Also, can we use an SVG here? I think Abhas sent you one. Not sure if it works with Google, but we can definitely serve an SVG assets just like we can a PNG. We use SVGs for a lot of our icons and illustrations.

app/assets/stylesheets/observations/search.scss
```diff
@@ -736,13 +736,16 @@ inat-taxon { white-space: nowrap; }
   }
   .btn-redo-search {
     padding: 7px 7px 5px !important;
+    .glyphicon {
+      margin-right: 2px;
+    }
```
kueda, line 741: This needs to be white as well.

kueda (changes requested):

app/assets/javascripts/ang/controllers/observation_search.js
benkeen: Thanks @kueda, updated.

app/assets/stylesheets/observations/search.scss
benkeen: Updated.

benkeen:
Hey folks, thanks again for all the help - I really appreciate your patience with all this! I think everything's looking good other than these: 

- Like I feared, @pleary, I couldn't find a way to target those central resizable circle icons for the rectangles. It's a simple thing: they just have a hardcoded 0.5 opacity, but the google component doesn't provide a way to add a class or some other css-targetable property, and the DOM elements are in some nested div elements. 
- The back/forward button problem is definitely solvable, but it's proving to be fiddly. I'm out of time for a bit (off on holiday on Sat!) so for now I just did a quick one-liner clean-up to remove the drag icon when you navigate back or forward. So it should be at least no _worse_ than now. But I'd be happy to either address this in this PR or in a follow-up PR. Whatever people fancy.

kueda (approved):

kueda:
Thanks for all this work, @benkeen. This will make a lot of people happy, including me. If you think you can address the back button issue in a separate PR that would be most welcome, but we didn't think it needed to block.

benkeen:
Thanks @kueda, no problem will do. :)

Issue #3444, opened 2022-05-17
In search results at admin/users, link observations number to Explore, not Edit Observations


tiwane:
The obs number link at admin/users goes to Edit Observations, but I almost never want to go there, I want to see that user's observations on Explore. So the link should point me to Explore, not Edit Observations.

grfeller:
Hello, all. I am new to iNaturalist/Rails development and would like to check my understanding of this issue before creating a PR.

The following steps/screenshots are based on app behavior on my development system.

When logged in as a user with `admin` privileges:

1. Open a view on `admin/users`

![fig-1-admin-users](https://user-images.githubusercontent.com/16858088/194361661-a8f93c96-800d-453e-a2c6-fd9930595801.png)

**Current behavior**

2. Follow the link in the `obs` column for the `testerson` user (above) and see the  `Edit Observations` view

![fig-2-edit-observations](https://user-images.githubusercontent.com/16858088/194560978-6440e7b3-42f8-4139-99fe-511d75c937ba.png)
 
**Desired behavior**

3. Follow the link in the `obs` column for the `testerson` user and see the `Explore` view

![fig-3-explore-observations](https://user-images.githubusercontent.com/16858088/194361804-38f0fe23-eab3-4295-a085-fd2d52272279.png)

Does the screenshot for `step 3` show the desired behavior? 

**Proposed change**

![diff-change](https://user-images.githubusercontent.com/16858088/194572042-a97e856c-597e-44bf-933d-88ff0a240dbc.png)

Does this change look reasonable? Is there a better approach?

PR #3508, opened 2022-09-01
Simplify conditional logic in `Comment#deletable_by?`

frewsxcv:
Trying out a PR in the Rails app

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -83,12 +83,10 @@ class Comment < ApplicationRecord
   end
 
   def deletable_by?( deleting_user )
-    return false if deleting_user.blank?
-    return true if deleting_user.id == user_id
-    return true if deleting_user.id == parent.try_methods( :user_id )
-    return true if deleting_user.is_admin?
-
-    false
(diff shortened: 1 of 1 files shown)


kueda:
Effort much appreciated, @frewsxcv, but this actually changes the logic of this method such that all conditions are required, whereas what we want is to allow a user to delete a comment if they created the comment OR if they created the resource the comment was made on OR they are an admin. Furthermore, I prefer the guard-clause style as I find it easier to read.

If you're looking for a good first issue in the Rails repo, might I recommend #3444? That should just require changing a link. You might also want to read through https://github.com/inaturalist/inaturalist/blob/main/CONTRIBUTING.md.

frewsxcv:
Ah good catch. So the conditional logic could be:

```ruby
deleting_user.present? && (
    deleting_user.id == user_id
      || deleting_user.id == parent.try_methods( :user_id )
      || deleting_user.is_admin?
)
```

kueda:
That looks like the right logic, but again, I think that style is harder to read so I'm not interested in merging in that change.

PR #3519, opened 2022-09-16
fix issue #3110 - sorting by date added

AdamPRankin:
Sorting obs by "Date Added" now correctly sorts by date added instead of date observed


kueda (changes requested):
This doesn't actually fix the issue, i.e. when I run this locally it still seems to sort by date observed. Did you see it working when you ran it locally?

Issue #3527, opened 2022-10-07
Hide serial numbers in EXIF data from public view


tiwane:
See https://forum.inaturalist.org/t/serial-numbers-in-exif-data/35153

Some photos include equipment serial numbers, eg `Lens serial number` here: https://www.inaturalist.org/photos/233472508 While unlikely, it's possible the serial number could be used to track someone across different sites (eg iNat and Flickr) and/or be used to make false warranty or theft claims, so we should hide it from public view like we do photo file names.

I tried out a few raw photos from Sony, Nikon, and Canon cameras in RawDigger and the only fields with "serial" in them are 

`Lens serial number`
`Serial number`
`Internal serial number`

so we should hide all of those.

todtb:
I can take a look at this. I'm thinking:
1) Hide the serial number fields in the UI
2) Prevent future extraction of EXIF serial number fields (if possible)

PR #3548, opened 2022-11-06
Hide serials in exif display and extract

todtb:
#3527 

- Updated fixtures with exif meta data `:lens_serial_number`
- Refactored EXIF display for `Photo` by pulling into helper
- Excluding display serial number tags based on `exiftool`'s list of supported tags (see https://github.com/inaturalist/inaturalist/compare/main...todtb:3527-hide-serials-in-exif?expand=1#diff-14c62713d26766370b2cff9e80bcc20a6f53b555e3d657d873eddc3d546ff130R5)
- Prevent these tags from being extracted into `Photo`'s associated `metadata`


app/helpers/photos_helper.rb
```diff
@@ -0,0 +1,46 @@
...
+    end.to_h
+  end
+
+  def visible_metadata( photo:, with_coords: )
+    return photo.metadata.select { _1.to_s =~ /copyright/i } unless logged_in?
+
+    photo.metadata.dup.tap do | md |
+      md.except!( *REJECTED_TAGS )
+      md.reject! { _1.to_s =~ /description/i }
+      md.reject! { _1.to_s =~ /(gps|date)/i } unless with_coords
```
kueda, line 39: `_1` is a new piece of syntax for me, and it's something I think I've always wanted without knowing how to search for it. Is it actually documented anywhere?

app/helpers/photos_helper.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+module PhotosHelper
+  REJECTED_TAGS = ( ExifMetadata::REJECTED_TAGS + %i[dc dimensions] ).freeze
+
+  def metadata_table( photo )
```
kueda, line 6: Can we call this `metadata_table_rows` since that's what it renders?

kueda (changes requested):
Works for me, only one very minor change requested.

app/helpers/photos_helper.rb
todtb: Yeah, that makes sense 👍

kueda (approved):

Issue #3540, opened 2022-10-27
Journal search


kueda:
I frequently want to search for posts in the official iNat blog, so IMO we just need a simple, database-based search for posts in a journal, like the `dbsearch` scope for guides.

kueda:
Closed in #3549

PR #3549, opened 2022-11-08
Add pg ilike search to posts title and body

todtb:
#3540

- Add pg ilike search to `posts.title` and `posts.body`
- Add search form to `/journal` (browse)
- Add post search results page which shows results, paginated

## Browse:
image

## Results:
image

image

### DIFF
--- a/app/views/posts/_header.html.haml
+++ b/app/views/posts/_header.html.haml
@@ -17,44 +17,32 @@
           - if @display_project.curated_by?(current_user)
             = link_to t(:new_post), new_project_journal_post_path(@parent.journal_slug), id: "newpostbutton",
                 class: "default button"
-        = render partial: 'search_form', locals: { parent_type: @parent.class.name, parent_id: @parent.id }
 - elsif @parent.is_a?(Project)
   - tools = capture do
-    #post_header
-      %div
-        - if @display_project.curated_by?(current_user)
-          #tools.buttonrow
-            = link_to t(:new_post), new_project_journal_post_path(@parent.journal_slug), id: "newpostbutton",
-              class: "default button"
-      = render partial: 'search_form', locals: { parent_type: @parent.class.name, parent_id: @parent.id }
+    - if @display_project.curated_by?(current_user)
+      #tools.buttonrow
+        = link_to t(:new_post), new_project_journal_post_path(@parent.journal_slug), id: "newpostbutton",
+          class: "default button"
   = render :partial => "shared/by_project_header_bootstrap", :locals => { :before => breadcrumbs, :after => tools, :model_name => "Journal", :project => @parent }
 - elsif @parent.is_a?( Site )
-  .row
-    #pageheader.col-md-12
-      #post_header
-        %div
-          = breadcrumbs
-          - if @site.logo_blog?
-            = image_tag @site.logo_blog.url
-          - else
-            = link_to(image_tag(@site.logo_square.url(:thumb), :class => "parenticon"), @site)
-            %h2
-              = link_to @site.name, site_posts_path
-            %ul#subnav.clear
-              %li=t :official_blog
-          - if @site.editable_by?( current_user )
-            #tools.buttonrow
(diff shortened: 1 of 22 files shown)


config/routes.rb
```diff
@@ -499,6 +499,7 @@
       get :for_user
     end
   end
+  get "/journal_search", to: "posts#search"
```
kueda, line 502: Let's make this `/posts/search` instead.

kueda (changes requested):
Many thanks, working for me, and sort of meets my original need, but not quite. To do that I'd like the following:

- [ ] Add the search boxes to posts/index (right-aligned in posts/_header would work for me)
- [ ] posts/search should take `parent_type` and `parent_id` params to scope search to a particular journal (doesn't have to be implemented that way, but I want journal-specific search so I can find particular posts in the iNat (site) journal, or my (user) journal)
- [ ] posts/search should show what journal you're searching

Sorry, another case of me writing an under-specified issue.

todtb:
Ok, reworked this a bit so that it applies to all `parent_types` (`User`, `Project`, `Site`). Since I was doing this I took the opportunity to create a `HasJournal` concern for those models, thus helping clean up the controller flow a bit.

Also removed original "Browse" search as I see that wasn't really the primary goal here and searching ALL might not be that helpful.

N.B.
- Does change `Project`
`has_many :journal_posts, class_name: "Post", as: :parent`
to
`has_many :journal_posts, class_name: "Post", as: :parent, dependent: :destroy`
Though I'm assuming it should be this way and consistent with other models now, I highlight his as it is a change.

## Project (new)
image
## Project (old)
image
## Site
image
## User (self)
image
## User (other)
image

app/views/posts/search.html.haml
```diff
@@ -0,0 +1,21 @@
+- content_for :title do
+  = @title = "#{t(:find_posts_for)} #{@parent.journal_display_name}"
```
kueda, line 2: Translators will not like this b/c they can't control the position of the journal name, they won't be able to control for journals that begin with a vowel or consonant, and they won't be able to control for gender of the journal name. Here's an alternative that skirts two of these issues:

```ruby
= @title = t( :find_posts_for, vow_or_con: @parent.journal_display_name[0].downcase, journal_name: @parent.journal_display_name )
```

And then in en.yml:

```yaml
en:
  find_posts_for: "Find Posts for: %{journal_name}"
```

Gender could still be a problem, but there's no perfect way to deal with that with arbitrary journal names.

app/views/posts/_header.html.haml
```diff
@@ -2,32 +2,59 @@
...
+  .row
+    .col-md-12
+      - title = t( :x_project_journal, project: strip_tags( @parent.title ), name: t( :journal ) )
+      %h1= title
+      #post_header
+        %div
+          - if @display_project.curated_by?(current_user)
+            = link_to t(:new_post), new_project_journal_post_path(@parent.journal_slug), id: "newpostbutton",
+                class: "default button"
+        = render partial: 'search_form', locals: { parent_type: @parent.class.name, parent_id: @parent.id }
```
kueda, line 20: I'm not a fan of the extraneous whitespace this creates below the title when the viewer can't see the "New Post" button. Could we remove the search form from the header at put it at the top of the right column instead?

kueda (changes requested):
Works great, and I like the concern. Just a few more changes and I think we'll be there.

app/views/posts/search.html.haml
todtb: Excellent call, and a great lesson for thinking in i18n

app/views/posts/_header.html.haml
todtb: With the right column being the "archives" section? Something like:
## own
image

## other user
image

## project
image


## w/o archives:
image
### blog
image

kueda: Exactly. Those screenshots look good to me.

kueda (approved):

kueda:
Thanks again!

Issue #3418, opened 2022-04-06
Editing comment on journal that disbaled comments results in untranslated and unclear error message


tiwane:
From https://forum.inaturalist.org/t/missing-translation-error-on-editing-journal-comment-when-comments-are-disabled/30808 I can replicate.

Step 1: Create a journal post.
Step 2: Comment on the post.
Step 3: Disable comments.
Step 4: Try and edit your comment.

See the following error

![image](https://user-images.githubusercontent.com/15896181/161871326-da4342a4-4e53-4143-a012-e6d6712335b1.png)

Since it's possible for end users to see that message, we should make it more clear and translatable. Maybe something like "Comment failed to save because comments are disabled for this journal post."

PR #3552, opened 2022-11-13
Add translation

todtb:
#3418 

Adds the translation.

Alternatively, we could add a message here: 
```rb
# app/models/comment.rb:118
errors.add( :parent, :prefers_no_comments,  I18n.t(:prefers_no_comment))
```
and reference the existing post translation, though I feel that's less explicit.

Side-note/question:

Should comments made prior to marking the post as non-comment-able be visible at all? (see edit screenshot) 
It's not really a "new comment" either...

## post
image

## edit 
image

### DIFF
--- a/app/views/comments/_comments.html.erb
+++ b/app/views/comments/_comments.html.erb
@@ -1,5 +1,6 @@
 <%- 
   parent ||= comments
+  parent_comments_disabled ||= parent && parent.respond_to?(:prefers_no_comments?) && parent.prefers_no_comments?
   hide_form ||= false
   hide_form = true if current_user && current_user.blocked_by?( parent.try(:user) )
   remote ||= nil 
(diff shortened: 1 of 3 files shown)


kueda:
> Should comments made prior to marking the post as non-comment-able be visible at all? 

Yes. While we often use this feature to prevent any comments, we've also used it to shut down conversation when we don't think it's providing further benefit. In the latter case, I think it's better to show all the previous comments so people can see for themselves why we wanted to shut it down. If the entire conversation is so troubling that it should be completely hidden, I think we need to make a different feature.

kueda (changes requested):
This solves the problem in #3418, but I think the messaging is still confusing, especially when editing an existing comment. Maybe we could change it to "has disabled comments"?

I'd also prefer to hide the edit button if comments have been disabled. I don't want people spending a lot of time rewording their comment only to learn it can't be saved.

kueda (approved):

Issue #3264, opened 2021-10-06
Fix FakeView in Rails 6


kueda:
Our FakeView class doesn't seem to work under Rails 6 due to some changes to ActiveView. FakeView was never a good idea to begin with, and a lot of what we're using it for (providing access to helpers outside of views, a global util class), can be handled differently in more recent versions of Rails, so we should try to ditch it. IMO, this can be divided thusly:

- [x] **Use [`helpers`](https://apidock.com/rails/ActionController/Helpers/helpers) in controllers.** This was introduced in Rails 5 to do exactly what we're using FakeView for in controllers.
- [x] **Remove it from views.** FakeView is redundant inside of a view, not sure why we have it there at all.
- [ ] **Move helper methods that don't require a view context into models or other utility classes.** For example, `ApplicationHelper#uri_join` probably doesn't need to be in a helper.
- [ ] **Replace URL helpers in models with...** This is where I get a bit stuck. For some things, like saving an image url for indexing or getting a path for cache expiry, URL helpers are really useful. Do we still need some class that can construct URLs and is accessible from models, or are there better ways of doing this?

kueda:
Well, I did the easy stuff. Not sure what to do about uses in models and elsewhere, like DarwinCore. We need some way to render templates and generate URLs outside of a controller/view context, so maybe something like FakeView is an ok way to do it. A lot of the logic in the models seems like it should be elsewhere, just not sure where.

todtb:
Hey @kueda,

Re: 
> Replace URL helpers in models with... This is where I get a bit stuck. For some things, like saving an image url for indexing or getting a path for cache expiry, URL helpers are really useful. Do we still need some class that can construct URLs and is accessible from models, or are there better ways of doing this?

I think `Rails.application.routes.url_helpers` could be of some use here. 

Another pattern inside of controllers is to use `helper_method`
```rb
# app/models/post.rb:162
def to_plain_s(options = {})
    s = self.title || ""
    s += ", by #{self.user.try(:login)}" unless options[:no_user]
    s
end
helper_method :to_plain_s
```
which allows the method to be used in controllers where needed and exposed to the associated views.

Is this still work that is on the table?

kueda:
Yeah, I think this is still on the table. We've been running on Rails 6 for a while now so clearly whatever we did is working, but `FakeView` remains an upgrade liability, so I think any good ways for refactoring `FakeView` out of the models would be a good thing.

todtb:
[...]

PR #3561, opened 2022-11-18
Replace FakeView url helpers in models

todtb:
#3264

First step in removing `FakeView`. 

- Creates `UrlHelper` which handles delegating to `Rails.application.routes.url_helpers` in a single purpose class. 
  - This should be a large portion, if not the majority, of `FakeView` usages.
- The next big step will be allow calls to these from outside of the view context:
  - `image_url`
  - `uri_join`
- Also need to address :
  - Calls to `render`
  - Calls to `ActionView::Helpers`
  - DarwinCore (not too familiar with what this is yet)

### DIFF
--- a/app/models/guide_section.rb
+++ b/app/models/guide_section.rb
@@ -146,7 +146,7 @@ class GuideSection < ApplicationRecord
   def reuse
     gs = clone
     gs.rights_holder = attribution_name
-    gs.source_url = FakeView.guide_taxon_url(guide_taxon_id)
+    gs.source_url = UrlHelper.guide_taxon_url guide_taxon_id
     gs
   end
 
--- a/app/models/guide_section.rb
+++ b/app/models/guide_section.rb
@@ -146,7 +146,7 @@ class GuideSection < ApplicationRecord
   def reuse
     gs = clone
     gs.rights_holder = attribution_name
-    gs.source_url = UrlHelper.guide_taxon_url guide_taxon_id
+    gs.source_url = UrlHelper.guide_taxon_url( guide_taxon_id )
     gs
   end
 
--- a/app/models/list.rb
+++ b/app/models/list.rb
@@ -181,9 +181,8 @@ class List < ApplicationRecord
     key
   end
 
-  def self.icon_preview_cache_key(list)
-    list_id = list.is_a?(List) ? list.id : list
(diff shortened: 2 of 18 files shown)


kueda (approved):
This all looks good to me, at least in that it removes a reference to a controller instance from model code (view FakeView). I'm curious what @pleary thinks, though.

pleary:
This looks good to me too for the cases covered so far, with a cleaner helper model.

One minor style suggestion - there are a few cases of method calls without parenthesis (e.g. [`UrlHelper.instance.respond_to? method, include_private` ](https://github.com/inaturalist/inaturalist/pull/3561/files#diff-f98dab56f29862edb3b9003dde7224a286c23262bef7328698c4864c4ade1c6fR13), [`UrlHelper.guide_taxon_url guide_taxon_id`](https://github.com/inaturalist/inaturalist/pull/3561/files#diff-ec5ee2729b46bc74523dfce0336b5261434250e573999d7328ee8330e4526717R149), [`UrlHelper.icon_preview_list_url list_id, locale: I18n.locale`](https://github.com/inaturalist/inaturalist/pull/3561/files#diff-52b06c9d36e8ab2e7afe4f11f44ce8a76306adf2d69e6df46c43e07e605e494aR186), and a couple [`expire_listed_taxon lt`](https://github.com/inaturalist/inaturalist/pull/3561/files#diff-d3385372358f836a55a3b72ad1d2f2ddaf93be391ddd610de4e3006cf03b3ceeR19)). While we don't have a rubocop rule for it (nor do we actively use rubocop), we tend to add parenthesis for most method calls when there are arguments, though there are exceptions (e.g. `puts`, and a lot of spec methods).

Thanks for the contribution!

todtb:
@pleary - Think I got them all, including adding some previously missing ones on lines that were changed as well (ie: `ctrl.send :expire_action, FakeView.home_url( user_id: user_id, ssl: true )`). I used to almost aways use parens, before switching to a "when necessary" approach (though still pretty much always on method definitions). I think it was noticing the rails idiomatic:
```rb
belongs_to :user
...
before_action :authenticate_user!
...
render json: @user
...
etc
```
that changed my mind. Plus it can save a few characters 😄 

In any case, should be set now. Thanks for taking a look!

Issue #3419, opened 2022-04-06
Add Identify keyboard shortcut to play observation's sound


tiwane:
From: https://www.google.com/url?q=https://forum.inaturalist.org/t/on-the-identification-screen-create-a-shortcut-key-to-play-the-first-sound-file-for-the-observation/30156

If one wants to play an observation's sound in the Identify modal, one has to use a mouse. Request is to add a keyboard shortcut to play the first sound of an observation. Ken-ichi suggested SPACE rather than CMD+P to not confuse it with the Print command.

grfeller:
Hello, I have a possible approach for this request, but would like to check my understanding before creating a PR. (I am new to React/Rails.)

**Desired behavior**

1. User opens `Identify` view and selects an observation with one or more associated sounds.

![f1-io-wsnd](https://user-images.githubusercontent.com/16858088/202496440-c8f97646-05d9-40da-bc42-2873711560e2.png)

2. User presses `Space` and the first sound is played. If the sound is already playing, it will pause.

3. User toggles Keyboard Shortcuts and sees mapping for the `Space` key.

![f1-ks-pfs](https://user-images.githubusercontent.com/16858088/202491772-53f8b4ba-ea57-4bb8-b06d-50adcd2c8c13.png)

**Implementation questions**

On my development system, I created a new `action` called  `togglePlayFirstSound` :

```
export function togglePlayFirstSound( ) {
  return ( ) => {
    const player = $(".obs-media .sounds").find("audio:first")[0];
    if ( !player ) { 
      return; 
    }
    if ( player.paused ) {
      player.play( );
    }
    else
    {
      player.pause( );
    }
  };
}
```

Does this seem like a reasonable approach? (The jquery to get the audio element seems a little brittle to me,) 

Thanks in advance for any suggestions for improvement,

kueda:
@grfeller that looks good to me. FWIW, whether you use jQuery or `getElementBy*`, I don't think there's another way to manipulate those dom elements. Maybe you could use a ref with the React `<audio>` component and control it with state, but that seems much more brittle to me.

My only additional request would be that in addition to playing the sound, you focus on the `<audio>` element. That way people can TAB to the next audio element, and someone using a screen reader will have a little more control.

PR #3562, opened 2022-11-21
Add Identify keyboard shortcut to play observation's sound

grfeller:
[#3419](https://github.com/inaturalist/inaturalist/issues/3419)

Set focus to first `<audio>` element and play associated sound file. If an `<audio>` element is already active, the `togglePlayFirstSound` action exits early to avoid interfering with default processing.

### DIFF
--- a/app/webpack/observations/identify/actions/current_observation_actions.js
+++ b/app/webpack/observations/identify/actions/current_observation_actions.js
@@ -1128,20 +1128,17 @@ export function subscribe( ) {
 
 export function togglePlayFirstSound( ) {
   return ( ) => {
-    const player = $(".obs-media .sounds").find("audio:first")[0];
-    const is_active_element_audio = $(document.activeElement).is("audio");
-    if ( !player ) { 
-      return; 
-    }
-    // Use default processing if an audio element is already active.
-    if ( is_active_element_audio ) {
+    const player = $( ".obs-media .sounds" ).find( "audio:first" )[0];
+    const hasFocus = player === document.activeElement;
+    if ( !player ) {
       return;
     }
-    player.focus();
+    if ( !hasFocus ) {
+      player.focus();
+    }
     if ( player.paused ) {
       player.play( );
-    }
-    else {
+    } else {
       player.pause( );
     }
   };
--- a/app/webpack/observations/identify/actions/current_observation_actions.js
+++ b/app/webpack/observations/identify/actions/current_observation_actions.js
@@ -1126,6 +1126,27 @@ export function subscribe( ) {
   };
 }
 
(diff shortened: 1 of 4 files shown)


app/webpack/observations/identify/components/observation_modal.jsx
```diff
@@ -391,6 +391,7 @@ class ObservationModal extends React.Component {
       { keys: ["i"], label: I18n.t( "add_id" ) },
       { keys: ["f"], label: I18n.t( "add_to_favorites" ) },
       { keys: ["z"], label: I18n.t( "zoom_photo" ) },
+      { keys: ["Space"], label: I18n.t( "play_first_sound" ) },
```
kueda, line 394: Please make "Space" all-caps to match the other named keys like SHIFT.

app/webpack/observations/identify/actions/current_observation_actions.js
```diff
@@ -1126,6 +1126,27 @@ export function subscribe( ) {
...
+    player.focus();
+    if ( player.paused ) {
+      player.play( );
+    }
+    else {
+      player.pause( );
+    }
+  };
+}
+
```
kueda, line 1149: Please abide by the eslint rules, at least for the code that you're adding. You can run `npx eslint --fix app/webpack/observations/identify/actions/current_observation_actions.js` if your editor doesn't automatically show you the eslint errors.

kueda (changes requested):
Works well in Firefox. In addition to the inline comments, here are some bugs / requests

- [ ] In Safari, SPACE will play the sound and focus on the player, but hitting SPACE again will not pause; seems like maybe SPACE does nothing while the player has focus, but only in Safari

I guess that was just one bug, singular. If you can find a workaround for Safari, great, but I think this is a step forward even if that seems impossible.

grfeller:
Firefox, Chromium and Safari should now behave as described (at least on macOS 12 and Ubuntu 20.04):
- SPACE will play the sound and focus on the *first* player. Hitting SPACE again will pause. 
- To avoid confusion, I removed the processing for subsequent `<audio>` elements; this makes the behavior *Play first sound* rather than *Play first / current sound* as previously implemented.

**Implementation**
- Binding to `keydown` instead of `keyup` suppressed the Firefox [shortcut for Toggle Play / Pause](https://support.mozilla.org/en-US/kb/keyboard-shortcuts-perform-firefox-tasks-quickly#w_media-shortcuts),

kueda (approved):

kueda:
Thanks, @grfeller! Sorry it took me a while to get back to this.

grfeller:
@kueda thanks for your help and feedback in making this PR successful.

PR #3577, opened 2022-12-06
identifications and registrations rtl

naamaarkin:


.vscode/settings.json
```diff
@@ -0,0 +1,22 @@
+{
```
jtklein, line 1: This is a file with settings for your local code editor and shouldn't be included in the pull request.

kueda: Yeah, I would add this to the `.gitignore` as well.

app/assets/stylesheets/bootstrap_bundle.scss
```diff
@@ -214,6 +214,9 @@ body > div > .popover {
     @import "bootstrap-10";
   }
   @import 'bootstrap-inat';
+  [dir="rtl"] & {
+    @import "rtl-bootstrap";
+  }
```
kueda, line 219: Also, assuming this is the way to go, this needs to be imported beofre bootstrap-inat, otherwise all the overrides in `bootstrap-inat` don't work (you can see this in the `btn-inat` button border radius).

kueda (changes requested):
Couple minor things to fix in the comments, and I need some convincing that we really need an entirely separate version of bootstrap to deal with RTL.

app/assets/stylesheets/bootstrap_bundle.scss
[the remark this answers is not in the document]
naamaarkin: Not sure I completely follow. Full duplication of bootstrap.css is not necessary, but to completely adjust it to rtl you need to do around 800 modifications. I can leave it as it is, but I guess that the single page that I will localize wouldn't look like it supposed, because of the cascading from bootstrap.
I also assumed I need to create this file based on looking at your rtl branch where you added  bootstrap-rtl.min.css and added it in the same manner I added it here 
https://github.com/inaturalist/inaturalist/blob/1c41297d6d198d3b37a1ca7371c4a7303c2a0579/app/assets/stylesheets/bootstrap_bundle.scss#L216-L218
I could also replace anything related to direction in bootstrap.css to logical property but then it can potentially have too big of affect on your production. 
Is by overrides you mean working like this? 
https://github.com/inaturalist/inaturalist/blob/1c41297d6d198d3b37a1ca7371c4a7303c2a0579/app/assets/stylesheets/index.scss#L14-L17
or maybe like this?
https://github.com/inaturalist/inaturalist/blob/1c41297d6d198d3b37a1ca7371c4a7303c2a0579/app/assets/stylesheets/users/registrations.scss#L37-L39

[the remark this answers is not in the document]
kueda: That’s definitely what I mean by overrides, just curious if you considered that approach and rejected it for some reason. If we don’t use overrides, we should probably make a completely separate rtl bundle, otherwise we’ll serve a lot of unused CSS (rtl csss for ltr people and vice versa). Did you try that and reject it for some reason?

naamaarkin:
Hey all, I just wanted to explain what I did in my last commit:
 1. Bootstrap.css is back to non-logical properties. The logical properties code is in bootstrap-rtl.css.
 2.  bootstrap-rtl isn't in the bootstrap bundle, but it's loaded conditionally after the bootstrap bundle in the main layouts (application and bootstrap erb).
 3. Some of the glyphicons have a direction that needs to be switched in rtl, so I overrode them with bootstrap-rtl icons (e.g. right arrows will become left).
  4. I saved bootstrap-rtl.css as a scss file, and then nested everything inside bootstrap class, so it would override.

app/assets/stylesheets/bootstrap.css
```diff
@@ -6118,7 +6117,6 @@ button.close {
   font-weight: normal;
   line-height: 1.42857143;
   text-align: left;
-  text-align: start;
```
kueda, line 6121: Since you're taking an override approach, I would not expect anything to change in this file. Is there a reason these changes can't be overridden in bootstrap-rtl.scss?

app/assets/stylesheets/users/registrations.scss
```diff
@@ -84,7 +84,7 @@
   .form-group.text-field:before {
     position: absolute; 
     top: 8px;
-    left: 13px;
+    inset-inline-end: 13px;
```
kueda, line 87: This should be `inset-inline-start` to match the existing style.

app/assets/stylesheets/users/registrations.scss
```diff
@@ -74,7 +74,7 @@
     background-color: #f8f8f8;
     color: #666;
     border-radius: 7px;
-    padding-left: 41.3px;
+    padding-inline-end: 41.3px;
```
kueda, line 77: This should be `padding-inline-start` to match the existing style.

app/assets/stylesheets/users/registrations.scss
```diff
@@ -19,7 +19,7 @@
   .about-observation {
     position: absolute;
     bottom: 0;
-    left: 0;
+    inset-inline-end: 0;
```
kueda, line 22: This should be `inset-inline-start` to match the existing style.

app/assets/stylesheets/rtl-container.scss
```diff
@@ -0,0 +1,3 @@
+#bodywrapper {
+    direction: rtl;
+}
```
kueda, line 3: Why does this need to be in a separate file? Couldn't it be in bootstrap-rtl.scss?

app/views/layouts/application.html.erb
```diff
@@ -19,6 +22,7 @@
     <%= javascript_include_tag "jquery/jquery-1.9.1.min" %>
     <%= javascript_include_tag "jquery/jquery-ui-1.9.1.min" %>
     <%= javascript_include_tag "bootstrap.min" %>
+    <%= stylesheet_link_tag 'bootstrap-rtl' if @rtl %>
```
kueda, line 25: Please use double quotes whenever possible.

kueda (changes requested):
Ok, looks like you're taking the override approach, which should be fine. RTL users will need to load that additional CSS file but it doesn't seem like too much. Getting closer!

In addition to my comments inline, please fix the following:

- [x] Consent check boxes on sign up are mis-aligned: Screen Shot 2023-01-11 at 11 30 35 AM
- [ ] Remember me check box on log in is mis-aligned: Screen Shot 2023-01-11 at 11 34 24 AM
- [ ] Mini observation labels are not start-aligned on /identifications/:login: Screen Shot 2023-01-11 at 11 39 42 AM


There are a few problems with the header that might be out of scope for this PR, but that will make things look broken for RTL users. Why don't you make this a test for now by only setting @rtl in filters_module.rb if there is a `test=rtl` param? So that line would become `@rtl = params[:test] == "rtl" && ["ar", "he"].include?( I18n.locale.to_s )`. That way we can merge this, test it out, and continue building out RTL support without negatively impacting people viewing the site in RTL languages.

app/assets/stylesheets/bootstrap.css
naamaarkin: I deleted this line in the original file because I thought it is a duplication, Isn't it a duplication?

kueda: This was in the original, which we didn't write. I suspect they added it as a progressive enhancement to help with RTL support. Regardless, I just wouldn't change this file.

app/assets/stylesheets/rtl-container.scss
kueda: Ok, you've added it to bootstrap-rtl.scss, but why does this file still need to exist?

.vscode/settings.json
kueda: @naamaarkin you still need to remove this.

kueda (changes requested):
Looking better, but please address all the unresolved comments and the layout of the mini observations on /identifications/:login.

naamaarkin (commented):
I think I managed to solve some of the issues with the mini observations layout but I couldn't figure out what to do with the align I see on the attributes part in the center(observer, date. location). Can I get some help? 
image

kueda:
@naamaarkin I'll try and take a look at this this week. In the meantime do you think you could try and resolve your conflicts?

kueda (changes requested):
Closer! I think all that remains is that

- [ ] there are the labels like תאריך and מקום are not start-aligned (looks like there's a `text-align: left` that should be `text-align: start`)
- [ ] you need to revert the changes to boostrap.css
- [ ] you need to resolve your conflicts

naamaarkin:
By resolving my conflicts I will change observations.scss, is it OK by you?

kueda:
@naamaarkin sorry, I missed your comment until now. Yes, you'll need to further change observations.scss. You wouldn't have conflicts unless you'd already changed it, so you just need to integrate the changes to that file that have occurred on the main branch.

kueda (changes requested):
I think that in merging from main you over-wrote some of your own changes instead of integrating them, because now I'm seeing this on the identifications page:

Screen Shot 2023-02-07 at 11 01 52 AM

naamaarkin:
@kueda I fixed the override and it looks OK from my side

kueda (approved):

kueda:
Thanks, @naamaarkin! I know that took a while, but I think it's looking better for it.

Issue #3546, opened 2022-11-04
Cannot merge two users who have faved the same observation


kueda:
**Describe the bug**
Admins cannot merge two users that have faved the same observation

**To Reproduce**
1. Create two users and have them both fave the same observation by a third user
1. Go to the user curation page for one of the two users (e.g. http://localhost:3000/admin/users/345), and try to merge with the second user
1. You should get a 500 error about a unique key violation

**Expected behavior**
Faves by the reject on observations that are faved by the keeper should be deleted before merge, so merge should succeed even when both users have faved the same observation

**Additional context**

Exception:

```
ActiveRecord::RecordNotUnique, PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_votes_on_unique_obs_fave" DETAIL:  Key (votable_type, votable_id, voter_type, voter_id)=(Observation, 123 User, 345) already exists.
```

PR #3588, opened 2022-12-14
Delete duplicate faves before merging users

rainhead:
Fixes: inaturalist/inaturalist#3546

I'm happy to factor out some of this responsibility, but it wasn't clear that there was anything generalizable.

I'm using `destroy_all` because I don't know what kind of post-destroy hooks there might be on the records, and because I don't expect there to be enough duplicates for it to be a performance problem.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -659,6 +659,20 @@ class User < ApplicationRecord
   def merge(reject)
     raise "Can't merge a user with itself" if reject.id == id
     reject.friendships.where(friend_id: id).each{ |f| f.destroy }
+
+    # Conditions should match index_votes_on_unique_obs_fave
+    reject.votes
+          .joins( "JOIN votes AS keeper_votes USING (votable_type, votable_id)" )
+          .where(
+            keeper_votes: {
+              voter_type: self.class.polymorphic_name,
+              voter_id: id,
+              vote_scope: nil
+            },
+            votable_type: Observation.polymorphic_name,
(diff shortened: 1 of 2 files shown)


app/models/user.rb
```diff
@@ -659,6 +659,20 @@ def index_observations
...
+          .where(
+            keeper_votes: {
+              voter_type: self.class.polymorphic_name,
+              voter_id: id,
+              vote_scope: nil
+            },
+            votable_type: Observation.polymorphic_name,
+            vote_scope: nil
+          )
+          .destroy_all
```
kueda, line 675: This does concern me a bit. I can think of one user with at least 55,000 faves, so that would cause problems here (though so would attempting to delete that user). On the other hand, I think you'd need to reindex all of the observations these votes are associated with if you used `delete_all`. Any thoughts on this, @pleary?

rainhead: This is destroying just the duplicates. I don't believe I have a way to check, but it seems unlikely that two user records have more than 100 faves in common.

pleary: I agree that there would be a potential performance problem if the two users had a lot of faves in common, but also that is unlikely to be the case. I tested the query against two users with the most votes, where none were in common, and it was quite fast. If we think that case could pop up, then we should turn this into a delayed job. But I think it's unlikely enough that I'm OK with the code as-is.

pleary (approved):

Issue #3601, opened 2022-12-18
Add observation_created_at to DeletedObservation


kueda:
It might help investigate some reports of deleted observations if we knew when the observation was created, not just when it was deleted, so `observation_created_at` in the `deleted_observations` table would store the value of `observation.created_at`.

- [ ] add the `observation_created_at` datetime column to `DeletedObservation`
- [ ] set its value to `observation.created_at` when the `Observation` gets deleted and the `DeletedObservation` gets created

todtb:
#3603 should address this

PR #3603, opened 2022-12-19
Add observation_created_at to DeletedObservation

todtb:
#3601

- Adds observation created at datetime column
- Sets observation created at on deleted observation `after_destroy` of observation

### DIFF
--- a/app/models/deleted_observation.rb
+++ b/app/models/deleted_observation.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class DeletedObservation < ApplicationRecord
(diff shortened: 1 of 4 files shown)


app/models/deleted_observation.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 class DeletedObservation < ApplicationRecord
   belongs_to :user
```
todtb, line 4: Considered validating presence `on: :create`, however not sure it's strictly needed.

kueda (approved):

PR #3605, opened 2022-12-20
3264 Remove More fakeview

todtb:
#3264 Removing more FakeView


todtb:
Wondering if the nil safe (i.e. `compacted` version of URI.join found in the ApplicationHelper is needed such as here https://github.com/inaturalist/inaturalist/pull/3605/files#diff-425128702d852c3911535bca2c48d06dced87616daf0adbb277e30b32ea78a53R102? I see elsewhere `URI.join` is used directly.

spec/controllers/registrations_controller_api_spec.rb
```diff
@@ -81,13 +81,13 @@ def register_user_with_params( params = {} )
   end
 
   it "should assign a user to a site" do
-    @site = Site.make!( url: "test.host" ) # hoping the test host is the same across platforms...
+    @site = Site.make!( url: "http://test.host" ) # hoping the test host is the same across platforms...
```
todtb, line 84: Assuming these should be URLs and not just the domains. Caused an issue `both URI are relative` which assuming all records in `sites` have proper URLs, shouldn't be occurring.

todtb:
Wondering if `TaxonDescribers` was once a gem that has since been brought into the core application? The file structure is a bit non-idiomatic, and, outside of DarwinCore, should be the last uses of `FakeView`. It's needed here to instantiate `ApplicationController` with custom view paths, only necessitated by the fact that these templates are in lib.

todtb:
Hey @kueda and @pleary, 

This PR takes a few approaches to removing `FakeView`:

1) Move model responsibilities out of helpers/controllers and into model concerns eliminating need for FakeView and calling directly on the instance
2) Accessing Rails Active::X helpers directly from ApplicationController helpers. The explicit yet lengthy invocation here (ie `ApplicationController.helpers.strip_tags`) hopefully nudges towards the another approach such as the one above or through some other means, though this is not always possible.
3) Remaining url helpers moved to `UrlHelpers` module with a more specific and limited scope of responsibilities.

Last step is to remove `FakeView` class itself.

spec/controllers/registrations_controller_api_spec.rb
kueda: Yeah, these should be URLs. Not sure why that was working.

kueda (changes requested):
Really sorry for letting this lie so long, @todtb. I think this is looking good. Just need to resolve that one conflict.

kueda (approved):

PR #3618, opened 2023-01-12
Touch the observation when flagging a comment or identification

kueda:
This ensures that the observation's `updated_at` changes after flagging an identification or a comment on that observation, this allowing a client to detect that something has changed with the observation and it probably needs to update its copy. This specifically emerged from trying to implement flagging in the React Native app, where (at present) everything is driven through the local database, so after creating a flag, the app needs to update its local database and the simplest way to do so seemed to be to re-fetch the observation after flagging.

Couple potential problems:

1. We refetch the observation after updating associated records in other clients, but do we still think that's a good approach?
2. If the observation is updated in postgres and reindexed in es during the request to create a flag, can we assume that the API will actually return an up-to-date copy of the observation on refetch, or does the client need to do something to accommodate some indexing lag?

### DIFF
--- a/app/models/comment.rb
+++ b/app/models/comment.rb
@@ -109,6 +109,7 @@ class Comment < ApplicationRecord
   end
 
   def flagged_with(flag, options)
+    parent&.touch
     evaluate_new_flag_for_spam(flag)
     index_parent
   end
--- a/app/models/identification.rb
+++ b/app/models/identification.rb
@@ -380,12 +380,13 @@ class Identification < ApplicationRecord
     true
   end
 
-  def flagged_with(flag, options)
-    evaluate_new_flag_for_spam(flag)
+  def flagged_with( flag, _options )
+    evaluate_new_flag_for_spam( flag )
     elastic_index!
-    if observation
-      observation.elastic_index!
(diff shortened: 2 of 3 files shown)


pleary (approved):
> We refetch the observation after updating associated records in other clients, but do we still think that's a good approach?

Yes this seems reasonable to me. Updates to things like IDs can effect other attributes like quality grade, community_taxon_id, etc, and the logic for changing other attributes is defined in a single place - the rails code. Without duplicating the logic on the client, I think refetching the observation is the most efficient want to get current values for all attributes on an observation and its associations.

> If the observation is updated in postgres and reindexed in es during the request to create a flag, can we assume that the API will actually return an up-to-date copy of the observation on refetch, or does the client need to do something to accommodate some indexing lag?

There should be no indexing lag after modifying an observation or its associations. If there is, I'd consider this a bug and should be handled on the server end. In this case I think the main issue is caching. There should be a way to bypass any server-side caching and get the current state of the observation, but that doesn't appear to be working in APIv2. I'll fix that so when a client wants the current state of the observation they can use `ttl=-1` as currently works in APIv1

Issue #3634, opened 2023-01-28
Add mute button to user profiles


kueda:
Blocking is often used when muting would suffice, which is not good because blocking can be abused as a way to avoid identifications from a particular person, when we would prefer that people deal with identifiers they don't agree with engaging those people in discussion to understand their reasoning, and/or by opting out of the Community Taxon.

As such, we'd like to make muting a more obvious recourse for dealing with people you find annoying, and one way to do that is to put it on the user profiles.

- [x] Put a "Mute" button on the user profiles to the right of the "Follow" button. IMO, it should be a btn-default with a [`fa-microphone-slash`](https://fontawesome.com/v4/icon/microphone-slash) icon, though we might have to adjust the design a bit based on feedback
- [x] It should only be visible to signed in users
- [x] Clicking it should open a confirmation dialog with the title "Are you sure?" and the body "Muting someone prevents you from being notified about any of their activity, including "@" mentions, messages, or comments and identifications on your observations." There should be a "Cancel" button to close the dialog, and a "Mute" button that requests [POST /user_mutes](https://github.com/inaturalist/inaturalist/blob/main/app/controllers/user_mutes_controller.rb#L5) a body like `user_mute["muted_user_id"]=1234` where 1234 is the ID of the user being muted. You can check out the network requests in user settings for a rough idea of how this works with the API, but since user profiles are just plain Rails pages, I think `form_for` or `button_for` and the Rails endpoint will suffice.
- [x] If you've already muted that user, it should be an "Unmute" button
- [x] Clicking "Unmute" should [DELETE /user_mutes/123](https://github.com/inaturalist/inaturalist/blob/main/app/controllers/user_mutes_controller.rb#L29) where 123 is the ID of the relevant UserMute record (not the ID of the user)
- [x] It would be nice if this worked without reloading the page like the follow button does, but not a hard requirement

PR #3642, opened 2023-02-07
Add mute/unmute button to profile header

rainhead:
Implements https://github.com/inaturalist/inaturalist/issues/3634.


kueda:
@rainhead sorry for the miscommunication, but @pleary actually worked on this yesterday and today and has it working so I'm going to close this. We'll try and be better about self-assigning issues in the future so it's clear that someone's working on it. If you want to work on an issue, please leave a comment on it so we can assign you.

Issue #3639, opened 2023-02-01
Observations export preview fails with list_id and projects param


kueda:
**Describe the bug**
Visiting observations/export with both the `projects` and `list_id` params causes a 500 error when loading the preview.

**To Reproduce**
1. Go to https://www.inaturalist.org/observations/export?verifiable=true&page=1&spam=false&list_id=4381953&place_id=any&projects=ebcnps
1. The preview never loads
1. The console shows a 500 error when requesting the preview contents

**Expected behavior**
The `list_id` and `projects` params should both work together.

**Context**
 - OS: MacOS 12.5.1
 - Browser: Firefox 109.0

**Additional context**
This might just be something unique to this specific project. Also, FWIW, I created the project and am the admin, but someone else created the list.

PR #3647, opened 2023-02-11
Allow taxon_ids to be ints or strings

rainhead:
Fixes: https://github.com/inaturalist/inaturalist/issues/3639

I wrote a test for this, but without clear type expectations around params/search_params/etc, the test had to be at the controller level, and I ran out of energy figuring out how to create the fixtures and stubs I needed there. The test did run far enough to exhibit the issue and confirm the fix, as did manual testing.

### DIFF
--- a/lib/observation_search.rb
+++ b/lib/observation_search.rb
@@ -140,7 +140,7 @@ module ObservationSearch
         api_params.delete(:observations_taxon_ids)
       end
       unless api_params[:taxon_ids].blank?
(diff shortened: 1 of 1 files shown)


rainhead:
@kueda if you'd like, I can work to make the various `params` types here clear and consistent.

kueda (approved):
Yeah, I took a stab at writing a test and realized I needed to stub `INatAPIService` and gave up too. Should be doable, but maybe not worth it here.

PR #3655, opened 2023-02-17
Tool to import a place from the Production API

rainhead:
Based on a script by @kueda, but:

- imports recursively, recording parent
- records uuid
- uses Faraday, so we can switch to v2 API + JWT token eventually

One trick with this is that Asia cannot be imported because something is amiss with its geometry. Because the script imports a place's parents recursively, nothing in Asia can be imported. Do you have a hunch for whether this is a problem with production data or with the import script?

### DIFF
--- a/tools/import_place_from_production.rb
+++ b/tools/import_place_from_production.rb
@@ -1,0 +1,74 @@
+# frozen_string_literal: true
+
+if Rails.env.production?
+  puts "This script is for setting up a dev environment only."
+  exit 0
+end
+
+module ImportPlaceFromProduction
+  API = Faraday.new( "https://api.inaturalist.org" ) do | faraday |
+    faraday.response :raise_error
+  end
+
+  def self.fetch_place( id_or_slug )
+    http_response = API.get( "/v1/places/#{id_or_slug.to_s.parameterize}" )
+    response = JSON.parse( http_response.body )
+
+    unless response["results"].one?
+      raise "Expected exactly one result from API, got #{response['total_results']}"
+    end
+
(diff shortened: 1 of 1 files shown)


kueda:
@rainhead yeah, unfortunately Asia has a geometry that is topologically valid in the database but doesn't pass our application-level validation test of not having impossible coordinates. The best fix would be to change the record in the database, but... that would kick off a cascade of re-indexing jobs that I don't think is warranted here.

I think this is fine to merge as-is, but if you're interested, one potential workaround is to have the script review every point in the new geometry and snap impossible values (e.g. longitude of -180.1) to possible ones (e.g. longitude of -180.0).

tools/import_place_from_production.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+if Rails.env.production?
```
todtb, line 3: `unless Rails.env.development?` ?

todtb (commented):
Very helpful, thanks!

tools/import_place_from_production.rb
kueda: Fair. No reason to be running this in a test environment.

rainhead: Sure, this was just copy-pasted from another file in the directory.

rainhead: Addressed in [bc29c2f](https://github.com/inaturalist/inaturalist/pull/3655/commits/bc29c2fe096d0d50047db774aa00685014d4cf20).

PR #3665, opened 2023-02-22
Add TaxonImporter tool, apiv2 client

todtb:
- Adds `TaxonImporter` for pulling production taxa.

### DIFF
--- a/lib/inat_api_service/v2/client.rb
+++ b/lib/inat_api_service/v2/client.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+module INatAPIService
+  module V2
+    class Error < StandardError; end
+
+    class Client
+      BASE_URL = "https://api.inaturalist.org/v2"
+
+      attr_reader :api
+
+      def initialize
+        @api = Faraday.new BASE_URL
+      end
+
+      # GETs taxon by id
+      #
+      # @param [String] id Taxon id(s). Multiple ids separated by comma
+      # @param [String] fields Fields to include in response
+      # @return [Hash] parsed JSON response
+      def get_taxon_by_id( id, fields: "all" )
+        get "taxa/#{id}?#{fields_from_opts fields}"
+      end
+
+      private
+
+      def fields_from_opts( fields )
+        return "" unless fields.present?
+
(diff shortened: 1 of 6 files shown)


lib/inat_api_service/v2/client.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+module INatAPIService
+  module V2
+    class Error < StandardError; end
```
todtb, line 5: It didn't seem like this existed yet so let me know if:
1) You'd like to approach this in a different pattern or it isn't following how v2 should be consumed
2) You'd like to isolate to tooling to prevent premature use

spec/lib/inat_api_service/v2/client_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      it "requests only specified fields" do
+        expect( a_request( :get, url ).with( query: { fields: "id,name" } ) ).to have_been_made
+      end
+    end
+  end
+
+  describe "parsing" do
+    subject { described_class.new.get_taxon_by_id 123 }
+    let( :url ) { "#{super()}/taxa/123" }
+    let( :body ) { File.read "spec/lib/inat_api_service/stubs/get_taxon_by_id.json" }
```
todtb, line 36: `VCR` can be handy for things like this, though probably a bit much for a one-off

tools/development_tools/taxon_importer/taxon_importer.rb
```diff
@@ -0,0 +1,72 @@
...
+    def import( taxon_id: )
+      new( taxon_id: taxon_id ).import
+    end
+  end
+
+  def initialize( taxon_id: )
+    @taxon_id = taxon_id
+  end
+
+  # Import will 'De-Limpkinize'
```
kueda, line 20: Haha, love this. Delimpkinize shall be the name of the Limp Bizkit cover band that will never form.

kueda (approved):
Cool! Some other ideas:

* option to include photos
* option to include names
* option to include statuses

Might also be worth changing the dev setup instructions to seed the initial db using this script.

Issue #3543, opened 2022-10-28
Allow infraspecific IDs to move the obs taxon to species


kueda:
See https://forum.inaturalist.org/t/leading-subspecies-ids-should-change-the-obs-taxon-like-leading-ids-of-other-ranks/139 and the 9/14 Support Report for context.

Infraspecific IDs already change the obs taxon if they are the only ID, so that doesn't need to change. We also don't want to change the way the Community Taxon is determined. We just want to deal with the situation where an infraspecific ID is the only leading ID, and in that case change the obs taxon to the *species* of the infraspecific taxon.

I believe the relevant code is around https://github.com/inaturalist/inaturalist/blob/main/app/models/observation.rb#L1918

I don't know if it's worth updating all past observations or not. I suppose most people will think it will, though that makes this quite a bit harder.

pleary:
This should be addressed with the linked commit. After this is deployed we can retroactively update observations that would have used this behavior. Observations affected should be a subset of those returned by this query `Observation.joins( :taxon, { identifications: :taxon } ).where( "taxa.rank_level > ? AND taxa_identifications.rank_level < ?", Taxon::SPECIES_LEVEL, Taxon::SPECIES_LEVEL )`, so we can run `set_taxon_from_probable_taxon` on everything that returns

PR #3668, opened 2023-02-23
Infraspecific ID behavior change

pleary:
Infraspecific IDs move the obs taxon to species #3543


kueda (approved):
As far as I can tell this is working to spec.

PR #3679, opened 2023-03-04
Change place size messaging and implementation to use square km

kueda:
Using square degrees means the restriction varies based on latitude, and it means we can't reasonably communicate the way the restriction works to users because no one knows what a square degree is. This changes most uses of the restriction to use square kilometers.

### DIFF
--- a/app/controllers/places_controller.rb
+++ b/app/controllers/places_controller.rb
@@ -258,19 +258,29 @@ class PlacesController < ApplicationController
     end
 
     if !current_user.is_admin? &&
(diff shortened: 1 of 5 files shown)


kueda:
Unless there's a good way to calculate area in km2 in RGeo, I think this probably just makes things more confusing by creating two separate ways we measure area, one for new uploads and another for places already in the database.

kueda:
I don't really see a way forward here. We need one way to calculate area, preferably *before* the places gets added to the database, and I don't think we have one.

PR #3690, opened 2023-03-14
Enable a content security policy implementing a frame_ancestors policy

kueda:


app/controllers/application_controller.rb
```diff
@@ -733,8 +733,15 @@ def self.accept_formats(*args)
...
-  def allow_external_iframes
-    response.headers["X-Frame-Options"] = "ALLOWALL"
+  def self.allow_external_iframes( options )
+    # modify the content security policy for the actions specified in options to
+    # disable the frame_ancestors policy, thus allowing iframes to request these actions
+    content_security_policy( **options ) do |p|
+      p.frame_ancestors nil
+    end
+    before_action( options ) do
+      response.headers["X-Frame-Options"] = "ALLOWALL"
```
kueda, line 743: According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options, `Content-Security-Policy` should make this unnecessary, right? Either way, `ALLOWALL` doesn't seem to be a valid value, so we should probably modify this if we're not going to remove it.

kueda (commented):
Didn't do a deep dive but it's setting the header and doesn't seem to break our own use of frames.

app/controllers/application_controller.rb
pleary: I updated the content security policy for these pages to "frame_ancestors 'https:'" which should allow these pages to be embedded as iframes on sites that implement https. I then removed references to `X-Frame-Options` as that header should no longer be needed

Issue #3636, opened 2023-01-28
Observations map missing markers or zoom/pan incomplete


synrg:
[//]: # (For feature requests and bug reports not specifically concerning code, please use https://forum.inaturalist.org/c/feature-requests/16. iNat staff will make github issues if/when an actionable solution emerges)

**Describe the bug**
When displaying an observations map, it sometimes fails to set up completely. This is often noticed as the markers being completely absent from the map, but also can result in the zoom/pan not finishing, leaving the view window incorrectly positioned.

This has been noted and discussed before in these threads:

https://forum.inaturalist.org/t/how-to-refresh-map-to-show-observations/25339/17
https://forum.inaturalist.org/t/empty-explore-map/28400
https://forum.inaturalist.org/t/observations-map-doesnt-reliably-zoom-to-place-so-markers-can-be-seen/38393

I started a review of https://github.com/inaturalist/inaturalist/blob/main/app/assets/javascripts/ang/controllers/observation_search.js.erb and am suspicious of these 300ms and 500ms timeouts introduced in https://github.com/inaturalist/inaturalist/commit/053fed356d266ac83982661f3c8bf12106100d0e

```js
    // waiting a bit after creating the map to initialize the layers
    // to avoid issues with map aligning, letting the browser catch up
    setTimeout( function( ) {
      if( !$scope.mapLayersInitialized ) {
        $scope.setMapLayers( true );
        // more delays before enabling onMoves so setMapLayers
        // can finish aligning the map if it needs to
        setTimeout( function( ) {
          $scope.map.addListener( "dragstart", function( ) { $scope.delayedOnMove( ); });
          $scope.map.addListener( "center_changed", function( ) { $scope.delayedOnMove( ); });
          $scope.map.addListener( "zoom_changed", function( ) { $scope.delayedOnMove( ); });
        }, 500 );
      }
    }, 300);
```

Since then, that `setupMap()` function has grown to be a bit more complex, now including the setup of rectangular and circle map tools. I'm unsure whether or not that may exacerbate the problem.

I'm not even entirely sure it's either of these timeouts that causes the issue, but setting timers vs. responding to an event that is triggered upon completion of the map render seems dodgy at best. Perhaps the third-party map API provides some means to pass a callback when the whole map is finally set up and ready that could be used instead? Better yet, if it returned a promise we could await it ...

I didn't have time to set up a full test environment to investigate further, nor to dig into the Google maps API.

**To Reproduce**

[...]

PR #3691, opened 2023-03-14
Upgrade Google Maps API

rainhead:
In service of #3636.

Changes:
- Upgrade Google Maps to from 3.48 to 3.51 ([ChangeLog](https://developers.google.com/maps/documentation/javascript/releases#3.48.1)). **NB: 3.48.1 removes support for Internet Explorer**.
- Use the same Maps API version regardless of environment.
- Upgrade type signatures package, whose name changed slightly.
- Do less raw string manipulation in the helper.
- Always load the Maps API over a secure connection.
- Fix a setting in `config.yml.example` that tripped me up when setting up my development environment.

### Async Loading

While Google [says](https://developers.google.com/maps/documentation/javascript/overview#sync) you can load the Maps API synchronously, this generates a warning in the console, which is visible in production. It's possible that Google would remove support for synchronous loading in the future.

I implemented a proof of concept of loading the Maps API asynchronously and triggering the initialization in `map3.js.erb` via a `Promise`. That change worked okay for me, but would probably require lots of regression testing.

### Next Steps

#3636 reads to me as a roundup of related but separate race conditions. @synrg is probably right to finger the timeouts in `observation_search.js.erb`. It would not be hard to weave a `Promise` through the setup functions, and probably make the overall setup process snappier as a result. Maps increasingly provides `Promise`s for the asynchronous parts of its API, but in this case, I think it's our own setup process we need to wait on. I can do this in my next PR.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1372,13 +1372,18 @@ module ApplicationHelper
   end
 
   def google_maps_js(libraries: [])
-    source = URI::HTTPS.build host: "maps.google.com", path: "/maps/api/js", query: {
-      callback: "GOOGLE_MAPS_CALLBACK",
+    javascript_include_tag google_maps_loader_uri(libraries: libraries)
+  end
+
+  # https://developers.google.com/maps/documentation/javascript/url-params
+  # https://developers.google.com/maps/documentation/javascript/libraries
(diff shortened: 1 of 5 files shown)


rainhead:
I'd appreciate some help with manual testing.

PR #3705, opened 2023-03-23
Fix load time zone geometries

rainhead:
### Context

https://github.com/inaturalist/inaturalist/wiki/Development-Setup-Guide gives instructions for loading timezone geometries from a shapefile. However, `tools/load_time_zone_geometries.rb` expects a geojson file, not a shapefile.

### Changes

* Allow the tool to import shapefiles in addition to geogjson files.
* Fix the connection string fed to `ogr2ogr` to include the port number

### DIFF
--- a/app/models/time_zone_geometry.rb
+++ b/app/models/time_zone_geometry.rb
@@ -22,6 +22,7 @@ class TimeZoneGeometry < ActiveRecord::Base
       pg_string = {
         dbname: ApplicationRecord.connection_db_config.configuration_hash[:database],
         host: ApplicationRecord.connection_db_config.configuration_hash[:host],
+        port: ApplicationRecord.connection_db_config.configuration_hash[:port],
         user: ApplicationRecord.connection_db_config.configuration_hash[:username],
         password: ApplicationRecord.connection_db_config.configuration_hash[:password],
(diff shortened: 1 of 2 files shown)


app/models/time_zone_geometry.rb
```diff
@@ -73,6 +74,7 @@ def load_shapefile( shp_path, options = {} )
       pg_string = {
         dbname: ApplicationRecord.connection_db_config.configuration_hash[:database],
         host: ApplicationRecord.connection_db_config.configuration_hash[:host],
+        port: ApplicationRecord.connection_db_config.configuration_hash[:port],
```
kueda, line 77: This will cause errors if the port is *not* specified in the `configuration_hash`, and it doesn't need to be if the default will work, so please make this work in case this value is blank (same applies to line 25).

kueda (changes requested):
Works for me, just the one change, please.

app/models/time_zone_geometry.rb
rainhead: Thanks, addressed in [c781dbc](https://github.com/inaturalist/inaturalist/pull/3705/commits/c781dbcb221bc402ea6991e8e0e13b8d848d91b6).

kueda (approved):

Issue #3343, opened 2021-12-28
Clicking on a taxon between genus and species (e.g. subgenus) loads the about tab, not taxonomy tab


tiwane:
From https://forum.inaturalist.org/t/clicking-on-a-taxon-between-genus-and-species-e-g-subgenus-loads-the-about-tab-not-taxonomy-tab/5077 I can replicate.

1. Go to https://www.inaturalist.org/taxa/121922-Eublemma
2. Go to Taxonomy tab
3. Click on a subgenus (eg Subgenus Odice).
4. That subgenus's taxon page loads, but the About tab is shown, not the Taxonomy tab.

Expected behavior: the subgenus page shows the Taxonomy tab, as it does if you click on a species or a coarser taxon.

Mayuresh0072:
@kueda Hey, Can you please assign me this issue ? Thank You.

Mayuresh0072:
@tiwane  can you tell for which specific ranks (rank levels to be specific) of taxon this bug is occurring ?

Mayuresh0072:
Hello @kueda I have raised a PR for this issue. Please check and approve the changes and workflow so that it can be merged into main branch. Thanks.

PR #3724, opened 2023-04-06
clicking on subgenus should open taxonomy tab

Mayuresh0072:
When we click on taxon of rank genus and species, About tab was loading instead of Taxonomy Tab.

### DIFF
--- a/app/webpack/taxa/show/components/taxon_page_tabs.jsx
+++ b/app/webpack/taxa/show/components/taxon_page_tabs.jsx
@@ -305,7 +305,7 @@ class TaxonPageTabs extends React.Component {
             />
           </div>
           <div
-            role="tabpanel"
+            role="tabpanel" 
             className={`tab-pane ${genusOrSpecies ? "" : "hidden"} ${chosenTab === "similar" ? "active" : ""}`}
             id="similar-tab"
           >
@@ -327,7 +327,7 @@ TaxonPageTabs.propTypes = {
 };
 
 TaxonPageTabs.defaultProps = {
-  chosenTab: "articles"
+  chosenTab: "taxonomy"
 };
 
 export default TaxonPageTabs;


kueda (changes requested):
Good start, @Mayuresh0072! However, you need to fix the issue in #3343 without changing the existing default tab. I think the source of the problem is around https://github.com/inaturalist/inaturalist/blob/main/app/webpack/taxa/show/containers/taxon_page_tabs_container.js#L22, so have a look at that and make sure it handles a subgenus the same way it handles a genus.

Also, in the future please name issue-specific branches starting with the issue number as we've [requested](https://github.com/inaturalist/inaturalist/blob/main/CONTRIBUTING.md#working-on-an-issue). In this case that would be something like `3343-subgenus-taxonomy-tab`. There's no need to change this branch, just something to keep in mind if you decide to work on other issues.

Mayuresh0072:
@kueda I have made the change based on your feedback. Please have a look at it. Thanks.

Mayuresh0072:
Hello @kueda , I have done code changes as per your feedback. Please check it. Thanks.

kueda (changes requested):
Works, but I'd like a few more changes.

Mayuresh0072:
Hello @kueda , I have removed 'BetweenGenusAndSpecies' as it was redundant. But it should behave as 'aboveGenusTabs' and not as 'genusTabs' because 'genusTabs' have extra tab (similar taxa tab). So, I have changed the condition in 'aboveGenusTabs' which is working perfectly. Second thing, When I ran the command - npx eslint --fix app/webpack/taxa/show/containers/taxon_page_tabs_container.js . It changed the package.json and package-lock.json. I dont know why this is happening. So, I corrected both the files by replacing from the main branch. Sorry for creating more commits , but all changes are correct now in this branch. Please have a look at it. Thanks.

kueda (approved):
Thanks, @Mayuresh0072, looking good. We'll probably merge and deploy this next week.

PR #3729, opened 2023-04-11
Ensure local Project model extends inatjs model; use instanceof

### LINKED ISSUE #3728: Flagging collection or umbrella project flags comment with the same ID number instead
From https://forum.inaturalist.org/t/attempting-to-flag-a-project-results-in-the-comment-with-the-same-id-number-being-flagged-instead/40770 I can replicate.

1. Go to a collection or umbrella project.
2. Flag it. Any reason works.
3. Nothing happens on the project page. 
4. Go to the comment that shares the same ID number as the project you flagged. Your flag is attached to that comment.

kueda:
Closes #3728

### DIFF
--- a/app/webpack/observations/show/components/flagging_modal.jsx
+++ b/app/webpack/observations/show/components/flagging_modal.jsx
@@ -7,7 +7,8 @@ import {
   Observation,
   Photo,
   Project,
-  Sound
+  Sound,
+  Comment
 } from "inaturalistjs";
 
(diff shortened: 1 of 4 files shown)


pleary:
inatjs wasn't turning comments or sounds into instances, and there wasn't even a sound model. I updated inatjs to add a sound model and turn them both into instances within observation instances. I also made sure Comment was being imported in flagging_modal. I think things are working as expected now for all places we use flagging_modal

Issue #3700, opened 2023-03-16
Users without organizer privileges can get to place creation page


tiwane:
1. Use an account that lacks organizer privileges.
2. Go to inaturalist.org/places and search for a place at "Find a place".
3. On the results page, you'll see this on the upper right-hand side:

<img width="283" alt="image" src="https://user-images.githubusercontent.com/15896181/225747676-ba10c4a7-dbed-4073-a2fc-3445e3bc62b5.png">

4. That link takes you to the place creation page, but if you try to save your new place you'll hit and error that says you don't have organizer privileges.

Since users without organizer privileges can't actually make a new place, we should remove that link for non-organizer users. 

TODO

- [x] Remove "Not finding what you want? [Add a New Place](https://www.inaturalist.org/places/new)" text and link from `/places/search` if they do not have the organizer privilege
- [ ] `PlacesController#new` should redirect the user back if they don't have the organizer privilege (I think we just need to add `:new` to `requires_privilege :organizer, only: [:new, :create, :edit, :update, :destroy]`)

Mayuresh0072:
Hello @kueda , Please assign me this issue . Thanks.

Mayuresh0072:
@pleary @kueda As we have discussed in the slack, I think 2nd todo task is not needed, as it is already redirecting it to current page if user manually types URL and go to place creation page. There is no way user is goes to place creation page.  Just the thing is we need to correctly display the error message if user manually tries to open place creation link. Just a suggestion,  I think we should create a different issue for this task.

PR #3733, opened 2023-04-15
Only users with organizer privilege can see place creation

Mayuresh0072:
This fixes the issue #3700 .

### DIFF
--- a/app/controllers/places_controller.rb
+++ b/app/controllers/places_controller.rb
@@ -51,17 +51,9 @@ class PlacesController < ApplicationController
 
   before_action :allow_external_iframes, only: [:guide_widget, :cached_guide]
 
-  requires_privilege :organizer, only: [:new, :create, :edit, :update, :destroy],
-    if: proc {| _c |
-      # Only check privileges if the current user didn't create the place, i.e.
-      # allow people to edit places they created even if they haven't earned the
-      # organizer privilege
-      if @place.blank?
-        true
-      else
-        current_user.id != @place.user_id
-      end
-    }
+  requires_privilege :organizer, only: [:new, :create, :edit, :update, :destroy] do
+    redirect_back_or_default(root_url)
+  end
   protect_from_forgery with: :exception, unless: lambda {
     request.parameters[:action] == "autocomplete" && request.format.json?
   }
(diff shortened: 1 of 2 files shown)


Mayuresh0072:
Hello @kueda , I have made code changes for solving the issue #3700  .Please have a look at it . Thanks.

kueda (changes requested):
There is no `:organizer` role so this condition will always be false and no one will be able to access this link, which doesn't address the first todo in https://github.com/inaturalist/inaturalist/issues/3700. "Organizer" is a `UserPrivilege`, not a `Role`. You can control what users have what privileges using the admin interface in your local environment.

You also need to address the second todo in https://github.com/inaturalist/inaturalist/issues/3700

app/controllers/places_controller.rb
```diff
@@ -51,17 +51,9 @@ class PlacesController < ApplicationController
...
-      # organizer privilege
-      if @place.blank?
-        true
-      else
-        current_user.id != @place.user_id
-      end
-    }
+  requires_privilege :organizer, only: [:new, :create, :edit, :update, :destroy] do
+    redirect_back_or_default(root_url)
+  end
```
kueda, line 56: Why are you removing the ability to edit a place you created regardless of privilege? Also, requires_privilege does not take a block. You can find the full definition at https://github.com/inaturalist/inaturalist/blob/main/lib/privileges.rb#L63

kueda (changes requested):

app/controllers/places_controller.rb
Mayuresh0072: @kueda Actually I thought the if block checks whether other users who have not created the place have the edit access. So, now i will keep the 'if block' as it is. But what exactly I have to do for the 2nd Todo task.Can You please explain?

Mayuresh0072: @kueda So, I think I should only add ':new' and ':create' in the 'only:' array of the organizer privilege.

kueda (approved):

pleary:
I merged the PR as it stands, which addresses the first TODO in #3700. As @kueda noted, the second TODO remains unresolved so far

PR #3739, opened 2023-04-21
Migrate postgresql queries to elasticsearch queries

sylvain-morin:


app/models/site_statistic.rb
```diff
@@ -203,67 +337,206 @@ def self.users_stats(at_time = Time.now)
...
+              }
+            }
+          }
+        }
+      }
+    ).response.aggregations.group_by_sites.
+      buckets.map do |b| 
+          if sites.key?(b["key"])
+            site_name = sites[b["key"]]
+            [ site_name, b["doc_count"] ]
```
pleary, line 506: This is using `doc_count` and not the value of the `distinct_taxon_id` aggregation, so it reflects observation counts and not distinct taxon counts

app/models/site_statistic.rb
```diff
@@ -203,67 +337,206 @@ def self.users_stats(at_time = Time.now)
...
+          range: {
+            "taxon.rank_level": {
+              gt: 0
+            }
+          }
+        }
+      ],
+      aggregate: {
+        ranks: {
+          terms: { field: "taxon.rank", size: 1000 }
```
pleary, line 530: The original was grouping by `rank_level` and not `rank`. `rank` is probably a more interesting and definitely more precise metric, but it will alter the stats a bit. For example what we have been calling "subspecies" includes varieties and forms, and with this approach they would each be separate values. We should make sure we're OK with the change since it will alter the chart on the stats page

app/models/site_statistic.rb
```diff
@@ -120,7 +120,31 @@ def self.observations_stats(at_time = Time.now)
...
             }
-          },
+          }
+        },
+        {
+          exists: { field: "community_taxon_id" }
+        }
+      ]
+    ).total_entries
+    community_identified_to_genus = Observation.elastic_search(
```
pleary, line 130: This count doesn't quite match up to the original, and it looks like we don't have the data in Elasticsearch to get the same count. This query returns the count of observations that have a community_taxon_id and where the taxon is rank genus or below - not where the `community_taxon` is rank genus or below as observation.community_taxon isn't necessarily the same as observation.taxon. We don't store the community_taxon rank in elasticsearch. This ends up being few percent higher than the original value.

I wonder if it's worth keeping the old long-to-calculate stats, or maybe this is even a stat we could get rid of entirely. Alternatively we could do something more complicated like fetching all the unique community_taxon_ids from ES, then querying the DB to figure out which ones are genera or below

pleary (changes requested):
This is mostly looking pretty good. The site species count should be fixed, and I left a few other smaller things to consider. There's also one failing spec that should be addressed. In my testing the IDs and projects numbers were extremely close, and the others were very close to the originals, sometimes a percent or less off. It may be that the obs and users indices have a small number of documents out of sync leading to the difference.

It would be interesting to try running these modified queries in production around the same time we generate site stats with the older queries to get a closer comparison at nearly the same time on nearly the same data.

Overall though it's looking really good and the time savings is huge. This will be a nice improvement over the current queries.

app/models/site_statistic.rb
sylvain-morin: Fixed

app/models/site_statistic.rb
sylvain-morin: Fixed, same value as original (returning rank_level from ES and replacing labels using Taxon::RANK_FROM_RANK_LEVEL)

sylvain-morin:
I replaced the last DB call for taxa_stats, related to "active" users.
This call was responsible for 2h45min of computation on smilax.
It was extracting the lists of distinct user IDs from the observations, identifications, posts and comments created during the last 30 days ==> very large lists (around 250K each)

In Elastic Search, I can do the same for observations and identifications ONLY (no posts or comments indices)
The computation on smilax is just a few seconds, but may be quite heavy (returning two 250k items aggregations)

The result is quite similar, even if we are not considering posts or comments anymore:
:active=>249723,		⇒ -980 (-0.39%)

sylvain-morin:
Here are my full results, executed on smilax
https://docs.google.com/document/d/1el8hocFtxkJqXQ0vPnl46HAfV6clHMQVgTsOqEBa6cs/edit?usp=sharing

app/models/site_statistic.rb
```diff
@@ -141,45 +197,179 @@ def self.observations_stats(at_time = Time.now)
...
+      filters: [
+        {
+          range: {
+            created_at: {
+              lte: at_time
+            }
+          }
+        },
+        { term: { spam: false } },
+        { term: { roles: "curator" } }
```
pleary, line 276: To match the original count we should include the admin role as well. So we should make this a `terms` filter that includes both curator and admin

sylvain-morin: Done

app/models/site_statistic.rb
sylvain-morin: Community taxon has been fixed

PR #3745, opened 2023-05-03
Listed taxa query fix

pleary:
Listed taxa without places have no primary_listed_taxa or related_list, so don't attempt to query for them with a nil place_id


kueda (approved):
Looks good to me. None of these methods seem relevant without a `place_id`.

PR #3753, opened 2023-05-17
3752 CID opt-in fix

### LINKED ISSUE #3752: Can't add disagreeing ID to obseration if user opted in to community ID
From https://forum.inaturalist.org/t/cant-disagree-with-one-specific-observation/41316 

The observation in question is [this one](https://www.inaturalist.org/observations/37868526). The observer opts out of CID by default, but opted in for this observation. The obs taxon is _Ramaria formosa_, but the CID is _Ramaria_. If you add an ID of _Ramaria_, you are not shown the disagreement modal.

I was able to replicate this on staging with a test account that opts out of CID by default but opted in for an observation. 

Expected behavior: when adding an ancestor ID of the obs taxon, I should see a disagreement modal in cases like this.

jwidness:
Resolves #3752.


kueda (changes requested):
Sorry it's taken so long to get to this, @jwidness. I think this works, but two requests:

- [ ] Please make sure you're following our eslint rules for line length and wrapping
- [ ] `!o.preferences.prefers_community_taxon === true` is just really hard to read. Please use `o.preferences.prefers_community_taxon === false` instead

jwidness:
Using `o.preferences.prefers_community_taxon === false` would cause it to mishandle cases where the value is `null`.

kueda:
@jwidness, how about `o.preferences.prefers_community_taxon !== true` or `o.preferences.prefers_community_taxon === false || o.preferences.prefers_community_taxon === null`?

jwidness:
I implemented a different option in the commit, but if you would prefer yours, I can change it.

For reference, this is the table I'm working off of:

image

app/webpack/observations/show/ducks/observation.js
```diff
@@ -796,7 +796,8 @@ export function addID( taxon, options = { } ) {
     let observationTaxon = o.taxon;
     if (
       o.preferences.prefers_community_taxon === false
-      || (o.user.preferences.prefers_community_taxa === false && !o.preferences.prefers_community_taxon === true)
+      || (o.user.preferences.prefers_community_taxa === false 
+      && o.preferences.prefers_community_taxon === null)
```
jwidness, line 800: Sorry for the ambiguity, I was talking about this second commit, where I think I addressed both your original comments.

kueda: Ah, the fault is mine. Sorry for not seeing that.

kueda (approved):

Issue #3713, opened 2023-03-30
Prevent certain numbers from being used in the usernames of third-party generated accounts


tiwane:
If someone makes an account via a third party, we often add numbers as part of their username. One user thought that a number, which is associated with white supremacy according to the ADL and SPLC, was a dog whistle by the user. 

I propose we prevent certain numbers from being used when [automatically generating usernames](https://github.com/inaturalist/inaturalist/blob/main/app/models/user.rb#L886) for accounts created via a third party. Here's a list sourced from the ADL: https://www.adl.org/resources/hate-symbols/search?f%5B0%5D=topic%3A1709

100
109
110
111
12
13
1352
1390
52
90
14
1410
1423
1488
18
21212
2316
28
23
16
311
318
336
38
43
511
737
83
88
9 
10

PR #3761, opened 2023-05-25
Don't suggest logins containing sensitive numbers

rainhead:
Fixes: #3713 

Do you want to include 666, 69, 420, or anything else in here?

- Clean up code. Remove all conditionals. You can read what happens in a straight line from top to bottom. Variables have meaningful names. The comments are hopefully helpful.
- The same logic is applied to `DEFAULT_LOGIN` as to requested logins. We try the base login, then try adding a single digit (except 9), then try twenty psuedorandom numbers.
- Random-number suffixes are limited at half of `MAX_LOGIN_SIZE` rather than a number derived from the storage of the `id` field.
- It will fail rather than trying logins indefinitely.

This could be done in a single SQL query, but I'd have to write custom Arel, which didn't seem worth it.

This method should perhaps raise an error rather than returning nil, but that would change its contract.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -279,7 +279,14 @@ class User < ApplicationRecord
   
   MIN_LOGIN_SIZE = 3
   MAX_LOGIN_SIZE = 40
-  DEFAULT_LOGIN = "naturalist"
+  DEFAULT_LOGIN = "naturalist".freeze
+
+  # Do not append these numbers to usernames when generating an unused username for a new user.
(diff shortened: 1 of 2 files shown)


kueda (changes requested):
Sorry it's taken a while to get to this, @rainhead. This looks good to me. I was going to request a test but I guess with your random approach that's impossible. One concern: the random suffixes are very big, e.g. `naturalist30332186623637276288`. Could we maybe cut those down to 5 digits?

kueda (approved):

Issue #3746, opened 2023-05-03
Wrong iconic taxon shown in search results for adding taxa to list


tiwane:
From https://forum.inaturalist.org/t/taxa-that-dont-belong-to-an-iconic-taxon-inherit-their-iconic-taxon-from-previous-taxa-listed-when-searching-for-taxa-to-add-to-a-list/41300 I can replicate.

1. Open up a list
2. Search for taxa to add and write "severe" in the form. 
3. Results show things like viruses with bird icons next to them

<img width="264" alt="image" src="https://user-images.githubusercontent.com/15896181/236051705-569e6987-6619-457b-851c-dbb01c699578.png">

Expected: taxa not in an iconic taxon should have no icon next to them

PR #3778, opened 2023-06-13
Don't mutate options when calling extend

rainhead:
Fixes: #3746.

Please try this out locally. It wasn't easy to figure out how to set up test data.

Before:
Screenshot 2023-06-13 at 1 36 30 PM

After:
Screenshot 2023-06-13 at 1 36 08 PM

In some places we're good about `extend`ing into a new object, in others we're not.

### DIFF
--- a/app/assets/javascripts/jquery/plugins/inat/taxon_selectors.js.erb
+++ b/app/assets/javascripts/jquery/plugins/inat/taxon_selectors.js.erb
@@ -97,6 +97,7 @@
     var status = $('<div class="status">' + I18n.t('species_unknown') + '</div>');
     $(status).css(
       $.extend(
+        {},
         $.fn.simpleTaxonSelector.styles.statuses['default'],
         $.fn.simpleTaxonSelector.styles.statuses.unmatched
       )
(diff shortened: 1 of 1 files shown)


kueda (approved):

app/assets/javascripts/jquery/plugins/inat/taxon_selectors.js.erb
```diff
@@ -97,6 +97,7 @@
     var status = $('<div class="status">' + I18n.t('species_unknown') + '</div>');
     $(status).css(
       $.extend(
+        {},
         $.fn.simpleTaxonSelector.styles.statuses['default'],
         $.fn.simpleTaxonSelector.styles.statuses.unmatched
       )
```
frewsxcv, line 103: In case you don't already know, you can do this instead without needing lodash/underscore:

```js
$(status).css({
    ...$.fn.simpleTaxonSelector.styles.statuses['default'],
    ...$.fn.simpleTaxonSelector.styles.statuses.unmatched
})
```

rainhead: Thanks for reaching out. There's quite a lot that could be improved with modern JS here, but I'm trying to stay consistent within a code unit, and to make minimal changes. Also, the most important calls to `extend` here perform deep merges, which I don't think I can do with plain JS.

Issue #3712, opened 2023-03-29
Include more new place restrictions on https://www.inaturalist.org/places/new


tiwane:
We receive emails from users who try to make a new place but their boundary contains too many observations and/or covers too large of an area. Those two restrictions are not publicly visible, so it leads to confusion and emails to help@inat and post on the forum.

I suggest we add the following two restrictions to the large yellow banner at https://www.inaturalist.org/places/new, under "Please keep in mind the following restrictions:"

- new places cannot contain over [200,000 observations](https://github.com/inaturalist/inaturalist/blob/main/app/models/place.rb#L205) ([10,000 during a content freeze](https://github.com/inaturalist/inaturalist/blob/main/app/models/place.rb#L206))
- new places cannot cover an area larger than X sq. km (about the size of Texas)

(I don't know the exact area limit)

rainhead:
Taking a look, but in a bit over my head.

PR #3785, opened 2023-06-21
Measure Places in square kilometers when applying policy limits

rainhead:
Fixes: #3712
Builds on: #3679

- Calculate size of places in square kilometers.
- Begin to factor usecase-sepecific policy expertise out of the models. Anything more than this seemed to disruptive.


app/controllers/places_controller.rb
```diff
@@ -221,11 +221,11 @@ def create
       assign_geometry_from_file
     elsif !params[:geojson].blank?
       @geometry = geometry_from_geojson( params[:geojson] )
-      @place.validate_with_geom( @geometry, user: current_user )
+      @place.validate_with_geom( @geometry, max_area_km2: max_area_km2, max_observation_count: max_observation_count )
     end
 
     if @geometry # && @place.valid?
-      @place.save_geom( @geometry, user: current_user )
+      @place.save_geom( @geometry, max_area_km2: max_area_km2, max_observation_count: max_observation_count )
```
kueda, line 228: I think this is the source of your spec failures, b/c the second hash arg of `save_geom()` is assumed to be extra attributes to be saved with the `PlaceGeometry` record. I can get specs to pass by removing those attributes from the has before [as we're doing with `user`](https://github.com/inaturalist/inaturalist/blob/main/app/models/place.rb#L474), but maybe it would be better to explicitly separate the extra `PlaceGeometry` attributes from options used by `save_geom()`, e.g. `def save_geom( geom, other_attrs = {}, options = {} )`.

kueda (changes requested):
This looks good to me except for the spec failures (left a comment about that inline). Once you fix that up I think this should be good to go. I like your way around my RGeo / PostGIS conundrum. I'm mildly worried about it getting stuck on a really big WKT creating a really big query, but hopefully file size limits will prevent that.

app/controllers/places_controller.rb
rainhead: Thanks. There was something funny with my test database preventing me from running many tests. Fixed now.

Let me know what you think of [f3bae47](https://github.com/inaturalist/inaturalist/pull/3785/commits/f3bae47d3d4b2f7ba922a258b33d3ca202714b85).

I think the root of the problem is that the responsibilities of data construction, normalization, validation, and fault-tolerant persistence are kind of muddled together in the Place model. That's pretty common for Rails code, and consistent with the rest of the code base, but I think refactoring this and related code would make it more concise and easier to understand.

app/models/place.rb
```diff
@@ -448,39 +439,40 @@ def validate_with_geom( geom, other_attrs = {} )
...
+        return false
+      end
+    end
+
     true
   end
   
   # Update the associated place_geometry or create a new one
-  def save_geom( geom, other_attrs = {} )
+  def save_geom( geom, max_area_km2: nil, max_observation_count: nil, **other_attrs )
```
kueda, line 463: I don't have a problem with this approach, but it's still causing tests to fail. They pass for me if I change this to

```ruby
def save_geom( geom, other_attrs = {}, max_area_km2: nil, max_observation_count: nil )
```

kueda (changes requested):

app/models/place.rb
rainhead: Thanks. Addressed in 40a7cdc47. I need to get less lazy about picking tests to run locally.

kueda (approved):

kueda:
Thanks, @rainhead!

PR #3789, opened 2023-06-27
Migrate photos between buckets based on license

sylvain-morin:


tools/migrate_photo_to_odp.rb
```diff
@@ -0,0 +1,20 @@
...
+while index < max_photo_id
+	LocalPhoto.joins("LEFT JOIN flags ON (photos.id = flags.flaggable_id)").
+  where("flags.id IS NULL").
+  where("photos.file_prefix_id=2")
+  where("photos.license not in (0,8,10)").
+  where("photos.id between ? and ?", index, index + batch_size -1).
+	find_each do |photo|
+	  LocalPhoto.delay(
+	      queue: "photos",
+	      unique_hash: { "LocalPhoto::migrate_open_photo_to_odp_bucket": photo.id }
```
pleary, line 16: this unique_hash should also be updated

tools/migrate_photo_from_odp.rb
```diff
@@ -0,0 +1,17 @@
+# Migrate all photos with restricted license (0) from the Open bucket to the Static bucket
+
+max_photo_id = LocalPhoto.last.id
+batch_size = 100000
+index = 0
+
+while index < max_photo_id
+	LocalPhoto.where( license: 0, file_prefix_id: 1 ).
```
pleary, line 8: Ideally this ID wouldn't be hard coded. Might be better to use `FilePrefix.where( prefix: "https://inaturalist-open-data.s3.amazonaws.com/photos" ).first.id`. You could also set that to a variable before the loop and use the assigned variable here. We could also use `Shared::LicenseModule::C` instead of 0, but looks like we don't actually use that anywhere else in the codebase though maybe we should

tools/migrate_photo_to_odp.rb
```diff
@@ -0,0 +1,20 @@
+# Migrate all photos with open licenses (not 0,8,10) from the Static bucket to the Open bucket
+
+max_photo_id = LocalPhoto.last.id
+batch_size = 100000
+index = 0
+
+while index < max_photo_id
+	LocalPhoto.joins("LEFT JOIN flags ON (photos.id = flags.flaggable_id)").
+  where("flags.id IS NULL").
+  where("photos.file_prefix_id=2")
```
pleary, line 10: This line needs a dot at the end, otherwise it will look like the statement ends here and a new statement begins on the next line starting with `where(` which is invalid syntax.

Same minor code formatting comments for this file - spaces before/after parens and operators, e.g. `where( "photos.file_prefix_id = 2" ).` on this line. The hard-coded prefix ID could be replaced with `FilePrefix.where(nprefix: "https://static.inaturalist.org/photos" ).first.id`

tools/migrate_photo_from_odp.rb
```diff
@@ -0,0 +1,17 @@
...
+	where("id between ? and ?", index, index + batch_size -1).
+	find_each do |photo|
```
pleary, line 10: minor detail here. `find_each` will by default fetch up to 1000 records sorted by ID, and if there are more than 1000 records it will make additional queries where ID > the last ID, again sorted by ID and limited to 1000. Sometimes that's helpful, but in this case the batches are being handled in a custom way because we know the table is huge. I would prefer to see this as an `each` and not a `find_each` to save the sort and limit in the query in hopes it would be a little more efficient

pleary (changes requested):
I left a few comments. Adding the dot after the where and updating the `unique_hash` is probably the highest priority, then maybe replacing some hard-coded values to make it more readable. Then there are the smaller code-formatting changes (indent each line with two spaces, spaces before/after parens and operators).

This scripts will likely take a while to run, and I often find it helpful to output debugging statements along the way so I can monitor how far along it is in the process. I'd probably include something like "Starting batch #{index}" just inside the while loop so I'd have an idea of how far along it is and how much longer it might take to finish

PR #3792, opened 2023-06-28
Migrate photos between buckets based on license

sylvain-morin:


tools/migrate_photo_to_odp.rb
```diff
@@ -4,20 +4,11 @@
...
-  where( "photos.license not in (?)", Shared::LicenseModule::LICENSE_NUMBERS - Shared::LicenseModule::ODP_LICENSES).
-  where( "photos.id between ? and ?", index, index + batch_size - 1 ).
-  each do |photo|
-    LocalPhoto.delay(
+  LocalPhoto.delay(
       queue: "photos",
-      unique_hash: { "LocalPhoto::change_photo_bucket_if_needed": photo.id }
-    ).change_photo_bucket_if_needed( photo )
-  end
+      unique_hash: { "LocalPhoto::migrate_to_odp_bucket": index }
```
pleary, line 11: we can add priorities to delayed jobs. Since we know these will take a while and we don't need them to happen immediately, maybe we should add `priority: INTEGRITY_PRIORITY` so if other photo moving jobs come in from user actions they could be run before all of these longer running jobs finish

app/models/local_photo.rb
```diff
@@ -600,4 +600,26 @@ def self.delete_images_from_bucket( s3_client, delete_from_bucket, images )
...
+    where( "photos.id between ? and ?", start_index, end_index ).
+    each do |photo|
+      LocalPhoto.change_photo_bucket_and_index_taxa_if_needed( photo )
+    end
+  end
+
+  def self.change_photo_bucket_and_index_taxa_if_needed( p )
+    return unless p = LocalPhoto.find_by_id( p ) unless p.is_a?( LocalPhoto )
+    p.change_photo_bucket_if_needed
+    TaxonPhoto.where( photo_id: p.id ).each do |taxon_photo|
```
pleary, line 618: Can you extract this into its own instance method - `reindex_taxa_later` or something like that? Then instead of putting it here, have it get called in `self.move_to_appropriate_bucket` near `photo.mark_observations_as_updated`. Then I think this method can be removed and you can call the existing `change_photo_bucket_if_needed` instead of `change_photo_bucket_and_index_taxa_if_needed`. That way taxa will get reindex after any move in the future, not just these related to `migrate_to_odp_bucket`

app/models/local_photo.rb
```diff
@@ -600,4 +600,26 @@ def self.delete_images_from_bucket( s3_client, delete_from_bucket, images )
     true
   end
 
+  def self.migrate_to_odp_bucket( start_index, end_index )
+    static_bucket_id = FilePrefix.where( prefix: "https://#{LocalPhoto.s3_host_alias}/photos" ) ).first.id
```
pleary, line 604: there's an extra closing parenthesis here causing a syntax error

pleary (changes requested):

PR #3804, opened 2023-07-09
3677 observation details rtl

naamaarkin:
rtl adaptation for observation details page


kueda (changes requested):
Good start! Here's a first round of requests. We're probably going to need a few more because this is such a complex page.

- [x] Edit button submenu is not RTL and the button border radius needs to be adjusted: Screenshot 2023-07-27 at 4 47 32 PM
- [x] The prev / next buttons are not RTL, e.g. the arrow button on the right still links to the next observation. Presumably in RTL it should link to the previous observation.
- [x] Map is still in LTR. This is due to some manual control positioning in map.jsx and taxonmap.js.erb. You should be able to switch them for RTL using the same `$( "html[dir='rtl']" )` trick we've used before, but let me know if that gets too complicated and I can take a look.
- [x] Map layers menu has some spacing problems next to the checkboxes: Screenshot 2023-07-27 at 4 51 17 PM
- [x] The "Standard" and "Community Curated" columns in the map details popover are not RTL: Screenshot 2023-07-27 at 5 07 57 PM
- [x] The close button in the flag modal is still in the upper right: Screenshot 2023-07-27 at 5 09 14 PM
- [x] Full screen photo viewer is not RTL (close button is in upper right, navigational buttons still treat right as meaning next)
- [x] Photo thumbnails are missing their margins: Screenshot 2023-07-27 at 5 12 17 PM
- [x] Research Grade label is broken: Screenshot 2023-07-27 at 5 13 59 PM
- [x] The taxon image in the Community Taxon is still flush right: Screenshot 2023-07-27 at 5 15 08 PM
- [x] activity item submenu icons need margin: Screenshot 2023-07-27 at 5 16 11 PM

naamaarkin:
I don't remember us using the $( "html[dir='rtl']" ) trick so I'm not sure how to use it. full screen photo navigational buttons are svg so I think they will be more complicated to change

kueda:
> I don't remember us using the $( "html[dir='rtl']" ) trick 

Check out https://github.com/inaturalist/inaturalist/blob/main/app/assets/javascripts/jquery/plugins/inat/generic_autocomplete.js#L68. It's just using jQuery to detect if the `dir="rtl"` attribute has been set on the `<html>` element.

> full screen photo navigational buttons are svg so I think they will be more complicated to change

I don't think you need to change the SVG, you just need to change the positioning of the buttons with CSS and use `transform: rotate(180deg)` to rotate them. Selecting them is probably the tricky part. Maybe something like `button[class*="arrow__direction__left"]`.

LMK if you want more specific feedback. There are still plenty of things that are not yet RTL so I'm assuming you're still working on them.

kueda (changes requested):
First, I'm really sorry it took me so long to get to this. Second, your changes to the content of the map dropdown look good, but the position of the dropdown control hasn't changed, and neither have any of the map controls. They should all flip around in RTL, right?

The rest of the unchecked todos at https://github.com/inaturalist/inaturalist/pull/3804#pullrequestreview-1551036518 still need doing too.

kueda (approved):

PR #3807, opened 2023-07-12
Ignore etag when moving multipart photos

sylvain-morin:


pleary (changes requested):
Was the only change in the most recent commit to put all the conditionals on the same line? If so, I'd say the previous version was preferred. We like to keep lines short, and putting different clauses on new lines reads a little easier. Some other small style comments - we prefer double quotes to single quotes, and to put spaces inside of parenthesis.

Could you also add a little more to the comment to make it more clear what this is doing? - that the etags of larger files uploaded multipart requests will contain hyphens, and that moving those same files with `copy_object` will generate different etags without the hyphen even though the files are the same

If the strings don't contain "-" then `index( "-" )` returns nil and `nil >= 0` raises an undefined method error. You can use `.include?` to check for if one string contains another, e.g.

```
return false unless (
  ( target_image.etag == source_image.etag ) ||
  target_image.etag.include?( "-" ) ||
  source_image.etag.include?( "-" )
)
```

Issue #3633, opened 2023-01-27
Make taxon page map filters "sticky"


tiwane:
See https://forum.inaturalist.org/t/option-to-always-show-gbif-data-on-maps/12451

Currently if I choose to add or remove map filters on a taxon page, eg

<img width="738" alt="image" src="https://user-images.githubusercontent.com/15896181/215222307-dd985cc8-735e-4c52-901e-0d091a5838ce.png">

then go to another taxon page, my choices for filter go back to the default:

<img width="517" alt="image" src="https://user-images.githubusercontent.com/15896181/215222394-d66c62dd-8ed5-4e5b-b846-5d61921f416f.png">

Request is to have those choices remain until I change them.

dartar:
I just discovered this functionality on the iNat forum and I wish I had known a GBIF observation layer existed, amazing! Making the filter selection sticky sounds a no brainer, but if this is not stored in user settings the selection would likely be only browser or device specific, correct? Have you considered:

- turning the GBIF layer on by default
- adding a user preference to make the preferred layer selection persistent _across browsers and devices_

I would also recommend using a more accessible color scheme for the GBIF layer. I don't have visual accessibility issues but the colors for verifiable observations and GBIF data are nearly indistinguishable especially in aggregate / when not zooming in the map to visualize individual points.

rainhead:
Happy to take this if no one is already on it.

PR #3808, opened 2023-07-12
Make GBIF layer on taxon page sticky

rainhead:
Implements: #3633 

Does this impact the right set of pages? I see maps on the "suggestions" tab on the identify page can have GBIF layers; should this affect them? Are there other pages with GBIF layers?

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1271,6 +1271,7 @@ protected
       :prefers_captive_obs_maps,
       :prefers_comment_email_notification,
       :prefers_forum_topics_on_dashboard,
+      :prefers_gbif_layer_maps,
       :prefers_identification_email_notification,
       :prefers_identify_side_bar,
       :prefers_message_email_notification,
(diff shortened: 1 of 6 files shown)


pleary:
This looks good to me for the taxon page. We could apply it to Identify as well, but that wasn't requested and I'm not sure if that would be generally preferred. When applied to Identify, toggling the GBIF layer does not affect all maps on the Identify suggestions tab at once. So if a user enables this on a taxon page, then ultimately doesn't want it on Identify, they have to disable it for each map one-by-one or know to disable it on one map, then reload the page to have the setting take affect for all maps. Given that, I say let's merge this in for taxon pages and hold off on Identify until that's requested.

One small suggestion indirectly related to this PR is - when enabling the GBIF layer with this PR the map will reload, ultimately zooming out and flickering a little which might be distracting. This is because we haven't implemented a `reloadKey` on taxon_page_map.jsx. If we added for example ```reloadKey={`taxa-show-map-${taxon.id}`}``` to https://github.com/inaturalist/inaturalist/blob/main/app/webpack/taxa/show/components/taxon_page_map.jsx it would avoid the map reloading when changing user settings like for the GBIF layer or the two other obs layers we offer. That doesn't need to block merging this in as its an existing problem with the other layers, but might be nice to add eventually

rainhead:
I believe this PR changes only the taxon page. Please correct me if it affects anything else.

PR #3826, opened 2023-08-02
S3 audit

sylvain-morin:


pleary:
All of these scripts reference fixed paths. Could you update them so that if writing files they are written relative to the application directory, or to a temporary directory, e.g. `Dir.mktmpdir`, or are they accept a path prefix parameter, and if they need to know a path they accept it as a parameter? Essentially remove hard-coded paths and replace them with parameters or paths that will work on any system (relative the application root or `Dir.mktmpdir`)

PR #3829, opened 2023-08-03
Fix remove_from_s3 with additional checks

sylvain-morin:


app/models/deleted_photo.rb
```diff
@@ -4,7 +4,15 @@ class DeletedPhoto < ApplicationRecord
...
-    return if removed_from_s3
+    return if removed_from_s3?
+    if orphan?
+      return if created_at > 1.month.ago
+    else
+      return if created_at > 6.month.ago
+    end
+    # do not remove from S3 if the DeletedPhoto is still associated to a Photo 
+    # (for example after resurrect)
+    return if photo?
```
pleary, line 15: This won't work with the question mark, it should just be `return if photo`

pleary:
Other than the one needed change, this is looking good and we can merge after that's fixed. Thanks!

Issue #3803, opened 2023-07-07
Hide media


kueda:
We want to use ModeratorActions to hide media. Currently the only way to do that for photos is with a copyright violation flag, but that conflates the action of getting someone's attention with the consequence of that person's decision, leading to several awkward outcomes, like flags that shouldn't be resolved and content getting hidden for the wrong reason (there's lots of inappropriate stuff that doesn't violate copyright). And there's no way to hide a sound.

## Photo
- [x] Add `has_moderator_actions %w(hide unhide)` to `Photo`
- [x] When a photo is hidden, replace it with a generic photo that indicates it was hidden, similar to how we do this for copyright violations, i.e. replace everything but the original size (use `moderated_with` similar to how we use `flagged_with`; the content of the photo doesn't matter right now, but we'll probably need it to be a design asset before merging; a better approach might be to use S3 pre-signed URLs to hide all sizes from everyone other than staff and curators, which would greatly expand the complexity here but probably save us some headaches down the road)
- [x] Hidden photos should move out of the public bucket if they're there, and unhidden photos should move back if appropriate

## Sound
- [x] Add `has_moderator_actions %w(hide unhide)` to `Sound`
- [x] When a sound is hidden... figure out how to prevent someone with the sound URL from accessing it (clearly I don't have a good solution for this yet, but open to ideas; S3-based approaches may be the best/only way)

All sounds are in the private bucket so there's no need to worry about that.

## Observation
- [x] Observations with hidden photos should be Casual Grade
- [x] Hiding an unhiding a photo should recalculate the observation's quality grade

## Obs Detail
- [x] Add a button next to the flag button for a photo or a sound (use `fa-eye-slash` to match the iconography we use for hiding other content). It should only be visible to staff and curators. Clicking it brings up the Moderator Action modal for the media. The modal should show a small preview of the media.
- [x] When a photo or sound is hidden, don't show the generic photo (or sound), but instead show the same "Content Hidden" notice we show for comments. Clicking it should show the same popover with the explanation (could probably just make a reusable component for this)
- [x] If the viewer has permission to view the hidden media (i.e. they are staff, a curator, or the person who uploaded the photo), add a "Show Hidden Content" link below the "Content Hidden" notice that links to a URL for the media object that the viewer can access (e.g. the original URL for the photo)

## Photo Detail
[...]

Issue #3802, opened 2023-07-07
Allow hiding of any identification


kueda:
The taxon you use in an identification can be inappropriate, not just the text that you add to an identification, so we don't think having remarks should be a requirement for hiding an identification, and we don't think hidden identifications should contribute to attributes of the observation, like the community taxon.

- [x] Show the "Hide Content" option for identifications even if there are no remarks
- [x] Remove hidden identifications from consideration when determining the Observation Taxon and the Community Taxon
- [x] Hiding and unhiding an identification should cause the observations to re-calculate the Observation Taxon, Community Taxon, and quality grade (use `moderated_with`, e.g. the way we're [using it](https://github.com/inaturalist/inaturalist/blob/main/app/models/user.rb#L1464) when suspending a user)
- [x] Remove hidden identifications from `/identifications/:login` unless you are viewing your own identifications
- [x] Omit hidden identifications anywhere we count identifications, so user profile, Explore, Activity Item, taxon detail

tiwane:
Just thought about this - should hidden IDs count toward the user's ID count on their profile page? I guess hiding them makes them inactive and thus the IDs wouldn't be counted anyway, correct?

kueda:
Up to you, though we should be careful about omitting hidden idents everywhere if we do that to avoid confusion over discrepancies. Whether or not an identification is hidden is in the index so it should be easy to omit hidden identifications from counts (famous last words).

tiwane:
I think anywhere we count active IDs only we should not count hidden IDs, since they're not active on the observation.

kueda:
This is looking good, Patrick. IMO, the only thing beyond the remaining todo regarding counts is this:

- [ ] The modal preview of the content that will be hidden should show the identification taxon. If there are no remarks, I don't think there's a need to include that empty blockquote.

pleary:
I updated the modal to include a preview of the ID using the ActivityItem component. I also updated it to use the same for comments since it was easy and looked nice - but we can revert that if we don't like it.

As for counts - I've updated the counts I noticed, but I left it unresolved because we show counts in a lot of places and it would be good to get some extra eyes on making sure all relevant counts are updated.

Issue #3800, opened 2023-07-07
ModeratorActions for all comments


kueda:
At present, ModeratorActions can only be created for comments and identifications in React-based interfaces, e.g. on obs detail. For identifications this is sufficient because they must be associated with observations, but comments can be associated with a variety of models, and people performing moderation duties need to be able to hide inappropriate comments in all places where comments appear, so we would like to add a Rails-based tools for creating ModeratorActions for comments

## GET `/comments/:id/hide`
- [x] Add GET `/comments/:id/hide`
- [x] It should render a form that POSTs to `/moderator_actions` to create a `HIDE` action for that comment
- [x] The form should include a preview of the comment
- [x] The form must contain a `textarea` for the reason attribute that enforces `ModeratorAction::MINIMUM_REASON_LENGTH` and `ModeratorAction::MAXIMUM_REASON_LENGTH`
- [x] The form should include the same warning text we do on obs detail (`hide_desc`)

## `shared/activity_item` partial
- [x] Add an option to "Hide" in the mouseover controls if the authenticated user is a curator, and link that to the new `/comments/:id/hide`
- [x] If a comment is hidden and the viewer can see it (viewer is an admin, viewer is a curator, or viewer is the person who created the comment), the meta row show "Content Hidden" and the `fa-eye-slash` icon like we do on obs detail
- [x] If a comment is hidden and the viewer cannot see it, the "Content Hidden" notice and the timestamp are all that should be visible (no user icon, no username, etc.)
- [x] When a user clicks the "Content Hidden" notice, a popover should appear that includes everything it does on obs detail:
  - [x] Who hid the comment
  - [x] When the comment was hidden
  - [x] The ModeratorActions `reason`
  - [x] Link to flags on the comment
  - [x] If the viewer is the author of the comment, a link to /help

kueda:
> Is this for any user? Right now if a user is not a curator they don't see anything if they click on "Content Hidden" on obs show.

I don't think that's the case. When I go to https://www.inaturalist.org/observations/156791372 while signed out, I can see a comment was hidden, and when I click "Content Hidden" I see the popover. We can remove that if you like, but that seems to be the way it works now.

tiwane:
OK, I can see the popover now. I think it's fine to keep.

PR #3850, opened 2023-08-29
Hiding content

pleary:
- Curators can hide media #3803
- Curators can hide IDs without remarks #3802
- Curators can hide any comment #3800


config/locales/en.yml
```diff
@@ -2322,12 +2323,11 @@ en:
...
     Hiding content will remove it from public view, but people will still be
     able to see that something was hidden, and site curators and the author of
-    the content will still be able to see the hidden content. Generally, this
-    only applies to written content, e.g. if you hide the content of an
-    identification, the identification is still visible but the identifier's
-    remarks are hidden. You cannot undo this, and only site staff can unhide
-    content once it has been hidden, so be careful and only use this for very
-    objectionable content.
+    the content will still be able to see the hidden content. Hidden
+    identifications will not contribute to an observation's Community Taxon.
```
kueda, line 2327: FWIW, I think it's better to change the key when meaning of the content changes. That way translators get notified about new strings to translate. I don't think they get notified in situations like this. The approval just gets removed.

PR #3883, opened 2023-09-29
fix: :bug: fix autocomplete leaving window on upload page

HannesOberreiter:
Currently the autocomplete popup will leave the visible window space if no space is left on the left side, which causes a scrollbar to appear.

jQuery UI has a build-in option to mitigate this problem (https://api.jqueryui.com/position/):

> "flipfit": First applies the flip logic, placing the element on whichever side allows more of the element to be visible. Then the fit logic is applied to ensure as much of the element is visible as possible.

I stumble quite often over this "small issue" as I upload frequently multiple images and though this simple fix could be a good addition. 

## After fix behaviour

PS: Don't know if I should write an exception for RTL as with my current implementation the RTL style would not be applied.
PPS: I was to lazy to look into the VisionAPI therefore I don't know if everything is ok when the results come in, but I would expect there is no manual css manipulation on the container.

Cheers
Hannes


pleary:
Thanks!

Issue #3869, opened 2023-09-14
Switch YIR publications to GBIF literature endpoint


kueda:
We're currently hitting https://www.gbif.org/api/resource/search but while that works, it doesn't seem to be officially supported anymore, so we should switch to https://www.gbif.org/developer/literature.

thatkaiguy:
I took a stab at this and have a WIP pr here https://github.com/inaturalist/inaturalist/pull/3887 but I do have a question..

Between the two endpoints, the responses seem to be the same except for the `_gbifDOIs` key. In the new response, the 'gbifDOI' is found in the returned 'tags'.

[non official api example call](https://www.gbif.org/api/resource/search?contentType=literature&gbifDatasetKey=50c9509d-22c7-4a22-a47d-8c48425ef4a7&limit=50&literatureType=journal&year=2020)

results in:
```javascript
"_gbifDOIs": [
  "doi:10.15468/dl.ruflo8"
]
```

[official api example call](https://www.gbif.org/api/resource/search?contentType=literature&gbifDatasetKey=50c9509d-22c7-4a22-a47d-8c48425ef4a7&limit=50&literatureType=journal&year=2020)

results in:
```javascript
  "tags": [
    ...,
    "gbifDOI:10.15468/dl.ruflo8",
  ]
```

The returned string has a different prefix also `doi:` vs `gbifDOI:`.

Does this prefix matter? I'm not familiar with what uses this value

My spec currently fails with
```
Diff:
       @@ -1,4 +1,4 @@
       -"_gbifDOIs" => ["doi:10.15468/dl.rwjrbf"],
       +"_gbifDOIs" => ["gbifDOI:10.15468/dl.rwjrbf"],
```

kueda:
Thanks for your work on this so far, @thatkaiguy.

> Does this prefix matter?

[...]

PR #3887, opened 2023-09-30
Switch YIR publications to GBIF literature endpoint

thatkaiguy:
Tackles https://github.com/inaturalist/inaturalist/issues/3869

This PR modifies the `YearStatistic.publications` method to use the official gbif official literature api (instead of the unofficial api)

The previous method was moved to `YearStatistic.publications_old`. [This spec](https://github.com/inaturalist/inaturalist/pull/3887/files#diff-d1517c0856fe9c8627e58c8660a1003184ed062aa788345a6da7ec10fe6ec641) checks that the output of the `YearStatistic.publications(2022)` and `YearStatistic.publications_old(2022)` methods are the same.

I intend to remove the `publications_old` method in a followup PR.

**Before**

- `YearStatistic.publications` queries `https://www.gbif.org/api/resource/search`

**After**

- `YearStatistic.publications` queries `https://api.gbif.org/v1/literature/search`
- returned "_gbifDOIs" are stripped of their 'doi:' prefix eg "10.15468/dl.rwjrbf" instead of "doi:10.15468/dl.rwjrbf"
- introduced the [vcr gem](https://github.com/vcr/vcr). the gem records the responses from http calls when running a spec and stores the responses in static files in the test suite. main intent is to prevent specs from making actual http calls to improve performance.


spec/spec_helper.rb
```diff
@@ -223,6 +223,13 @@ def request( flickraw, type, method, params )
   end
 end
 
+VCR.configure do |config|
+  config.allow_http_connections_when_no_cassette = true
```
thatkaiguy, line 227: this config is by default `false` and was causing build errors b/c there were other specs making actual http calls.

spec/spec_helper.rb
```diff
@@ -223,6 +223,13 @@ def request( flickraw, type, method, params )
...
+  config.cassette_library_dir = "fixtures/vcr_cassettes"
+  config.hook_into(:webmock)
+  config.ignore_localhost = true
```
thatkaiguy, line 230: setting this to true b/c specs were making calls to elasticsearch locally

spec/models/year_statistic_spec.rb
```diff
@@ -1,6 +1,30 @@
...
+  describe ".publications" do
+    it "should match the YearStatistic.publications method" do
+      VCR.use_cassette( "search_gbif_literature" ) do
+        publications = YearStatistic.publications( 2022 )
+        old_publications = YearStatistic.publications_old( 2022 )
+
+        publications[:results].each.with_index do | result, i |
+          expect( result.except( "_gbifDOIs", "title" ) ).to eq(
+            old_publications[:results][i].except( "_gbifDOIs", "title" )
+          )
```
thatkaiguy, line 18: the official and unofficial apis returned slightly different titles. for example:
```
-"title" => "Updating the distribution of Vampyrum spectrum (Chiroptera, Phyllostomidae) in Colombia: new localities, potential distribution and notes on its conse…",
+"title" => "Updating the distribution of Vampyrum spectrum (Chiroptera, Phyllostomidae) in Colombia: new localities, potential distribution and notes on its conservation",
```

kueda (changes requested):
This looks good and works well for me. VCR is neat, I haven't used it before, but it seems like a nicer way to handle making fixtures from HTTP responses.

That said, we don't need to maintain the old code or indefinitely test that the new code generates something like the old code, so please remove `YearStatistic.publications_old`. If you want to test the processing, maybe just test that that the `gbifDOI` prefixes get removed instead of testing the new and old processing results.

thatkaiguy:
Sg, will work on an update this week

thatkaiguy:
@kueda pr updated and ready for a re-review!

kueda (approved):

kueda (approved):

PR #3913, opened 2023-10-26
Require controlled term labels

kueda:
Requires the presence of a label for new controlled terms and validates that all new labels have the `label` and `description` attributes.

This kind of throws testing for a loop since to make a valid controlled term you now need to build out an associate, so I'm open to ideas about how to make that better.


pleary (approved):
This all looks good to me

PR #3919, opened 2023-10-30
Limits to taxon lookups from photo filenames

kueda:
We try to use word combinations from filenames to look up taxa to accommodate names that might be in the middle of a filename, but this can result in some gargantuan database queries. This tries to limit this problem by:

* Only looking for 1-, 2-, and 3-word combinations
* Ensuring the list of tags extracted from a photo is unique


pleary (approved):
looks good to me

Issue #3903, opened 2023-10-14
Add an "Are you sure?" warning when flagging a user as spam


tiwane:
From https://forum.inaturalist.org/t/create-an-are-you-sure-popup-when-suspending-a-user-for-spam/45978

For curators, it's easy to accidentally select the "flag as spammer" option in the Admin Tools on a user's profile, so some user are getting flagged by mistake. Can we add a warning pop-up that says "Are you sure you want to flag this user as a spammer?" with buttons that say "Yes" and "No"?

I think the "Flag as spammer" option should be above the line as well, in the same section as "Suspend user".

Maxhou00:
Hello, 

I made the changes and created a [PR](https://github.com/inaturalist/inaturalist/pull/3925) for this issue.

It will need translation for different languages.

Note: I didn't change the position of the option since it was your personal suggestion and not an issue talk on the initial topic. I can make the change if needed.

PR #3925, opened 2023-11-01
#3903: Add confirm dialog when flag a spammer

Maxhou00:


app/helpers/users_helper.rb
```diff
@@ -98,4 +98,20 @@ def you_or_login(user, options = {})
...
+      set_spammer_path(user, spammer: true, return_to: return_to)
+    else
+      set_spammer_path(user, spammer: true)
+    end
+
+    link_to(
+      t(:flag_as_spammer),
+      spammer_path,
+      method: :post,
+      data: { confirm: t(:are_you_sure_you_want_to_flag_as_spammer, user: user) },
```
kueda, line 113: Please change this to text specified in the issue: "Are you sure you want to flag this user as a spammer?" Adding a variable isn't really necessary here, and it generally makes text harder to translate.

config/locales/fr.yml
```diff
@@ -664,6 +664,7 @@ fr:
   are_you_sure_you_want_to_remove_this_observation_from_project: |
     Êtes-vous sûr de vouloir supprimer cette observation de %{project}?
   are_you_sure_you_want_to_remove_this_taxon_from_checklist: "Êtes-vous certain de vouloir supprimer ce taxon de la check-list de « %{place} » ?"
+  are_you_sure_you_want_to_flag_as_spammer: Êtes-vous sûr de vouloir signaler %{user} comme spammeur ?
```
kueda, line 667: Please don't add translations. We manage translations through https://crowdin.com/project/inaturalistweb, so the text will be translated by volunteers once the source strings from en.yml are merged into the main branch.

kueda (changes requested):
Apologies for the late review. This looks good, just a few minor changes. Please go ahead and move the link above the line in the dropdown menu on user detail as requested in the issue as well. @tiwane is a member of iNat staff and his requests should be considered part of the requirements of the issue.

kueda (approved):

PR #3934, opened 2023-11-14
YIR 2023 shareable

### LINKED ISSUE #3930: YIR 2023 shareable
We want a new style of YIR shareable for 2023. Here's what it looks like:

User:
YIR2023-User

Global:
YIR2023-Global

Site:
YIR2023-Network Rebranded

And here's the background as an SVG: 
YIR2023-BG-2 (1)

Some requirements:
- [x] Wordmarks should have white text. I think this can be achieved for the SVGs by replacing `fill="#000000"` with `fill="#FFFFFF"`. For non-SVG site wordmarks let's just try rendering them as-is for now
- [x] This should only apply to shareables where year >= 2023. If someone regenerates their 2022 YIR it should use the old shareable.
- [x] Should not impact the generation of YearStatisticLocalizedShareableImage records, i.e. when they're created or regenerated, they should just get the new shareable as their attachment

Some nice-to-haves

- [ ] For global and site shareables, it might be nice to floor to 2-3 significant figures, like instead of 34,584,112, we would show 34,000,000+. That will probably require new i18n keys, though.
- [ ] Is there a way we can not lay these things out with imagemagick? It makes layout and changes to layout quite difficult. If we could render HTML/CSS with headless Firefox or something that would be much easier from the development standpoint, but maybe the sysadmin lift is to much

kueda:
Add a new YIR shareable design for 2023 and onward, but also makes a few other
changes that might merit a glance:

* made some changes to make shareable image testing a little easier in dev,
  like disabling asset caching if caching isn't enabled generally, and
  providing a way to alter the temp dir to see intermediate files
* used Whitney for all shareables that only use characters in that font. This
  may or may not be worth it. Some Latin script languages look better, but
  some diacritics on capital letters may get clipped. We may need to walk
  this back for some languages if that clipping is intolerable
* stopped forcing capitalization of text; this takes case control away from
  translators, and in some cases may change meanings or at least tone of the
  text
* added tests for YearStatistic generation; this stubs out network requests
  and arguably doesn't really test real world conditions since there's no
  data, but it gave me some reassurance that I wasn't breaking things while I
  was altering the shareable generation

Closes #3930


pleary (approved):
Looks good. Working well for me locally

PR #3957, opened 2023-12-13
Paperclip path versioning

pleary:
- Allow paperclip path interpolation of any instance attribute or method
- Models implement a method `{attachment}_versions` that returns a hashed version of `{attachment}_updated_at` that can be used in paperclip path interpolation
- Allow models to have multiple paths for an attachment to enable no-downtime transitioning from one path scheme to another, or indefinitely support multiple paths for attachments
- Add additional paths for some attachments whose paths include attachment_version for unique paths as attachments are updated
  - User.icon
  - Project.icon
  - Project.cover
  - YearStatisticLocalizedShareableImage.shareable_image
- Add `{attachment}_path_version` for the above model attachments to record which path version is being used by the attachments
- Remove `invalidate_cloudfront_caches` for the above attachments as cache invalidation is not needed if the URLs change based on attachment version


kueda (approved):
Apologies for the late review. This looks good to me, only minor comments about clarification as it was not immediately obvious to me what was happening by reading the code in the models.

PR #3975, opened 2024-01-10
Observation accuracy experiment

loarie:
functionality to conduct assessments of observation accuracy.
3 new models: ObservationAccuracyExperiment model has many observation_accuracy_samples and observation_accuracy_validators
Screenshot 2024-01-10 at 8 17 35 AM

ObservationAccuracyExperiment defines the experiment parameters and statistics
ObservationAccuracySample is a snapshot of sample observations from the time they were sampled
ObservationAccuracyValidator has columns for the user_id assigned to validate the observation and the corresponding observation_id and observation_accuracy_experiment_id

The protocol is to create a ObservationAccuracyExperiment of a fixed sample _size, validator_redundancy_factor
this automatically runs generate_sample which stores the sample in ObservationAccuracySample and them distribute_to_validators which creates ObservationAccuracyValidator by finding suitable user_ids to cover the sample (with redundancy from validator_redundancy_factor).

contact_validators sends emails to each potential validator with links to their sample (replace_validators_for_testing replaces the ObservationAccuracyValidator with a random subset of the sample assigned to provided user_ids for testing)

assess_experiment uses new identifications created on observations by users stored in ObservationAccuracyValidator to calculate accuracy statistics and store them on the ObservationAccuracyExperiment

generate_report uses the statistics stored onObservationAccuracyExperiment to generate a pretty report

issues:
1) distribute_to_validators is very slow because get_candidate_identifiers_by_taxon uses multiple API calls including an a call to the identifications endpoint for each taxon represented in the sample to find candidate validators based on improving identifications. 

2) To quickly crunch continent level stats, the sample doesn't store raw lat-lon but rather continent strings (e.g. Europe), to get these it makes calls to the Observation API endpoint to fetch the place_id, this is a bit slow

3) Likewise, to quickly crunch precision stats the sample stores number of descendants which is a bit slow to calculate since it involves querying the number of leaf descendants for each taxon represented in the sample

^With these 1-3 speed issues, a sample of 1,000 still runs in a few hours. This is the size of the sample we plan to launch with. But if we choose to work with larger sample sizes down the road (e.g. 100,000) we'll need to figure out a way to speed them up.

4) If we get a situation where multiple validators of the same observation disagree with one another the code currently doesn't try to assess accuracy stats. I expect this to be a rare occurrence. If it does occur we'll deal with it on a case-by-case basis by intervening to try to reach consensus.

5) Major assumptions are that the validators don't provide incorrect identifications, that the sample is representative of the iNat dataset, and that we'll get enough skilled participation to estimate useful statistics.

Here's an example of what the generated emails look like
Screenshot 2024-01-10 at 8 36 24 AM
and here's an example of a subsample link created for a top beetle identifier
https://www.inaturalist.org/observations/identify?quality_grade=needs_id%2Cresearch%2Ccasual&id=9536296,102286521,102286521,140482043,165962622,171630233,125537075,21747186,91083260,25440044,168787251,142444414,144605201,144605201,76228678,50272817,118253841,180160137,122475191,122475191,82253842,82253842,30156627,85251849,194951372,121377663

Here's an example of a report generated on a sample before being sent to validators. This is reflected in the low low number of validated samples and responding validators (all coming from existing IDs already made by the validator)
 
```
Sample consistes of 1000 observations generated on 2024-01-10 10:54:09 UTC

Sample grouped by taxonomic rank:
subspecies: 17, species: 731, genus: 120, family: 51, order: 19, class: 19, phylum: 7, kingdom: 9, life: 27

Sample grouped by observation count:
1-100: 32, 100-1000: 94, 1000-10000: 256, 10000-100000: 432, >100000: 186

Sample grouped by Quality grade:
research: 578, needs_id: 326, casual: 96

Sample grouped by Iconic taxon name:
Plantae: 430, Insecta: 220, Aves: 125, Fungi: 65, Arachnida: 29, : 27, Mammalia: 26, Animalia: 19, Mollusca: 19, Amphibia: 15, Reptilia: 12, Actinopterygii: 10, Protozoa: 2, Chromista: 1

Sample grouped by Year:
2023: 263, 2022: 208, 2021: 182, 2020: 148, 2019: 89, 2018: 56, 2017: 21, 2015: 13, 2016: 10, 2014: 5, 2013: 3, 2024: 1, 2010: 1

Sample grouped by Continent:
North America: 562, Europe: 182, Asia: 89, Oceania: 52, South America: 46, : 40, Africa: 29

Distributed to 1796 validator(s)
with a validator redundancy factor of 10
Experiment assessed on 2024-01-10 15:11:33 UTC
At that time 323 validator(s) responded (18.0%)
and validated 347 sample(s) (35.0%)

Accuracy (lower): 0.344 +/- 0.0154
Accuracy (higher): 0.999 +/- 0.001
Precision: 0.7819 +/- 0.0127
```


loarie:
After talking with @pleary I'm going to make some changes to the structure of the models. So balls in my court

loarie:
@pleary, I rewired this up like this as we discussed. This is ready for a review when you have a chance
Screenshot 2024-01-10 at 3 31 03 PM

app/models/observation_accuracy_experiment.rb
```diff
@@ -0,0 +1,559 @@
...
+      self.high_accuracy_mean = quality_stats[2]
+      self.high_accuracy_variance = quality_stats[3]
+      self.precision_mean = precision_stats[0]
+      self.precision_variance = precision_stats[1]
+      save!
+    end
+  end
+
+  def get_sample_for_user_id( user_id )
+    return nil unless validator == ObservationAccuracyValidator.where( user_id: user_id ).first
```
pleary, line 543: is validator defined somewhere? If not, did you intend to assign to validator with `=` instead of comparing with `==`?

db/migrate/20240109035854_create_observation_accuracy_validator.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class CreateObservationAccuracyValidator < ActiveRecord::Migration[6.1]
+  def change
+    create_table :observation_accuracy_validators do | t |
```
pleary, line 5: I think it would be good to have a column for `observation_accuracy_experiment_id` and to have a unique index on `user_id` and `observation_accuracy_experiment_id`. That way its clear which experiment these users are associated with, ensure that users are associated with a single experiment once, but that they can also be associated with different experiments (and be emailed about the individually) over time

app/models/observation_accuracy_validator.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class ObservationAccuracyValidator < ApplicationRecord
+  belongs_to :user
+  has_and_belongs_to_many :observation_accuracy_samples
+end
```
pleary, line 6: if you do add `observation_accuracy_experiment_id` to `observation_accuracy_validators`, it would be good to add a uniqueness validator here, e.g. `validates_uniqueness_of :user_id, scope: :observation_accuracy_experiment_id`

pleary (changes requested):
I left a few comments and suggestions. One of which is a potential data model change to add `observation_accuracy_experiment_id` to `observation_accuracy_validators` and make a unique index on that and `user_id`, to make it easier to keep track of users across multiple experiments over time.

Overall looks pretty good to me. Code formatting is great.

loarie:
Thanks for the review @pleary - I made all the changes you suggested and I think this is now ready to go
Screenshot 2024-01-11 at 11 50 59 PM

app/models/taxon.rb
```diff
@@ -1645,6 +1645,14 @@ def descendant_of?( taxon )
     ancestor_ids.include?( taxon.id )
   end
 
+  def sibling_of?( taxon )
```
pleary, line 1648: I'd change the name of this method, since I think it would be common to think of "siblings" as only the taxa that share the same parent, not any two taxa with disjoint ancestries. Maybe invert some of the logic and call it `in_same_branch_of?` or something like that? A taxon is in the same branch of another taxon if one is an ancestor of the other, or one is a descendant of the other

pleary (approved):
I made one small naming change suggestion, but otherwise this looks good to me

PR #3989, opened 2024-01-25
Observation accuracy experiment

loarie:
@pleary here's a first pass at ObservationAccuracyExperiment/show that does pretty much everything dynamically. Curious for feedback


app/controllers/observation_quality_experiments_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+    if @tab == "methods"
+      @candidate_validators, @mean_validator_count, @mean_sample_count = get_assignment_methods
+      @mean_validators_per_sample, @validators_per_sample, @validators_per_sample_ylim = get_validation_methods
+    else
+      @stats, @data, @precision_data = get_results_data
+    end
+    render "show"
+  end
+
+  private
```
pleary, line 32: these private methods in this controller seem to be about gathering data from experiments. These would preferably be implemented as methods in the experiment model to get this logic out of the controller. The instance variables the methods depend on could be changed to method parameters

app/assets/javascripts/observation_accuracy_experiments/show.js.erb
```diff
@@ -0,0 +1,124 @@
+$(function() {
```
pleary, line 1: Some syntax here doesn't conform to our JS linting rules defined at https://github.com/inaturalist/inaturalist/blob/main/app/assets/javascripts/.eslintrc. I'd recommend getting a JS linter set up in your editor. Alternatively you can run eslint directly, e.g. `./node_modules/.bin/eslint app/assets/javascripts/observation_accuracy_experiments/show.js.erb`

db/migrate/20240124214427_set_default_values_to_zero_in_observation_accuracy_experiments.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class SetDefaultValuesToZeroInObservationAccuracyExperiments < ActiveRecord::Migration[6.1]
+  def change
+    change_column_default( :observation_accuracy_experiments, :responding_validators, 0 )
```
pleary, line 5: I tested reverting this migration and got an error `change_column_default is only reversible if given a :from and :to option.`. Ideally this would list the previous value so this could be rolled back without an error, e.g. 
```
change_column_default( :observation_accuracy_experiments, :validated_observations, from: nil, to: 0 )
```
I think its a good general practice to test rolling back new migrations before committing so be sure they can be created and destroyed, in case we ever need to roll back a deploy. That can be tested with `rake db:migrate:redo` or `rake db:migrate:down VERSION=20240124214427`

db/migrate/20240124214436_set_default_reviewers_to_zero_in_observation_accuracy_samples.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class SetDefaultReviewersToZeroInObservationAccuracySamples < ActiveRecord::Migration[6.1]
+  def change
+    change_column_default( :observation_accuracy_samples, :reviewers, 0 )
```
pleary, line 5: Another place including `from` and `to` options would be preferred

app/controllers/observation_quality_experiments_controller.rb
```diff
@@ -0,0 +1,101 @@
+# frozen_string_literal: true
+
+class ObservationAccuracyExperimentsController < ApplicationController
```
pleary, line 3: This file should be named `observation_accuracy_experiments_controller.rb`, not `quality`

app/controllers/observation_quality_experiments_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+  layout "bootstrap"
+
+  def get_more_validators
+    experiment_id = params[:id]
+    @experiment = ObservationAccuracyExperiment.find( experiment_id )
```
pleary, line 8: If an invalid ID is provided, this will raise an error. We handle this in different ways in different controllers. See for example the Announcements controller https://github.com/inaturalist/inaturalist/blob/main/app/controllers/announcements_controller.rb#L6 where we use a before_action to attempt to load the record, or raise a 404 if the record is not found

app/controllers/observation_quality_experiments_controller.rb
```diff
@@ -0,0 +1,101 @@
...
+    @more_validators = @experiment.get_validator_names( limit: nil, offset: 20 )
+    respond_to do | format |
+      format.html { render partial: "additional_validators", locals: { validators: @more_validators } }
+    end
+  end
+
+  def show
+    @experiment = ObservationAccuracyExperiment.find( params[:id] )
```
pleary, line 16: This method could leverage the same before_action if you define one based on my above comment

app/models/observation_accuracy_experiment.rb
```diff
@@ -586,16 +431,224 @@ def get_sample_for_user_id( user_id )
...
+    else ">100k"
+    end
+  end
+
+  def normalize_counts( counts )
+    total = counts.values.flatten.count
+    stats = counts.transform_values do | ids |
+      norm = ( total.zero? ? 0 : ids.count / total.to_f * 100 ).round( 2 )
+      {
+        url: FakeView.observations_url( verifiable: "any", place_id: "any", id: ids.join( "," ) ),
```
pleary, line 498: Not a big deal, but FakeView is a bit of a hack that if we can avoid would be better. Instead of having the model generate the URL, it could return an array of observation_ids which could then be converted to a URL in the view which has access to URL helpers

app/models/observation_accuracy_experiment.rb
```diff
@@ -586,16 +431,224 @@ def get_sample_for_user_id( user_id )
...
-    oids = samples.map( &:observation_id ).join( "," )
-    FakeView.identify_observations_url(
-      place_id: "any",
-      reviewed: "any",
-      quality_grade: "needs_id,research,casual",
-      id: oids
-    )
+    counts = {
+      incorrect: observation_accuracy_samples.
+        where( [subset_filter, "correct = 0"].compact.join( " AND " ).to_s ).count,
```
pleary, line 443: instead of doing a compact and join here, this can just chain multiple `.where` together. e.g. `observation_accuracy_samples.where( subset_filter ).where( "correct = 0" ).count`. When subset_filter is nil, adding a `.where( nil )` is just ignored. Alternatively, you can set up a scope at the top of the method and append to it in the case switch before querying. Something like this: 
```
    counts_base_scope = observation_accuracy_samples
    case subset
    when "verifiable_results"
      counts_base_scope = counts_base_scope.where( "quality_grade = 'research' OR quality_grade = 'needs_id'" )
    when "all_results"
      # do nothing
    else
      counts_base_scope = counts_base_scope.where( "quality_grade = 'research'" )
    end

    counts = {
      incorrect: counts_base_scope.where( "correct = 0" ).count,
      uncertain: counts_base_scope.where( "correct IS NULL OR correct = -1" ).count,
      correct: counts_base_scope.where( "correct = 1" ).count
    }
```

app/views/observation_accuracy_experiments/_methods.html.haml
```diff
@@ -0,0 +1,103 @@
...
+            %text{ transform: "rotate(-90)", y: "6", dy: ".71em", style: "text-anchor: end;"} Frequency
+
+          %g.x.axis{ transform: "translate(0,100)" }
+            - @validators_per_sample.keys.each_with_index do |label, index|
+              %g.tick{ transform: "translate(#{35 + 54 * index},0)", style: "opacity: 1;" }
+                %line{ y2: "6", x2: "0" }
+                %text{ dy: ".71em", y: "9", x: "0", style: "text-anchor: middle;" }= label
+
+          - @validators_per_sample.values.each_with_index do |values, index|
+            %a{ href: @explorable ? FakeView.observations_url( verifiable: "any", place_id: "any", id: values.join( "," ) ) : '#' }
```
pleary, line 66: no need to reference FakeView in a view as they have access to URL helpers directly

pleary (changes requested):

PR #4008, opened 2024-02-15
inaturalist#3988: Add species rank for capitalizing name

Maxhou00:
This fix is only solving the issue for the species rank. I don't have enough knowledge to know if this kind of situation can happen on other rank.

If it can then it would be better to improve the `capitalize_scientific_name` method in `Taxon` model.

I could work on it if needed.

Also after deploying this PR, every taxon that was created before it the should be forced capitalized and reindexed. 
Something like this can be enough, just need to manage the case where `×` is used with a real `x` 

```
Taxon.where("name LIKE '%×%'").each(&:save)
```

I don't know if I have to do a script, a task or a migration for it.

### DIFF
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -760,7 +760,7 @@ class Taxon < ApplicationRecord
   def self.capitalize_scientific_name( name, rank )
     return name.capitalize if rank.blank?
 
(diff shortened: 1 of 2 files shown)


pleary:
Thanks!

PR #4009, opened 2024-02-15
Taxon Change retry and cleanup

pleary:
- catches and retries after some errors experienced when updating associates
- run a task after a taxon change ends to assess and update observations that for whatever reason have an incorrect taxon


pleary:
@loarie when you get a chance, could you check if this is doing what you'd want it to do. The two main changes are wrapping `find_batched_records_of` in a `try_and_try_again` catching the errors you described experiencing, and a method that runs at the end of a taxon change to validate the taxon_id of any observation that should have been affected by the change

PR #4013, opened 2024-02-21
Multiple photos dqa

loarie:
We estimate there are ~350k observations on iNaturalist with [multiple photos not pertaining to the same subject](https://www.inaturalist.org/observations/19543605). Usually this happens when new users don't realize in the observation flow that they are adding unrelated photos to the same observation rather than to each of new photos.

Because iNat assumes that each observation has a single subject, this creates problems in features like the CV model (when we train on images of plants labeled with frog names), [browse photos](https://www.inaturalist.org/taxa/62345-Duttaphrynus-melanostictus/browse_photos) (when pictures of plants are showing up on the frog taxon page).

The norm on iNaturalist is to:
1. add a disagreeing ID of the common ancestor (e.g. Arthropoda for an observation with photo 1 of a beetle and photo 2 of a spider)
2. vote "no" to "Based on the evidence, can the Community Taxon still be confirmed or improved?"

But there is continued conflict in the community about the best way to remove these from the verifiable pool.

Here, we've added a  DQA flag "Evidence all relates to the same subject" as an easier way to remove such observations from the verifiable pool.

Screenshot 2024-02-20 at 4 32 06 PM

Note:

1. this is not intended to be a way to salvage the data (e.g. a way for the community to force the subject to be the first photo and indicate that all other photos should be ignored)
2. a tool for flagging controversial practices like habitat photos or bat sonograms that DO relate to the same subject
3. a tool for resolving ambiguous subject situations - e.g. observer posts a photo of a mallard and a pigeon and doesn't indicate through an ID or description what the subject is

Questions:

@kueda @budowski - made these changes to the API https://github.com/inaturalist/iNaturalistAPI/tree/multiple_photos_dqa
do changes need to be made to the mobile apps to prevent things from breaking?

@abhas - any feedback on the icon?

any other feedback on this feature and the implementation


budowski:
This does need an update on the Android side, but should be a small fix

loarie:
Thanks @budowski, I made one here https://github.com/inaturalist/iNaturalistAndroid/issues/1333
@alexshepard would adding a new DQA flag break anything on the iOS app?

PR #4014, opened 2024-02-22
Observation accuracy experiment updates

loarie:
A few updates to Observation Accuracy Experiments including:
* allowing location to be considered in selecting validators
* messages instead of emails when contacting validators


loarie:
@pleary, If you have time for a quick review would love to deploy this Thurs Feb 22 if possible so this is available over the weekend in preparation for Feb 26 - thanks!

app/models/observation_accuracy_experiment.rb
```diff
@@ -336,15 +465,63 @@ def generate_sample
     puts "Sample generated in #{duration} seconds."
   end
 
-  def contact_validators( validator_deadline_date: 1.week.from_now.strftime( "%Y-%m-%d" ) )
+  def observation_accuracy_validator_contact( validator )
+    return false unless ( user = User.where( id: validator.user_id ).first )
+
+    admin = User.where( email: "admin@inaturalist.org" ).first
```
pleary, line 471: This might be better set up as a config variable. You're looking to load a particular iNaturalist user by email, and maybe the email to use for lookup should be defined in config.yml and the config value used here to look it up, rather than hardcoding the email in this model. We don't reference this email address anywhere else in this codebase

app/models/observation_accuracy_experiment.rb
```diff
@@ -336,15 +465,63 @@ def generate_sample
...
+      identify_observations_url(
+        place_id: "any",
+        reviewed: "any",
+        quality_grade: "needs_id,research,casual",
+        id: obs_ids.join( "," )
+      )
+    experiment_url = FakeView.observation_accuracy_experiment_url( self, params: { tab: "methods" } )
+    delimited_num_obs = ApplicationController.helpers.number_with_delimiter( num_obs )
+    subject = I18n.t( :observation_accuracy_validator_email_subject2, version: version )
+    message_body = <<~HTML
```
pleary, line 484: You previously used custom Emailer templates for emails for experiments in the past. Any reason to switch back to this non-standard approach to sending emails now?

app/models/observation_accuracy_experiment.rb
```diff
@@ -136,9 +145,57 @@ def get_candidate_identifiers_by_taxon( observation_ids_grouped_by_taxon_id, top
     identifiers_by_taxon_id
   end
 
-  def associate_observations_with_identifiers( obs_ids_grouped_by_taxon_id, identifiers_by_taxon_id, top_iders )
-    validator_candidates = identifiers_by_taxon_id.values.flatten.uniq
-    observation_ids = obs_ids_grouped_by_taxon_id.values.flatten.uniq
+  def get_candidate_identifiers_by_taxon_and_continent( observation_ids_grouped_by_taxon_and_continent, top_50_iders )
+    continents = Place.where( admin_level: -10 ).map( &:id )
```
pleary, line 149: `Place::CONTINENT_LEVEL` would be preferred over using `-10` here

app/models/observation_accuracy_experiment.rb
loarie: many people didn't notice/recieve emails for whatever reason so we're changing to using the messaging infastructure (which also generates emails)

app/models/observation_accuracy_experiment.rb
loarie: Thanks - I'll slack you about how to make something like CONFIG.admin_user_email available

Issue #4018, opened 2024-02-25
Treat ancestor disagreements as non-disagreements if all IDs of the disagreeing taxa have been withdrawn/deleted


loarie:
One small thing we could do to reduce the number of times IDers have to alter their ancestor disagreements to keep form "blocking" the flow of the community taxon is make it so that the community taxon calculation treats ancestor disagreements as non-disagreements if all IDs of the disagreeing taxa have been withdrawn/deleted.

That would help in this situation https://www.inaturalist.org/observations/60257864
<img width="856" alt="Screenshot 2024-02-25 at 8 27 56 AM" src="https://github.com/inaturalist/inaturalist/assets/598026/9b2623df-ff78-491a-a598-6f1467d61d5b">

PR #4019, opened 2024-02-26
Ancestor disagreement change

loarie:
Addresses this issue
https://github.com/inaturalist/inaturalist/issues/4018
which was discussed with the community here 
https://www.inaturalist.org/posts/25514-clarifying-ancestor-disagreements#activity_comment_29df9130-3fa0-49bb-84c3-d6666ca08b05

It is a minor tweak to how the Community Taxon is calculated that treats Ancestor Disagreements as Ancestor Non-Disagreements when the taxon being disagreed to (previous_observation_taxon) is no longer represented in IDs or as the ancestor of IDs (because they were withdrawn or deleted)

### DIFF
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -1837,6 +1837,9 @@ class Observation < ApplicationRecord
           if (
             i.disagreement? &&
             i.previous_observation_taxon &&
+            # if all ids with the previous_observation_taxon have been withdrawn, treat
+            # the ancestor id as a non-disagreement
+            ( working_idents.map( &:taxon_id ).include? i.previous_observation_taxon.id ) &&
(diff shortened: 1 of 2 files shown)


kueda (changes requested):
I don't disagree with the intent here, but you need tests, preferably under the [Observation community_taxon](https://github.com/inaturalist/inaturalist/blob/main/spec/models/observation_spec.rb#L924) section. At a minimum I want to see a test that the ancestor disagreement does not count when all prior descendant IDs have been withdraw, AND one that shows that the ancestor disagreement *does* count when some but not all prior descendant IDs have been withdrawn.

kueda (approved):

PR #4020, opened 2024-02-27
CICD / Docker image

sylvain-morin:
Create new config files using ENV variables (injected by docker compose or kubernetes manifest)

Update (site_config.rb) the way we load them to get ENV variables replacement

Rework the gitlab CI files to have 3 workflows:
- main: build the main branch and push 2 docker images, tagged "latest" and using the git commit ID
- dev: build any dev branch and push a docker image, tagged using the (sanitized) branch name
- pr: build any PR, running only the test (no docker image creation/push)


.github/workflows/CICD-pr.yml
```diff
@@ -0,0 +1,8 @@
+name: Rails CI/CD Pull Request
+
+on:
+  pull_request:
+
+jobs:
+  build-and-test:
+    uses: ./.github/workflows/CI-build-test.yml
```
kueda, line 8: Do we want to notify Slack about PR test runs?

config/config.docker.yml
```diff
@@ -1,169 +1,170 @@
...
+        - interia.pl
+        - o2.pl
+        - wordpressmails.com
+        - hotspotmails.com
+        - proconzon.com
+        - postfreeth.com
+        - waymymails.com
+        - jelly-life.com
+        - jokermails.com
+        - yourmailhost.com
```
kueda, line 160: Should these be checked in? Currently these are semi-secret, no?

kueda (commented):
Again, I don't have enough Docker expertise to help on that front. Checking in some stuff that used to be handle privately by Chef seems a bit dubious, but I'm ok with it if Patrick is.

.github/workflows/CICD-pr.yml
sylvain-morin: This workflow is rather for external devs who will fork our repo. If they create PR, it will run the tests but that's all.
I don't think we want to receive slack notifications for these PRs.
For a staff dev, we are commiting in our repo, so the branch workflow will run and we will get slack notifications

config/config.docker.yml
sylvain-morin: You are probably right. I will remove this.

.github/workflows/CICD-pr.yml
pleary: I think it would be useful to get slack notifications for PRs, both from external devs and internally. There could be conflicts with what we developed in a branch when it comes time to merge into main and the PR might fail when the branch did not

pleary (approved):
other than the minor comment about Slack notifications for PRs, with the recent `config/config.docker.yml` changes this looks good to me

sylvain-morin:
I removed all the hardcoded values from the `config.docker.yml `and added the Slack notification for PRs.

PR #4026, opened 2024-03-01
Wrong value for logged_in field in Rails logs

sylvain-morin:
I removed this initialization to false in `payload_from_request`, because otherwise it overwrites the setting done here: https://github.com/inaturalist/inaturalist/blob/main/app/controllers/application_controller.rb#L787

### DIFF
--- a/config/initializers/logstasher.rb
+++ b/config/initializers/logstasher.rb
@@ -83,6 +83,8 @@ module Logstasher
     if hash.key?( :request )
       if hash[:request].is_a?( ActionDispatch::Request )
         hash.merge!( Logstasher.payload_from_request( hash[:request] ) )
+        # set default value to logged_in if no value was set
+        hash[:logged_in] = false if hash[:logged_in].nil?
       end
       hash.delete( :request )
     end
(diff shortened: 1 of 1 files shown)


config/initializers/logstasher.rb
```diff
@@ -101,6 +99,8 @@ def self.replace_known_types!( hash )
       end
       hash.delete( :user )
     end
+    # set default value to logged_in if no value was set
+    hash[:logged_in] = false if hash[:logged_in].nil?
```
pleary, line 103: My recommendation would be to put this line inside the conditional above `if hash[:request].is_a?( ActionDispatch::Request )`, under `hash.merge!( Logstasher.payload_from_request( hash[:request] ) )`. That would most closely fix the existing behavior. Other logging events (delayed jobs, exceptions, etc) call `replace_known_types!` and those logs did not previously include a `logged_in` attribute. With this PR as submitted, they would begin including a `logged_in` attribute. The goal is to modify behavior that was coming from `payload_from_request`, so best to limit the fix to where that is used.

With that change, this would resolve the issue and be OK to merge.

pleary (changes requested):

config/initializers/logstasher.rb
sylvain-morin: Indeed, it will make less behavior change.

pleary (approved):

PR #4033, opened 2024-03-05
Login without email confirmation

### LINKED ISSUE #4030: Allow unconfirmed access for at least the next few years
As of Jan 1, 2023, after creating accounts new users must confirm emails before logging in. 

Make it so that new users after creating accounts are logged in and don't need to confirm emails along the way

loarie:
closes #4030

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -405,9 +405,6 @@ class User < ApplicationRecord
     child? && UserParent.where( "user_id = ? AND donorbox_donor_id IS NULL", id ).exists?
   end
 
-  EMAIL_CONFIRMATION_RELEASE_DATE = Date.parse( "2022-12-14" )
-  EMAIL_CONFIRMATION_REQUIREMENT_DATETIME = DateTime.parse( "2025-01-15 12:00" )
-
   # Override of method from devise to implement some custom restrictions like
   # parent/child permission and gradual confirmation requirement rollout
   def active_for_authentication?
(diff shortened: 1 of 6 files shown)


config/initializers/devise.rb
```diff
@@ -102,11 +102,11 @@
...
-  # Disallow unconfirmed access
-  # Note: in spring of 2022 we are gradually requiring email confirmation by
-  # allowing existing users some time to sign in and manually send themselves
-  # a confirmation email. This is managed in User#active_for_authentication?
-  config.allow_unconfirmed_access_for = 0
+  # Allow unconfirmed access
+  # Note: in spring of 2022 we encouraging email confirmation and also required
+  # new user to have completed email confirmation before logging in. In spring
+  # 2024 we began once again allowing users to login before confirming their email.
+  config.allow_unconfirmed_access_for = 20.years
```
kueda, line 109: "we encouraging" is a typo. 20 years is basically forever, not "a few years". Is that really the intent?

spec/controllers/oauth_tokens_controller_spec.rb
```diff
@@ -48,12 +48,9 @@
...
   describe "for an unconfirmed user" do
     let(:user) { create :user, :as_unconfirmed }
-    it "should return a conventional spec-compliant error response" do
+    it "should not return a conventional spec-compliant error response" do
       post :create, format: :json, params: default_params_for_strategy
-      expect( response.code ).to eq "400"
-      json = JSON.parse( response.body )
-      expect( json["error"] ).to eq "invalid_grant"
-      expect( json["error_description"] ).not_to be_blank
+      expect( response.code ).to eq "200"
```
kueda, line 53: We still want to issue errors for unconfirmed users after the grace period, which ends on `User::EMAIL_CONFIRMATION_REQUIREMENT_DATETIME`, so we still want a test like this. Instead of inverting this test like you've done, I would change the describe block to something like "for an unconfirmed user after the grace period" and try to mock Time.now or `allowed_unconfirmed_grace_period?` so the test can assess what happens after the grace period expires.

Then I would make different describe block that includes the test you have here to test that unconfirmed users *within* the grace period do receive successful responses to their oauth token requests.

spec/controllers/provider_oauth_controller_spec.rb
```diff
@@ -42,13 +42,13 @@
...
+      it "should return a token for a new user" do
         expect( User.find_by_email( google_response[:email] ) ).to be_blank
         post :assertion, format: :json, params: assertion_params
-        expect( response ).not_to be_successful
+        expect( response ).to be_successful
         expect( User.find_by_email( google_response[:email] ) ).not_to be_confirmed
-        expect( JSON.parse( response.body )["access_token"] ).to be_blank
-        expect( JSON.parse( response.body )["error"] ).not_to be_blank
+        expect( JSON.parse( response.body )["access_token"] ).not_to be_blank
+        expect( JSON.parse( response.body )["error"] ).to be_blank
```
kueda, line 51: Same feedback here. When the grace period is over, 3rd party sign in should stop working.

kueda (changes requested):

loarie:
I removed the grace period code to simplify things. My intention is that this should now be indefinitely allowing unconfirmed accounts to log in. Do you see any issues with this approach?

app/models/user.rb
```diff
@@ -405,9 +405,6 @@ def child_without_permission?
     child? && UserParent.where( "user_id = ? AND donorbox_donor_id IS NULL", id ).exists?
   end
 
-  EMAIL_CONFIRMATION_RELEASE_DATE = Date.parse( "2022-12-14" )
-  EMAIL_CONFIRMATION_REQUIREMENT_DATETIME = DateTime.parse( "2025-01-15 12:00" )
-
```
kueda, line 410: If you're intent is to remove email confirmation requirements entirely, you need to audit the code to remove all references to these constants. Some of your spec failures are legit and related to this.

kueda (changes requested):
This fulfills the requirements laid out in https://github.com/inaturalist/inaturalist/issues/4030, but you have some lingering bugs reflected in the spec failures.

spec/controllers/oauth_tokens_controller_spec.rb
kueda: I guess the decision was to not have a grace period at all, so this isn't necessary.

kueda (approved):

PR #4055, opened 2024-03-19
Expansive terms

loarie:
added `term_id_or_unknown` annotation param


config/locales/en.yml
```diff
@@ -1329,6 +1333,7 @@ en:
     male: Evidence indicates that organism can produce sperm for use in sexual reproduction
     molt: Discarded skin or exoskeleton
     no_evidence_of_flowering: Media provides no evidence of reproductive structures
+    not_established: organism belongs to an unestablished population including released/escaped pets, hitch-hikers and vagrants.
```
pleary, line 1336: Pretty minor thing, but we've so far capitalized all term definitions and I think that's worth doing here too: ` not_established: Organism belongs...`

pleary (changes requested):
I request one minor change, otherwise OK with me to merge

config/locales/en.yml
loarie: fixed thanks

PR #4057, opened 2024-03-19
Allow `week` to be passed as a search param for `/observations`

frewsxcv:
Akin to the other date filtering parameters (`observed_on`, `year`, `month`, and `day`), this rounds out these options with the `week` option which is already part of the ES observation index.

The use case for this is for <https://naturalists.nyc/>, where we have a eBird-esque bar visual that shows observation frequency for that week, as seen here:

Screenshot 2024-03-19 at 2 22 11 PM

What we'd like is the ability to have each bar link to the iNaturalist observation page for the corresponding week across all years.

Thank you for your consideration!

### DIFF
--- a/app/controllers/observations_controller.rb
+++ b/app/controllers/observations_controller.rb
@@ -2140,6 +2140,7 @@ class ObservationsController < ApplicationController
     @observed_on = search_params[:observed_on]
     @observed_on_year = search_params[:observed_on_year]
     @observed_on_month = [ search_params[:observed_on_month] ].flatten.first
+    @observed_on_week = search_params[:observed_on_week]
     @observed_on_day = search_params[:observed_on_day]
     @ofv_params = search_params[:ofv_params]
     @site_uri = params[:site] unless params[:site].blank?
--- a/app/es_indices/observation_index.rb
+++ b/app/es_indices/observation_index.rb
@@ -619,6 +619,7 @@ class Observation < ApplicationRecord
     # params to search based on value
     [ { http_param: :rank, es_field: "taxon.rank" },
       { http_param: :observed_on_day, es_field: "observed_on_details.day" },
+      { http_param: :observed_on_week, es_field: "observed_on_details.week" },
       { http_param: :observed_on_month, es_field: "observed_on_details.month" },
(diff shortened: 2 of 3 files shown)


pleary:
Thanks @frewsxcv . We can merge this in, but I'm curious what kind of links you'll want to generate. You may be interested to link to the main observations search/explore page, e.g. https://www.inaturalist.org/observations?month=4&place_id=any&subview=map&taxon_id=204321 . If so, then these changes won't affect the results you'll see there. Rather the `week` parameter would need to be added to `INatAPIService::MAPPABLE_PARAMS` to affect that search page. Could you please provide an example link where you want to use the `week` parameter? Again we can merge this PR, but it will only affect older or lesser-used and deprecated endpoints, which may not be what you're hoping to do.

frewsxcv:
An example link would be https://www.inaturalist.org/observations?week=23&place_id=48&subview=map&taxon_id=204321

Indeed I thought these changes would make that possible but I must have misread the code.

Looking at `INatAPIService::MAPPABLE_PARAMS`, I see they get mapped to [hidden inputs](https://github.com/inaturalist/inaturalist/blob/4e00a2037ae52a1f13400e70d01de995dae4d452/app/views/observations/index.html.haml#L30-L31). Do those hidden values get sent to the InaturalistAPI service directly?

frewsxcv:
(Thanks for the review by the way)

pleary:
Yes the hidden inputs created from `INatAPIService::MAPPABLE_PARAMS` allow those values to be sent to the API even though there won't be user-visible form fields for them. This allows users to enter those parameters in the URL and see results filtered by them independently of us updating the UI to accommodate them. I merged this in, then made commit 1eaf9da which should allow these for explore search and export. We'll deploy this later today. Thanks again!

frewsxcv:
@pleary Thank you for making the change! Looking at https://www.inaturalist.org/observations?place_id=674&taxon_id=55851&week=31, I would have expected this to return no results, but it's instead returning observations from this past week. Has the change been deployed yet? No worries if not

frewsxcv:
Nevermind it's working now! 🎉

PR #4063, opened 2024-03-26
Segmentation Statistics

sylvain-morin:
The aggregated segmentation data will be stored every month in this new table.
A method `get_viz_stats_for_day` will return a view of these data for the D3 Treemap.

In `production_schedule.rb`, the generation will be scheduled like that:
```
every :month, at: '5:00 pm', on: :first_day
  nice_rails_command "SegmentationStatistic.generate_stats_for_day(Time.now.utc - 1.day)"
end
```

### DIFF
--- a/app/models/segmentation_statistic.rb
+++ b/app/models/segmentation_statistic.rb
@@ -1,0 +1,602 @@
+# frozen_string_literal: true
+
+class SegmentationStatistic < ApplicationRecord
+  def self.generate_stats_for_day( at_time = Time.now, options = {} )
+    at_time = at_time.utc.end_of_day
+    if options[:force]
+      SegmentationStatistic.where( "DATE(created_at) = DATE(?)", at_time.utc ).delete_all
+    elsif stats_generated_for_day?( at_time )
+      return
+    end
+    sleep 1
+    SegmentationStatistic.create!(
+      data: generate_monthly_segmentation_metrics( at_time ),
+      created_at: at_time.beginning_of_day
+    )
+  end
+
+  def self.stats_generated_for_day?( at_time = Time.now )
+    SegmentationStatistic.where( "DATE(created_at) = DATE(?)", at_time.utc ).exists?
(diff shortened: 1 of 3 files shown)


loarie:
Looks good to me - I was able to use it to drive the visualization here #4065

app/models/segmentation_statistic.rb
```diff
@@ -0,0 +1,602 @@
...
+      }
+    }
+
+    obs_data = Observation.elastic_search(
+      size: 0,
+      filters: filter,
+      aggregate: agg
+    ).response.aggregations.user_id.buckets
+
+    obs_data.each do | u |
```
pleary, line 121: I find code more readable when variable names are more descriptive. Instead of `u` I might have used `user_counts_bucket` or just `bucket` since that's what it is - one bucket from an Elasticsearch aggregation response. Generally I prefer very descriptive variable names where its pretty clear what they are immediately without having to look back, even if that means very long variable names

app/models/segmentation_statistic.rb
```diff
@@ -0,0 +1,602 @@
...
+  def self.generate_monthly_segmentation_metrics( at_time = Time.now )
+    monthly_segmentation_data = generate_monthly_segmentation_data( at_time )
+    generate_main_metrics( monthly_segmentation_data )
+  end
+
+  # Generate all the users data for a month
+  def self.generate_monthly_segmentation_data( at_time = Time.now )
+    monthly_segmentation_data = {}
+    day = at_time
+    30.times do
```
pleary, line 39: To me this method is misnamed, or its not looking at the right time frame. This is generating data for the last 30 days from the given day, it does not generate data for a calendar month. It may be getting data from multiple months or part of a month

pleary (approved):
I left a few comments that aren't blockers. The one I'd most like to see addressed is the hard-coded oauth_application_ids. But that doesn't need to block merging this in

PR #4065, opened 2024-03-27
Segmentation statistics viz

loarie:
first pass at stats/user_statistics page on top of SegmentationStatistic model


sylvain-morin:
@loarie You don't need the get_viz_stats_for_day method in SegmentationStatistics, right?
The JSON data object I store in the DB is enough for the D3 TreeMap (with a few modification in your code)
If true, I will remove the code in SegmentationStatistics

loarie:
> @loarie You don't need the get_viz_stats_for_day method in SegmentationStatistics, right? The JSON data object I store in the DB is enough for the D3 TreeMap (with a few modification in your code) If true, I will remove the code in SegmentationStatistics

whoops I missed that you implemented that method - do you think that logic should be in the SegmentationStatistics model or the Stats controller

PR #4091, opened 2024-04-29
Observation accuracy experiment 4 updates

loarie:
Three changes:
1. use country rather than continent to match validators to sample observations to provide more localized matching
2. add disclaimer to message to not-reply and provide optional link to accompanying blog post for more context/feedback
3. add hand-holding text to message to help people accessing this message through the iNat Android app


loarie:
@pleary in an ideal world, we'd kick off Experiment #4 after CNC is over, but no rush so let me know when a good time to merge this in and kick off the experiment is

pleary (approved):

PR #4093, opened 2024-05-02
Seek common names update

kueda:
Mainly updates list of languages supported by Seek. Also Rubocop compliance.

### DIFF
--- a/tools/seek_common_names.rb
+++ b/tools/seek_common_names.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require "rubygems"
 require "optimist"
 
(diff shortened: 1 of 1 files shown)


tools/seek_common_names.rb
```diff
@@ -57,20 +90,19 @@
...
     Place.where( admin_level: Place::COUNTRY_LEVEL, code: region ).first
   end
-  memo
 end
 
 common_names = []
 taxon_ids = relevant_taxa.keys
-taxon_ids.in_groups_of( 1000, false ) do |group|
+taxon_ids.in_groups_of( 1000, false ) do | group |
   taxa = Taxon.where( id: group ).includes( :taxon_names )
```
pleary, line 103: Unrelated to changes in this PR, but I noticed this could be made a little more efficient by including `place_taxon_names` as well, e.g. `.includes( { taxon_names: :place_taxon_names } )`. I can add that after merging if you don't want to make that change here

pleary (approved):
Looks good to me

tools/seek_common_names.rb
kueda: Thanks. Added that, will merge.

PR #4106, opened 2024-05-20
fix: zoom taxon detail history chart to last 10 years

Maxhou00:
This a WIP and a workaround to make the history chart show the last 10 years data by default.

The zoom was enable but didn't work since the migration from C3.  I made it work back.

The option `axis.x.extent` use for creating a default zoom/subchart doesn't seems to work or I am missing something. I have created an [issue](https://github.com/naver/billboard.js/issues/3768) on billboard.js for it.

To workaround this option I forced a zoom on the chart after the initialization. It's working when the Taxon has recent observation but I think it won't work when the last observation is too old.

This solution is only a Frontend solution. 

It would probably be better that the server is sending only the 10 years data to the client. Since I don't know how it was working before I preferred to let all the data available to the client.

### DIFF
--- a/app/webpack/taxa/show/components/charts.jsx
+++ b/app/webpack/taxa/show/components/charts.jsx
@@ -2,9 +2,8 @@ import React from "react";
 import PropTypes from "prop-types";
 import ReactDOM from "react-dom";
 import _ from "lodash";
-import bb, { areaSpline, spline } from "billboard.js";
+import bb, { areaSpline, spline, zoom } from "billboard.js";
 import { schemeCategory10 } from "d3";
-import moment from "moment";
 import { Modal } from "react-bootstrap";
 import { objectToComparable } from "../../../shared/util";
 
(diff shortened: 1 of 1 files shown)


kueda:
Hey @Maxhou00. It's been almost a year and it looks like that bug you filed with billboard.js has been fixed. However, when I try to run your branch, I don't see the chart getting zoomed to a 10 year window. Is this something you still want to work on or should I close this?

Maxhou00:
Hello @kueda ,

Yes I was gonna working on it after I saw the fix on billboard.js but got busy. I probably sure that I will have time to work on it during the month (february). If it's ok for you.

kueda:
Sure, sounds good.

Maxhou00:
Hello @kueda,

The issue in billboard wasn't solve as I thought it would be. So there is no need to update billboardJS. I have added a default zoom that is now working.

Also, a personal opinion, drag zoom seems easier and more smooth to use than the default zoom (the wheel)

kueda (approved):
This is working for me. In the future, please try and abide by our eslint rules for all new code.

Issue #3758, opened 2023-05-22
Add pagination to content flag pages


tiwane:
From https://forum.inaturalist.org/t/not-all-flags-shown-on-a-taxon/41674

We currently only show the 30 most recent flags for a taxon and there's no UI to see older flags for that taxon. For example, [_Salvia_](https://www.inaturalist.org/taxa/50644-Salvia) shows that there's an unresolved flag:

<img width="362" alt="image" src="https://github.com/inaturalist/inaturalist/assets/15896181/566a5269-5696-4d84-b745-e781df0adb87">

But if you click on it, you're [taken to the 30 most recent flags](https://www.inaturalist.org/taxa/50644/flags) and have no way to find that unresolved flag via the UI.

We should add simple pagination functionality to the bottom of these pages taxon flag pages (eg https://www.inaturalist.org/taxa/50644/flags). It should be added to other content flag pages as well (eg journals, comments, IDs, projects observations, etc) but taxa are like the content that will accumulate the most flags.

farhadjaman:
![image](https://github.com/inaturalist/inaturalist/assets/89217255/53cad644-df9b-4906-8ebc-f129a0385ecb)
the flag button is not showing here

farhadjaman:
hey there i would like to resolve this issue.If you kindly assigned it to me,I will start.

pleary:
Fix has been merged and deployed. Thanks @Maxhou00!

PR #4110, opened 2024-05-24
#3758: Add pagination on flags

Maxhou00:
Just added the `will_paginate` helpers to the correct view.

Also took the opportunity to clean just a bit. 
`@unresolved` and `@resolved` were not use at all in this part of the code and were loading for nothing.


pleary:
Thanks!

Issue #4099, opened 2024-05-10
Assign Daily Active User Model category to users


sylvain-morin:
We want a new table (e.g. user_daily_active_category):
* user_id
* todays_category
* yesterdays_category
* updated_at
* created_at

This script will populate these values, adding new users as necessary. User associations should be remove automatically via rails model associations. Maybe populated in SiteStatistics, or via a separate script. Inactive users will not be included in this table. If a user goes from active to inactive, they should still have a row with todays_category = "inactive" and yesterdays catgegory="active"

PR #4120, opened 2024-06-06
First observation update

loarie:
Updated the first observation email experiment so that its more consistent with the engagement strategy and metics

### DIFF
--- a/app/models/segmentation_statistic.rb
+++ b/app/models/segmentation_statistic.rb
@@ -19,6 +19,19 @@ class SegmentationStatistic < ApplicationRecord
     SegmentationStatistic.where( "DATE(created_at) = DATE(?)", at_time.utc ).exists?
   end
 
+  def self.generate_segmentation_int( start_date, end_date, use_database: false )
+    segmentation_data = {}
+    current_date = end_date
+    num_days = ( end_date.to_date - start_date.to_date ).to_i
+    num_days.times do
+      puts "Processing #{current_date}..."
+      users_from_kibana_data( current_date, segmentation_data )
+      current_date -= 1.day
+    end
+    users_from_db( end_date, segmentation_data ) if use_database
+    segmentation_data
+  end
+
   # private
 
   NUMBER_OF_DAYS = 30
--- a/app/views/emailer/captive_observation.html.haml
+++ b/app/views/emailer/captive_observation.html.haml
@@ -8,12 +8,11 @@
 
 %p= "Hello, #{@user.published_name}!"
 
+%p= "Thank you for  posting this  observation to #{user_site.name}!"
+%p= "While #{user_site.name} is a great place to learn about captive animals and cultivated plants, did you know that observations of <b>wild organisms</b> can become <b>Research Grade</b> observations shared with scientists and represent an important source of biodiversity data for science and conservation?"
 %p
-  = "Congratulations on posting "
-  = link_to "your first observation", observation_url( @observation )
-  = " to #{user_site.name}!"
-%p= "While #{user_site.name} is a great place to learn about captive animals and cultivated plants, did you know that observations of wild organisms can become  Research Grade observations shared with scientists and represent an important source of biodiversity data for science and conservation?"
-%p= "Will you try observing a wild species to share with us? Here are some wild plants we think you can find nearby:"
+  = link_to "Are you ready to make your first observation?", user_site.url
+  = " Here are some wild plants we think you can find nearby:"
 %tr
   %td{ valign: "top", class: "bodyContent", "mc:edit": "body_content00", style: "color: #505050; font-family: Helvetica; font-size: 16px; line-height: 150%; padding-top: 20px; padding-right: 20px; padding-bottom: 20px; padding-left: 20px; text-align: left;" }
     %table{ border: "0", cellpadding: "0", cellspacing: "0", width: "100%", style: "border-collapse: collapse !important;" }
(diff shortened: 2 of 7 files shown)


loarie:
@pleary - this is just a replacement of an existing tool and some existing emailer methods/views that makes what we were already experimenting with a bit more sophisticated. It stops short of storing the data in the database by design, but thats where this will need to go eventually.

The one weird thing you might want to comment on is that its leveraging some of the site statistics stuff (and adds a new method to that https://github.com/inaturalist/inaturalist/pull/4120/files#diff-07b738825d42ad23db1bde6fb0af1131be9951098593eff1a9dcdb3d593243a7R22) to find active users.

thanks for taking a look - if possible would love to merge and deploy by this weekend so those emailer methods/views are in production as we keep iterating/evolving the a/b test in the tool around this

app/views/emailer/captive_observation.html.haml
```diff
@@ -0,0 +1,28 @@
...
+- user_site = @user.site || Site.default
+
+- if @observation.photos? && @observation.appropriate?
+  %div{ style: "float: right; width: 118px; margin: 10px; text-align: center;" }
+    = link_to observation_url( @observation ) do
+      %div{ style: "width:110px; height:110px; overflow:hidden; border-radius:50%; margin: 10px auto; margin-right:10px;background-image:url('#{observation_image_url( @observation, size: "medium" )}'); background-size:cover; background-position:center; border: 4px solid #74AC00;" }
+
+%p= "Hello, #{@user.published_name}!"
+
+%p= "Thank you for  posting this  observation to #{user_site.name}!"
```
pleary, line 11: there's some extra spaces in here

app/views/emailer/captive_observation.html.haml
```diff
@@ -0,0 +1,28 @@
...
+%p= "While #{user_site.name} is a great place to learn about captive animals and cultivated plants, did you know that observations of wild organisms represent an important source of biodiversity data for science and conservation?"
+%p
+  = link_to "Are you ready to make your first wild observation?", user_site.url
+  = " Here are some wild plants you might be able to  find nearby:"
```
pleary, line 15: double space in here

tools/first_observation_experiment.rb
```diff
@@ -13,13 +15,135 @@
...
+
+  if captives
+    filter += [
+      { exists: { field: "observed_on" } },
+      { exists: { field: "location" } },
+      {
+        bool: {
+          should: [
+            { term: { captive: true } },
+            { term: { "taxon.id": 43_584 } },
```
pleary, line 82: would be good to look these taxa up dynamically rather than hard code an ID. Maybe we'd want a Taxon::HUMAN constant similar to the Taxon::LIFE constant we already have, or at least get the ID from a query? This also looks like it would return observations that are captive OR genus Homo OR species Homo sapiens. Just want to make sure that's what you intended

app/models/segmentation_statistic.rb
```diff
@@ -19,6 +19,19 @@ def self.stats_generated_for_day?( at_time = Time.now )
     SegmentationStatistic.where( "DATE(created_at) = DATE(?)", at_time.utc ).exists?
   end
 
+  def self.generate_segmentation_int( start_date, end_date, use_database: false )
```
pleary, line 22: What does "int" mean here? If that's an abbreviation for something can we please use the full word so the meaning of this method is clear?

tools/first_observation_experiment.rb
```diff
@@ -13,13 +15,135 @@
 
 where [options] are:
 EOS
-  opt :user_data_path, "Path to the user data file.", type: :string, short: "-t"
+  opt :cohort_data_path, "Path to the cohort data file.", type: :string, short: "-t"
+end
+
+def get_obs( day, users )
```
pleary, line 21: This "tool" is very complicated with a handful of methods and then a few hundred lines of basically a "main" method. It doesn't have to happen now, but eventually it would be great to create a class for all of this stuff and break up the code that isn't wrapped in methods now, into smaller logical methods. That way it could be tested, and it would be easier to follow. Then the tool can instantiate the class if it needs instances, or call static methods of the class with readable method names, which would make it easier to understand what the tool is doing.

tools/first_observation_experiment.rb
```diff
@@ -33,136 +157,275 @@ def quality_metric_observation_ids( observation_ids, metrics )
...
-    longitude: geoip_response.results.ll[1],
-    city: geoip_response.results.city,
-    group: group,
-    errors: errors,
-    set: set
-  }
+# now sort out current day
+start_time = Time.now
+end_date = current_day
+ad = SegmentationStatistic.generate_segmentation_int( end_date - 1.day, end_date, use_database: true )
```
pleary, line 223: Will this eventually get replaced when https://github.com/inaturalist/inaturalist/issues/4099 is implemented? I think this is OK, but FWIW I don't consider Kibana to be critical production infrastructure. We don't have the same uptime expectations, monitoring, or backups for it, so keep that in mind when deciding to query it

pleary (approved):
I left a few suggestions - nothing that needs to block release

tools/first_observation_experiment.rb
loarie: yes - the plan is to experiment with this for a few weeks and if it looks like its succeeding at driving retention or yielding useful data towards that goal, then move this tool into the main codebase and also replace the CSV file with some database tables.

tools/first_observation_experiment.rb
loarie: this is unrelated to #4099 as it deals with first week retention - but the plan is to move it into the infrastructure if its successful at increasing retention or yielding useful data towards that goal.

Issue #4086, opened 2024-04-23
Add optional UUID column to Export page


tiwane:
See https://forum.inaturalist.org/t/add-uuid-to-observations-export-csv/49921

Allow users to export UUIDs as part of the CSV export by adding an optional UUID column to the Basic columns section at https://www.inaturalist.org/observations/export

PR #4127, opened 2024-06-11
feat: :sparkles: add new default export field uuid

HannesOberreiter:
Closes https://github.com/inaturalist/inaturalist/issues/4086

In this update, I have added the UUID to the basic columns for the export, positioning it as the second column. This placement is consistent with the default basic export, and it aligns with the position of the ID column, given that the list is not in alphabetical order.

~~The spec test did report some errors, but I believe it is not related to the update, as they are all some geoprivacy problems.~~ (only local issue as the CI did not complain)
`bundle exec rspec spec/models/observations_Export_flow_task_spec.rb`

interface

generated csv

### DIFF
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -199,6 +199,7 @@ class Observation < ApplicationRecord
   ]
   BASIC_COLUMNS = [
     "id",
+    "uuid",
     "observed_on_string",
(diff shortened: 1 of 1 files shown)


pleary:
Thanks!

PR #4130, opened 2024-06-18
not-in-list-id-explore-param & rubocop appeasement

loarie:
added not-in-list-id to explore params and also did some rubocop cleanup on lib/inat_api_service.rb

### DIFF
--- a/app/assets/javascripts/inaturalist/map3.js.erb
+++ b/app/assets/javascripts/inaturalist/map3.js.erb
@@ -11,7 +11,7 @@ var loadMap3 = function(){
 
 
 // requires GoogleMap classes
-if (typeof(google) == 'undefined' || typeof(google.maps) == 'undefined') {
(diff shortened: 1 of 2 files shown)


pleary (changes requested):
Looks good to me. We need to do one additional step right now which is to make any modification to the map3.js file that invokes MAPPABLE_PARAMS (see https://github.com/inaturalist/inaturalist/blob/not-in-list-id-explore-param/app/assets/javascripts/inaturalist/map3.js.erb#L428). Our asset compilation will process files that have changed, and it bases "changes" on the file directly, not on what it looks like when the ERB template code is fully processed.

So if we don't make some modification to map3.js, it will not be reprocessed and the previously generated version will continue to be served, and the new API parameter will not be included. The change could be anything - adding or removing a space or anything that modifies the file contents.

PR #4139, opened 2024-06-21
First week cohort model

loarie:
Moved logic from tool which uses CSV as datastore over to model


loarie:
@pleary I made that change as discussed so ready to review at your convenience

db/migrate/20240618044707_create_cohort_lifecycles.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class CreateCohortLifecycles < ActiveRecord::Migration[6.1]
+  def change
+    create_table :cohort_lifecycles do | t |
+      t.date :cohort, default: nil
+      t.references :user, null: false, foreign_key: true
```
pleary, line 7: We don't current use foreign key constraints in the DB. While this may be a good practice, I'm not sure we should start here and have this be the only use. I would however add `has_one :cohort_lifecycle, dependent: :delete` to the User model so these will get cleaned up if a user deletes their account

db/structure.sql
```diff
@@ -10718,6 +10794,7 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20240124214436'),
 ('20240222032444'),
 ('20240326135332'),
-('20240429211140');
+('20240429211140'),
+('20240618044707');
```
pleary, line 10798: I just want to point out there's a merge conflict with structure.sql which likely has to do with the semicolon here since new migrations have since been committed to main

app/models/cohort_lifecycle.rb
```diff
@@ -0,0 +1,366 @@
...
+            ( !observation.photos? && !observation.sounds? ) || observation.human? ||
+            !observation.quality_metrics_pass?
+
+          missing_fields = ["georeferenced", "observed_on"].reject {| field | observation.public_send( "#{field}?" ) }
+          missing_fields << "evidence" unless observation.photos? || observation.sounds?
+
+          if missing_fields.any?
+            errors.concat( missing_fields.map {| field | error_key.fetch( field, field ) } )
+          elsif ["recent", "evidence", "location", "date"].
+              any? {| field | quality_metric_observation_ids( [observation.id], field ).count == 1 }
```
pleary, line 153: Seems like this method is no longer defined in the context of this code. I see it as an instance method in ObservationAccuracyExperiment. If that's where its singularly defined, maybe that could be turned into a static method, calls updated accordingly in the ObservationAccuracyExperiment model, and then it could be referenced here statically?

db/migrate/20240618044707_create_cohort_lifecycles.rb
```diff
@@ -0,0 +1,26 @@
...
+      t.boolean :retention, default: nil
+      t.string :observer_appeal_intervention_group, default: nil
+      t.string :first_observation_intervention_group, default: nil
+      t.string :error_intervention_group, default: nil
+      t.string :captive_intervention_group, default: nil
+      t.string :needs_id_intervention_group, default: nil
+
+      t.timestamps
+    end
+  end
```
pleary, line 25: It looks like CohortLifecycles is often looked up by `cohort` or `cohort` plus `user_id`. Might be worthwhile adding an index at least on `cohort`. If you want to enforce that combination be unique, then maybe a unique index on both with `cohort` being the first column since that's most often queried, e.g `add_index : cohort_lifecycles, [:cohort, :user_id], unique: true`

pleary (changes requested):
I left a few comments/suggestions. The structure.sql conflict will need to be fixed as will the (what I believe is a) missing method `quality_metric_observation_ids`. I'd also recommend removing the `foreign_key: true` from the migration. The index is a nice-to-have, but we'll likely see the benefits of it as the table grows.

Aside from those things I'm not noticing other issues. Thanks for splitting this up. It makes it easier to follow and with the data self-contained in the database, easier to maintain in the long run.

loarie:
thanks @pleary I made those changes

pleary (approved):

PR #4141, opened 2024-06-24
Assign Daily Active User Model category to users

sylvain-morin:


pleary (changes requested):
Overall I think this looks good. I think it would be good to update to the unique index with the uniqueness validation on user_id, as well as the specified dependency and deletion in the User model. The changes I recommended for `update_user_daily_category` aren't necessary, but could lead to nice performance improvements for creating/updating those records

PR #4147, opened 2024-07-01
Build info

sylvain-morin:


sylvain-morin:
Will update inaturalistjs in package.json once intaturalist.js is merged

app/views/admin/app_build_info.html.haml
```diff
@@ -0,0 +1,52 @@
+<h1>Build Information</h1>
```
pleary, line 1: As this is using HAML, this can be written `%h1 Build Information`. The `h3` headers can be written similarly. Very minor - but no newline at the end of this file

app/controllers/build_info_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+      git_branch: ENV["GIT_BRANCH"],
+      git_commit: ENV["GIT_COMMIT"],
+      image_tag: ENV["IMAGE_TAG"],
+      build_date: ENV["BUILD_DATE"]
+    }
+    render json: build_info
+  end
+
+  def app_build_info
+    @app_build_info_json = INatAPIService.get(
```
pleary, line 18: It's not a problem to, but its also not necessary to use an instance variable here by prefixing `app_build_info_json` with `@`. As this is currently written, `app_build_info_json` is not used outside this method as there is no template so it doesn't need `@`.

That said, there is an alternative way to handle this and the below method. `app_build_info` and `app_build_info_html` do the same thing, but render different templates. Rails can define different outputs for different formats in the same method. You could alternatively define a single method, e.g.
```
  def app_build_info
    @app_build_info_json = INatAPIService.get(
      "/app_build_info",
      { authenticate: current_user },
      { endpoint: INatAPIService::ENDPOINT.sub( "v1", "v2" ) }
    )
    respond_to do | format |
      format.json { render json: @app_build_info_json }
      format.html { render template: "admin/app_build_info", layout: "admin" }
    end
  end
```

Then in routes.rb call the same method from the different paths, with a default format of :json for the json version, e.g.
```
  get "build_info", to: "build_info#index"
  get "app_build_info", to: "build_info#app_build_info", defaults: { format: :json }
  get "admin/app_build_info", to: "build_info#app_build_info", as: "admin_app_build_info"
```

Is there a reason to have the JSON version at `/app_build_info` but the HTML version at `/admin/app_build_info`? If the JSON version didn't need a different path, you could leave that route off entirely and both formats could be accessed at `/admin/app_build_info.json` and `admin/app_build_info.json`.

Also, do you have a linter installed displaying rubocop rules? If not I'd recommend it. There are some spacing issues in this file it would have complained about.

config/routes.rb
```diff
@@ -607,6 +607,10 @@
     end
   end
 
+  get "build_info", to: "build_info#index"
+  get "app_build_info", to: "build_info#app_build_info"
```
pleary, line 611: See my comment above in the controller. Unless there's a need for `/app_build_info` to have a different path than `/admin/app_build_info`, they can be combined into the admin version to simplify things a little

pleary (approved):
Left some comments on potential ways to improve, but they don't need to be blockers

Issue #4153, opened 2024-07-04
New flag not created, old flag unresolved if flag reason matches user's previous flag for taxon


tiwane:
From: https://forum.inaturalist.org/t/flags-mysteriously-unresolving/52650/

1. Find a taxon which you have already flagged, and make sure that flag has been resolved.
2. Flag that taxon again, and make sure the "Reason" field matches the "Reason" field from your resolved flag.
3. New flag is not saved, old flag gets unresolved.

pleary:
resolved by #4157

PR #4157, opened 2024-07-09
[Bug Fix] New flag not created, old flag unresolved if flag reason matches user's previous flag for taxon

rzlim08:
This PR fixes the issue #4153  
* The code explicitly looked for flags that matched the reason/user/taxon and unresolved the flag. This PR removes that code path. 
* This is my first PR in this repo, so let me know if I need to do anything else!

### DIFF
--- a/app/controllers/flags_controller.rb
+++ b/app/controllers/flags_controller.rb
@@ -149,15 +149,11 @@ class FlagsController < ApplicationController
     end
 
     @create_comment = false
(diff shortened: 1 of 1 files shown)


pleary:
Thanks @rzlim08! This looks great. In theory if we had complete test coverage a test might have failed since a branch with specific behavior was removed. But that case was covered by tests, and since you're not introducing new functionality no new tests are needed. So nothing else is needed for this PR. I merged it in and we'll deploy within the next day or so.

Issue #4152, opened 2024-07-03
Remove sliding transition between photos in Identify


abhasinat:
Right now, when a user switches between photos inside of identify, we show a slide effect to move between photos. Several users in the [forum](https://forum.inaturalist.org/t/remove-sliding-transition-between-photos-in-identify-accessibilty-for-people-with-motion-sensitivity/52954/1) would prefer if this animation was removed, and instead photos were loaded instantly, just as they are when you move between observations.

@tiwane raised concern that this may cause visual confusion between moving between observations vs moving between photos in a single observation. However, I think the square indicator around the photos beneath the obs photos in identify is a good indicator of movement between photos in a single observation, so i don't think it should be an issue.

PR #4158, opened 2024-07-09
feat: :wheelchair: disable gallery animation to reduce motion sickness

HannesOberreiter:
Closes https://github.com/inaturalist/inaturalist/issues/4152

This pull request disables the transition effect in the identify modal gallery (and suggestions gallery), making the behavior consistent with the PhotoGallery in the Observation View. This change addresses user feedback requesting the removal of the slide animation when switching between photos.

The transition effect removal can be observed in the provided video.
Note: The suggestion tab functionality cannot be tested locally due to lack of access to the vision API.

[disabled-transitions.webm](https://github.com/inaturalist/inaturalist/assets/16878981/7bc06330-9dbc-42d6-80c7-c745b39ebb20)

Additional note: The same modal is used for compare taxa.

### DIFF
--- a/app/webpack/observations/identify/components/observation_modal.jsx
+++ b/app/webpack/observations/identify/components/observation_modal.jsx
@@ -265,6 +265,7 @@ class ObservationModal extends React.Component {
           <ZoomableImageGallery
             key={`map-for-${observation.id}`}
             items={images}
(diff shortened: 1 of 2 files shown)


pleary:
Thanks!

Issue #4146, opened 2024-06-28
Observation Detail: Compare button only filters by taxon of the first use of Compare


tiwane:
From https://forum.inaturalist.org/t/observation-compare-always-shows-first-species-compared/52714

1. Go to a observation with at least two different IDs

3. There are compare buttons on different IDs

4. Click on one of the Compare buttons. It filters by the taxon of the ID on which the Compare button appears.

5. Click on the other Compare button. It filters by the taxon of the ID of the first time you clicked on the Compare button. This happens regardless of taxon.

PR #4159, opened 2024-07-09
fix: :bug: fix stale compare modal due to caching

HannesOberreiter:
Closes https://github.com/inaturalist/inaturalist/issues/4146

This fixes the issue, where the compare observation modal failed to update the suggested taxon upon user interaction. The root cause was identified as unintended caching behavior, originally designed to preserve filter states in the identification modal.

Implemented a check to determine if the taxon object attribute has changed when a user clicks on a different compare button. This ensures that the compare observation modal updates the suggested taxon accordingly.

- Manual tests were conducted to verify that the compare observation modal now correctly updates the suggested taxon.
- Additional tests confirmed that the identification modal's caching and filter preservation functionalities remain unaffected by this change.

### DIFF
--- a/app/webpack/observations/identify/ducks/suggestions.js
+++ b/app/webpack/observations/identify/ducks/suggestions.js
@@ -81,12 +81,13 @@ export default function reducer(
       const { observation } = action;
       const isFeaturedObs = observation.uuid === state.query.featured_observation_uuid
         || observation.id === state.query.featured_observation_id;
+      const isTaxonChange = observation.taxon && state.query.taxon_id !== observation.taxon.id;
       newState.query = {
         source: state.query.source,
         order_by: state.query.order_by
       };
-      // If observation currently in query (tab change in modal), preserve taxon and place filters
-      if ( observation && isFeaturedObs ) {
+      // If observation currently in query (tab change in modal), preserve taxon and place filters. Except if a taxon change is detected.
+      if ( observation && isFeaturedObs && !isTaxonChange ) {
         newState.query = Object.assign( newState.query, {
           taxon: state.query.taxon,
           taxon_id: state.query.taxon_id,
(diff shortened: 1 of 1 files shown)


HannesOberreiter:
@pleary ah damn saw that I missed a console log is in the pull request. Can you revert the merge? Then I push an update.

pleary:
Hello @HannesOberreiter - I noticed that and thought that instead of a back-and-forth it would be faster for me to remove it in a separate commit, so I did so here https://github.com/inaturalist/inaturalist/commit/155a055291bef383a157645702fbfa71ff71fb88. In the future would you rather me request small changes like this before merging? If you're not using one, I'd recommend using a linter in your text editor. We define eslint rules for the webpack directory at https://github.com/inaturalist/inaturalist/blob/main/app/webpack/.eslintrc . It would have complained about the long lone and the `console.log` statement. There are surely lots of existing formatting errors, and we don't expect developers to fix existing problems. But ideally issues in new code, or lines that have been edited should be resolved.

Thanks for the PR!

HannesOberreiter:
First, I apologize for the inconsistent pull request.

> In the future would you rather me request small changes like this before merging?

It depends on what works best for you. In most cases, the pull request author still has the branch active, making these small changes easy to implement. Additionally, if you use squash merging strategy for pull requests, it results in a single commit, which is in my opinion preferable for a community-driven public repo with feature branches.

> We define eslint rules for the webpack

I apologize for the oversight; I disabled the linter for the iNaturalist repo in my VSCode settings a while ago and forgot about it. It might be helpful to add an `npm run lint` script, as I completely overlooked the presence of ESLint config in the project. In the long term, a pre-commit hook that enforces linting (e.g., using Husky with npx lint-staged) would be even more future-proof.

Cheers
Hannes

PR #4181, opened 2024-08-06
fix: Edited taxobox for Wikipedia entries

RichardLitt:
I found an issue today where taxobox uses old available fields for the taxobox. This issue has been known for some time - see this discussion, https://forum.inaturalist.org/t/fix-taxobox-parameters-in-create-this-page-on-wikipedia-prompt/18395/12. I've tried to make minimal efforts to remove those fields and to add others.

### DIFF
--- a/app/views/taxa/_description.html.haml
+++ b/app/views/taxa/_description.html.haml
@@ -94,17 +94,22 @@
         box = <<-WIKI
           {{Automatic taxobox
           | taxon = #{@taxon.name}
-          #{"| rank = #{Taxon::WIKIPEDIA_RANKS[@taxon.rank]}" unless @taxon.rank.blank?}
-          #{"| parent = #{parent_title}" if parent_title}
           #{"| image = #{image}" if image}
+          #{"| image_caption = #{image}" if image}
+          | authority =
+          | type_species =
+          | type_species_authority =
           }}
         WIKI
         if @taxon.species? && @taxon.genus
           box = <<-WIKI
             {{Speciesbox
-            | parent = #{@taxon.genus.name}
             | taxon = #{@taxon.name}
             #{"| image = #{image}" if image}
+            #{"| image_caption = #{image}" if image}
+            | authority =
+            | type_species =
+            | type_species_authority =
             }}
           WIKI
         end
--- a/app/views/taxa/_description.html.haml
+++ b/app/views/taxa/_description.html.haml
@@ -108,8 +108,6 @@
             #{"| image = #{image}" if image}
             #{"| image_caption =" if image}
             | authority =
-            | type_species =
-            | type_species_authority =
             }}
           WIKI
         end
--- a/app/views/taxa/_description.html.haml
+++ b/app/views/taxa/_description.html.haml
@@ -95,7 +95,7 @@
           {{Automatic taxobox
           | taxon = #{@taxon.name}
           #{"| image = #{image}" if image}
-          #{"| image_caption = #{image}" if image}
+          #{"| image_caption =" if image}
           | authority =
           | type_species =
           | type_species_authority =
@@ -106,7 +106,7 @@
             {{Speciesbox
             | taxon = #{@taxon.name}
             #{"| image = #{image}" if image}
-            #{"| image_caption = #{image}" if image}
+            #{"| image_caption =" if image}
             | authority =
             | type_species =
             | type_species_authority =


pleary:
Thanks for the pull-request @RichardLitt. The only bit about this I'm wondering about is `image_caption`. Is it recommended that it be set to the image file name, the same as `image`? I see in the [taxobox template documentation](https://en.wikipedia.org/wiki/Template:Taxobox#Images) says that:

> [image_caption] should be provided if it can convey any additional information about the image, such as the sex or life stage of the individual, the location where the picture was taken, the artist (if an engraving or other illustration), or (for higher taxa) the particular species depicted

Is the file name sufficient `additional information` that it be included? In my (somewhat uninformed as I am not a Wikipedia editor) opinion it is not. The editor could easily choose to copy the file name in there if they want. In my opinion it would be preferable to include the stub `image_caption` parameter with no value. What do you think? The rest looks OK me.

RichardLitt:
Ah! That was me misunderstanding Ruby. Not my native language. What I wanted to say was: if there is an image, then include image caption. If there isn't, don't. 

Something like this? 

`#{"| image = #{image}\n| image_caption = " if image}`

pleary:
Something like that could work. To keep spacing consistent, I think something like this might be better. It has an extra conditional but it is easier to understand at a glance, e.g.

```
        box = <<-WIKI
          {{Automatic taxobox
          | taxon = #{@taxon.name}
          #{"| image = #{image}" if image}
          #{"| image_caption =" if image}
          | authority =
          | type_species =
          | type_species_authority =
          }}
        WIKI
        if @taxon.species? && @taxon.genus
          box = <<-WIKI
            {{Speciesbox
            | taxon = #{@taxon.name}
            #{"| image = #{image}" if image}
            #{"| image_caption =" if image}
            | authority =
            | type_species =
            | type_species_authority =
            }}
          WIKI
        end
```
Essentially remove `#{image}` from `#{"| image_caption = #{image}" if image}`. If you want to make that change, I'll go ahead and merge this and deploy it later today

RichardLitt:
I went through and double-checked the [Taxobox](https://en.wikipedia.org/wiki/Template:Taxobox#Images) and [Speciesbox](https://en.wikipedia.org/wiki/Template:Speciesbox) params, and I think this is a better edit than we had before. I also worked in the image_caption, and I added the optional image_alt because I think it'll be useful to screenreaders.

pleary:
Unfortunately, this latest commit has a few issues. `@taxon.species` and `@taxon.genus` do not return strings, rather they return instances of Taxon. The reason the code was using `@taxon.species? && @taxon.genus` is it was checking the given taxon is at rank species and that the taxon has a genus taxon. By changing that to `@taxon.species && @taxon.genus` it is then checking if the taxon has a species taxon, which is true for all subspecific taxa as well as species. I don't know that we have a method that just returns the specific epithet alone.

Since `@taxon.genus` returns an instance of Taxon, `genus = #{@taxon.genus}` will not format correctly. It will appear as e.g. `genus = <Taxon 6: Genus Aramus>`. You could potentially write it as `genus = #{@taxon.genus.name}`. I'm not sure what edge cases we might hit for uncommon scenarios like hybrids, complexes, etc. In my opinion sticking with the original commit plus the change I recommended is the best way to go for now - leaving `taxon =` instead of using `name`, `genus`, and `species`. Wikipedia seems to do a good job of figuring it out and easy enough for an editor to split it if they want.

Can you please revert to using `@taxon.species? && @taxon.genus` for the proper conditional check, and not using `species = #{@taxon.species}` or `genus = #{@taxon.genus}` since those will not be rendered correctly?

RichardLitt:
Right! Thank you for being patient. 

I was concerned because `taxon` isn't a parameter in Wikipedia. `name` is, and that may work better - I am trying to minimize work on the part of Wikipedia editors. Also - note that there are two different types of taxobox here - a plain Automatic Taxobox, and a Speciesbox with its own parameters (hence `name`, `genus`, and `species` in one of the sections). I can revert it back - happy to, even - but I'm not sure it makes sense in the long run. 

Edge cases is a good point. Maybe we shouldn't be smart at all, and just remove the third conditional block.

RichardLitt:
I had of help from a Wikipedia editor [here](https://en.wikipedia.org/wiki/User_talk:Richlitt#c-Plantdrew-20240808213600-Taxonomy_templates). I think you're right - we can just use `taxon=`. This PR is now slimmer and does less, but does fix the original issue we had. 

Let me know what you think. Thank you!

RichardLitt:
Thank you! I think this is my first landed PR in iNat. Exciting. :)

PR #4191, opened 2024-08-20
OpenStreetMap base map

kueda:
* Adds the default OpenStreetMap tiles (from tiles.openstreetmap.org) as a base map option for all maps
* Added OSM attribution to all maps when OSM tiles shown
* Changes base map control to a dropdown by default

### DIFF
--- a/app/assets/javascripts/jquery/plugins/inat/taxonmap.js.erb
+++ b/app/assets/javascripts/jquery/plugins/inat/taxonmap.js.erb
@@ -68,19 +68,6 @@ inatTaxonMap.setup = function ( elt, opts ) {
   options.taxonLayers = options.taxonLayers || $( elt ).data( "taxon-layers" );
   options.mapTypeControl = ( options.mapTypeControl !== false && $( elt ).data( "map-type-control" ) !== true );
   options.mapTypeControlOptions = options.mapTypeControlOptions || $( elt ).data( "map-type-control-options" );
-  // eslint-disable-next-line prefer-object-spread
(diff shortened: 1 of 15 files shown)


app/assets/javascripts/jquery/plugins/inat/taxonmap.js.erb
```diff
@@ -69,6 +69,12 @@ inatTaxonMap.setup = function ( elt, opts ) {
   options.mapTypeControl = ( options.mapTypeControl !== false && $( elt ).data( "map-type-control" ) !== true );
   options.mapTypeControlOptions = options.mapTypeControlOptions || $( elt ).data( "map-type-control-options" );
   options.zoomControl = ( options.zoomControl !== false && $( elt ).data( "zoom-control" ) !== false );
+  options.zoomControlOptions = options.zoomControlOptions || $( elt ).data( "map-type-control-options" );
```
pleary, line 72: Should this be `$( elt ).data( "zoom-control-options" )` ?

pleary (approved):
This looks good to me. I left a comment about one minor potential fix. One other small issue is that on the observation edit page, the zoom controls are defaulting to the bottom-right causing map attribution to appear above them. We might want to add

```
    zoomControlOptions: {
      position: google.maps.ControlPosition.BLOCK_START_INLINE_START
    }
```

to `iNaturalist.Map.createMap` in https://github.com/inaturalist/inaturalist/blob/main/app/assets/javascripts/observations/edit.js#L12

PR #4217, opened 2024-09-14
fix: :bug: set correct link for identifier icon subtitle

### LINKED ISSUE #4216: Obs Detail: links under top identifiers usernames show number of IDs but link to user's observations.
From: https://forum.inaturalist.org/t/clicking-top-identifier-link-leads-to-their-observations-not-ids/55336

1. Go to: https://www.inaturalist.org/observations/241382144
2. Scroll down to top identifiers and you'll see something like this. Each number of under the username shows how many IDs of that taxon they've made. 
3. Mouseover any of those numbers and the link leads to their observations, eg https://www.inaturalist.org/observations?user_id=samfellows566&place_id=any&verifiable=any

That link should lead to a search of observations of that taxon they've add an ID to.

HannesOberreiter:
Closes #4216 

This update introduces a new optional parameter, `subtitleLinkOverwrite`, to the ` user_with_icon` React component. This parameter allows for overriding the default link behavior.

This parameter is used to update the link on the observation page for the top identifiers widget, specifically pointing it to, the identifications page eg. https://www.inaturalist.org/identifications?user_id=samfellows566&taxon_id=4345.

We could add a parameter for `for=others`, but I believe the count currently includes self-identifications as well.

Cheers
Hannes

### DIFF
--- a/app/webpack/observations/show/components/identifiers.jsx
+++ b/app/webpack/observations/show/components/identifiers.jsx
@@ -49,7 +49,7 @@ class Identifiers extends React.Component {
     } else {
       panelContents = identifiers.map( i => (
         <div className="identifier" key={`identifier-${i.user.id}`}>
-          <UserWithIcon user={i.user} subtitle={i.count} subtitleIconClass="icon-identification" />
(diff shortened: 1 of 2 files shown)


pleary:
Looks good to me. Yes the count on the observation page does include self-identifications, so no need for the `for=others` parameter. The count also is limited to research grade observations, but we currently don't have a filter for quality_grade on the identifications page. A potential future feature. But for now this will be an improvement over what the observation page is linking to right now, which isn't really relevant to that section. Thanks!

Issue #3644, opened 2023-02-08
Obs show: in the Top Identifier section, add a place filter


tiwane:
From https://forum.inaturalist.org/t/show-top-identifiers-by-continent-by-default/15310

Some identifiers are prolific, but only in a narrow geographic area. So they may be the top identifer of a family, but only becuase they have identified a lot of observations in the US and my not be familiar with taxa in that family outside of their region of interest. However, they're listed as a top identifier on obs show pages for observations made in entirely different parts of the world. They get requests for ID help but can't really help out because they're not familiar wtih that area. 

The request is that, for observations identified from family to genus, we add a place filter with a dropdown menu for the Top Identifiers section and by default set it to the continent in which the observation appears. So by default, we show the top identifiers of that taxon on that continent. If someone clicks on the menu, we show them all standard places in which the observation appears and they can select those places, which will then filter the top identifiers of that taxon for that place. 

For observations with private geoprivacy, we shouldn't filter top identifiers by location and instead just show global top identifiers like we currently do.

Hussell:
No need to restrict it to observations identified from family to genus, I think. At species or narrower it might, maybe, make sense to have the default be global, but being able to filter out non-local identifiers would still be of great value. In some cases there's enough regional variation in the appearances of a species that non-local identifiers would have difficulty recognizing that they're even looking at the same species.

On observations identified as wider taxonomic categories the need to recommend local identifiers by default seems to me to be even greater.

Not having any sort of location filter on private observations is something I hadn't thought about before, but seems obviously correct now that tiwane has brought it up. Putting a minimum size limit on the places which can be used to filter identifiers on obscured observations might also be necessary.

kfc35:
I will tackle this as my next task

PR #4218, opened 2024-09-14
fix: :bug: remove preferred_place_id from leader widget

HannesOberreiter:
If a user's state includes a `preferred_place_id`, it is currently appended to the top identifier and observer links on the taxa page. While this behavior could be relevant in some contexts, it may not always make sense, as the counts are global and not restricted to a specific place.

This pull request removes the `preferred_place_id` from the object when generating these links, ensuring the links remain global as intended.

## Example of the current link generated on the widget:

image

Cheers
Hannes

### DIFF
--- a/app/webpack/taxa/show/containers/top_identifier_container.js
+++ b/app/webpack/taxa/show/containers/top_identifier_container.js
@@ -19,6 +19,7 @@ function mapStateToProps( state ) {
   }
   const urlParams = defaultObservationParams( state );
   urlParams.view = "identifiers";
+  delete urlParams.preferred_place_id;
   return Object.assign( props, {
     name: leader.user.login,
     imageUrl: leader.user.icon_url,
--- a/app/webpack/taxa/show/containers/top_identifier_container.js
+++ b/app/webpack/taxa/show/containers/top_identifier_container.js
@@ -19,7 +19,7 @@ function mapStateToProps( state ) {
   }
   const urlParams = defaultObservationParams( state );
   urlParams.view = "identifiers";
-  delete urlParams.preferred_place_id;
+  urlParams.place_id = urlParams.place_id || "any";
   return Object.assign( props, {
     name: leader.user.login,
     imageUrl: leader.user.icon_url,
--- a/app/webpack/taxa/show/containers/top_observer_container.js
+++ b/app/webpack/taxa/show/containers/top_observer_container.js
@@ -19,6 +19,7 @@ function mapStateToProps( state ) {
   }
   const urlParams = defaultObservationParams( state );
   urlParams.view = "observers";
+  delete urlParams.preferred_place_id;
   return Object.assign( props, {
     name: leader.user.login,
     noContent: false,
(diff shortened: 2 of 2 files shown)


HannesOberreiter:
Found this semi related ticket: https://github.com/inaturalist/inaturalist/issues/3644

pleary:
Is there a particular bug this is meant to address? `preferred_place_id` is an API parameter distinct from `place_id`. Including `preferred_place_id` does not restrict results to a place, rather it only informs localization of common names. `place_id` on the other hand does restrict results to a place, and does not inform localization. While the change in this pull request won't cause any problems, it also shouldn't have any visible change at all since localization won't affect the response at all in this case.

If there is a bug this is meant to address, it would be good to have that documented somewhere. Again, I don't think this pull request will have any effect on what is presented on the taxon page, and if so, will not fix any outstanding issues.

HannesOberreiter:
Hi pleary,

Thanks for your response! I’ve looked into it further, and you’re right—preferred_place_id isn’t causing the issue. My apologies for the confusion. I initially thought that was the right parameter, but it seems the problem might be with my development environment (the map functionality isn't working properly locally, likely due to issues with the Google API setup).

What I’m trying to achieve is the following:

Example: https://www.inaturalist.org/taxa/41644-Ursus-maritimus

Clicking on the leaderboard currently takes me here:

image

However, what I’d like is to see the global leaderboard without any country filters applied.

As this country session filter seems to be related to ones session. There is probably no easy fix.

Anyway sorry for the unnecessary pull request!

Cheers
Hannes

HannesOberreiter:
Edit: Just came to my mind that we could add `place_id=0` to the query. 

Do you think that would make sense?

Cheers
Hannes

pleary:
The taxon page can filter by a place. This can happen when visiting one of the regional iNaturalist sites which can set a default place_id, e.g. https://www.inaturalist.ca/taxa/41644-Ursus-maritimus, or when a user chooses a place with the place filter (underneath the taxon search box in the upper-right). The only case that comes to mind to replicate the problem you are seeing is: perhaps you have set your default site to have a default place filter of Austria, and you have removed that default place filter from the taxon page, so when you go to explore that default filter is added back.

One change we could make to address that somewhat edge case is to add the parameter: `place_id=any` if there is no `place_id` parameter already in `defaultObservationParams`. That would ensure that explore applies no place filter when one wasn't already set. To be clear, there are many cases `place_id` will already be set and we want to preserve it, but if it is not set, it is indeed preferred to add a default of `place_id=any`. That would be a good improvement to make. Thanks!

HannesOberreiter:
Ah I see, never realised that the taxon page by itself has a region filter.

I changed it to overwrite the `place_id` to `any` if not set, but maybe the logic should be applied inside the `defaultObservationParams` function but that may cause too much havoc on other links.

Now the linked discussion makes even more sense for me. Wait for the merge, I think how I could solve the linked discussion problem over the next few days if I find free time, because depending on how we solve the linked discussion this change may be unnesecarry.

Cheers
Hannes

HannesOberreiter:
> Edit: Thinking about adding the preferred place id as default aktive filter when one is visiting the taxa page.

I decided not to add the default `place_id` on load for the top identifiers and observation calculation on the taxa page. Found no time to setup Google Cloud and Google API key for local integration, and local testing without easily setting locations is problematic.

Therefore, I think we can merge this pull request. It has a small footprint and, as you mentioned, is a minor improvement and one should anyway only make small changes per branch if possible.

Cheers
Hannes

pleary:
Great - thanks for this!

PR #4219, opened 2024-09-14
fix: :bug: remove loading spinner if result is empty

HannesOberreiter:
Currently, when visiting an observation tab (e.g., view=species, view=observers, or view=identifiers), if the result set is empty, the loading spinner continues to display indefinitely. For example: https://www.inaturalist.org/observations?place_id=143465&subview=map&taxon_id=41644&view=observers.

This patch removes the loading spinner when the result set is empty, improving the user experience.

We can use the corresponding `noXXXX` function, which correctly returns false while the search is ongoing. Alternatively, we could use the `searchingStopped` function directly, but that would require some refactoring to make it compatible with the infinite scroll and loading icon behavior.

Cheers
Hannes

### DIFF
--- a/app/assets/javascripts/ang/templates/observation_search/observers_table.html.haml
+++ b/app/assets/javascripts/ang/templates/observation_search/observers_table.html.haml
@@ -21,7 +21,9 @@
           {{ shared.numberWithCommas( u.observationCount ) }}
         %td{ :class => "{{ observersSort == '-speciesCount' ? 'sorting' : '' }}" }
           {{ shared.numberWithCommas( u.speciesCount ) }}
-  .spinner.ng-cloak{ "ng-show": "pagination.searching || observersPagination.searching" }
+  .spinner.ng-cloak{ "ng-show": "(pagination.searching || observersPagination.searching) && !noObservers( )" }
     %span.fa.fa-spin.fa-refresh
+    {{ pagination.searching }}
+    {{ observersPagination.searching }}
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/ang/templates/observation_search/observers_table.html.haml
```diff
@@ -21,7 +21,9 @@
           {{ shared.numberWithCommas( u.observationCount ) }}
         %td{ :class => "{{ observersSort == '-speciesCount' ? 'sorting' : '' }}" }
           {{ shared.numberWithCommas( u.speciesCount ) }}
-  .spinner.ng-cloak{ "ng-show": "pagination.searching || observersPagination.searching" }
+  .spinner.ng-cloak{ "ng-show": "(pagination.searching || observersPagination.searching) && !noObservers( )" }
     %span.fa.fa-spin.fa-refresh
+    {{ pagination.searching }}
```
pleary, line 26: These two lines should be removed

pleary (changes requested):
Thanks for making this fix! Looks like there are two lines in here that may have been added when debugging - in the observers tab. Those are resulting in some booleans being printed when searching. If you could remove those I'll go ahead and merge this in

app/assets/javascripts/ang/templates/observation_search/observers_table.html.haml
HannesOberreiter: Good catch, sorry about that.

PR #4220, opened 2024-09-15
fix: :bug: change order of icons and count

### LINKED ISSUE #4165: Explore: In Russian, "Research Grade" text messes up alignment in grid view
From https://forum.inaturalist.org/t/misleading-spacing-between-ids-and-observation-date-makes-time-calculation-appear-wrong-russian/53075

1. Change locale to Russian
2. Look up RG observations from 2 months ago, in Grid view
3. The number of IDs gets pushed into the relative date, so rather than "2 months ago" it says "22 months ago"

image

HannesOberreiter:
Closes #4165 

In the current Observation Grid view, long research quality names (such as translations) cause the count and age to overlap or get pushed together, affecting readability.

I updated the order of elements: The count now always appears before the icon to improve layout consistency and better differentiation between age and counts. Applied the same adjustment to the Table view for visual consistency across different views.

Alternative Consideration:
An alternative solution could involve enforcing a maximum length for research quality names to prevent layout issues.

## Example of new grid view, with extra long text for research grade

image

## Example of new table view, with extra long text for research grade

image

### DIFF
--- a/app/assets/javascripts/ang/templates/observation_search/results_grid.html.haml
+++ b/app/assets/javascripts/ang/templates/observation_search/results_grid.html.haml
@@ -15,14 +15,14 @@
             %span{:class => "quality_grade {{ o.quality_grade }}"}
               {{ o.qualityGrade( ) }}
             %span.identifications{"ng-show" => "o.identifications_count > 0", title: "{{ shared.t('x_identifications', {count: o.identifications_count}) }}"}
-              %i.icon-identification
               {{ o.identifications_count }}
+              %i.icon-identification
             %span.comments{"ng-show" => "o.comments_count > 0", title: "{{ shared.t('x_comments', {count: o.comments_count}) }}"}
-              %i.icon-chatbubble
               {{ o.comments_count }}
+              %i.icon-chatbubble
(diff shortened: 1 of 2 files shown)


pleary:
Thank you for your work on this. I'm not going to merge this in right now as I believe we need to give this a little more consideration. There are more issues with this section than just long quality grade labels. For example with the numbers of IDs, comments, or faves have additional digits that also causes problem, e.g. https://www.inaturalist.org/observations?order_by=votes&popular. I think simply changing the order of numbers of icons might lead to more confusion in the short term, and won't address the more fundamental styling problems in this section. I suspect we'll need to move components to a new line, and we might benefit for different font colors for the date. So I'll hold off on merging this pull request until we've had a design review and recommendations from our designer.

Thanks again for the time already put into this. I'll update if we make other changes to this section

kueda:
Alrighty, we discussed with our designer and decided we don't want to adopt this approach with the numbers before the icons. We came up with a more explicit spec if you're still interested in working on this, @HannesOberreiter: https://github.com/inaturalist/inaturalist/issues/4165#issuecomment-2619792256. Either way, thanks for the effort here.

HannesOberreiter:
Thank you for sharing your decision with me. I’ve subscribed to the issue, and while my current focus is on paid gigs, I may revisit it if I find some spare time for open source contribution. I appreciate your understanding!

Issue #4108, opened 2024-05-22
Blockee can join blocker's projects


tiwane:
Just tested this out, and as a blocked user I could join projects my blocker had created - both collection and traditional projects. I assume umbrella projects too. They should definiteky not be able to do this. 

I'm wondering if we should remove blocked user from projects the blocker owns?

rainhead:
Can I take this?

pleary:
Yes feel free to work on this. Thanks!

PR #4231, opened 2024-09-20
Blocking removes users from each other's projects

rainhead:
Fixes https://github.com/inaturalist/inaturalist/issues/4108

Regardless of project type,
- When blocking a user:
  - remove blocker and blockee from one another's projects
  - delete any pending project invitations
- Prevent either user from adding the other to their project

This does not affect the requirements of any umbrella project.


pleary:
Looks good. Thanks!

PR #4242, opened 2024-09-29
First observation update

loarie:
Admin only Cohort Lifecycle Needs ID Intervention
see https://docs.google.com/document/d/1F6420hXo5GvUyxyAMreBDkjg3BCqBH9DvFtTR2HVCq4/edit?usp=sharing
new dashboard functionality to allow top IDers (admin only for now) to opt into a pilot to see up to 30 obs to ID a day
these are from the  treatment group of the needs_id cohort established and A/B tested in Cohort Lifecycle 
existing ObservationAccuracyExperiment and CV score image infrastructure used to assign needs_id cohort obs to qualified identifiers


app/models/cohort_lifecycle.rb
```diff
@@ -382,4 +411,178 @@ def self.build_result_hash( obs_data )
...
+    results = response.map do | taxon_bucket |
+      {
+        taxon_id: taxon_bucket["key"],
+        place_ids: taxon_bucket.place_ids.buckets.map {| place_bucket | place_bucket["key"] } & place_ids
+      }
+    end
+    results.map {| a | [a[:taxon_id], a[:place_ids]] }.to_h
+  end
+
+  def self.get_place_obs( oids, place_ids )
```
pleary, line 463: Another minor naming suggestion, I find it much preferable to be explicit with variable names even if they're long. I'd prefer `observation_ids` here over `oids`. Later on you define `observation_ids` in a loop - than could be renamed to `batch_observation_ids` or something like that if `observation_ids` is the method parameter.

pleary (approved):
Minor naming suggestions aside, this looks good to me

PR #4244, opened 2024-10-01
Task Logger

sylvain-morin:
linked to https://github.com/inaturalist/inaturalist-cookbook/pull/60


lib/task_logger.rb
```diff
@@ -0,0 +1,73 @@
...
+
+class TaskLogger
+  # Initialize logger with default parameters
+  def initialize( task, identifier, type, subtype )
+    @task = task
+    @identifier = identifier || default_identifier( task )
+    @type = type
+    @subtype = subtype
+    @start_time = nil
+    @log_file = File.join( CONFIG.task_logger_output, "#{task}.log" )
```
pleary, line 14: We'll probably want to handle the case where `CONFIG.task_logger_output` is undefined. I don't think we'd want it to throw an error - we likely want people to be able to run the code w/o needing to define the config. Alternatively we could do what we're doing with the [application logstash logs](https://github.com/inaturalist/inaturalist/blob/main/config/initializers/logstasher.rb#L25) and put them in the rails log directory which we know will exist. This to me seems preferred that way all the logs are in the same place, we don't need an extra config, there will be different logs for different environments, and we know we'll be able to create a task_logger log file and write to it.

tools/dwca.rb
```diff
@@ -93,11 +93,23 @@
...
+if opts.log_task_name
+  task_logger = TaskLogger.new( opts.log_task_name, nil, "export", "rails" )
+end
+
+begin
+  task_logger&.start
+  DarwinCore::Archive.generate( opts )
+  task_logger&.end
+rescue => e
+  task_logger&.error( e.to_s )
```
pleary, line 114: Might consider catching a few lines of the error `backtrace` as well if we want to use this to pinpoint where errors are occurring. If so, make sure that if you use newlines to separate backtrace lines, those are escaped and don't interfere with keeping each task logger log entry on a single line in the log file

lib/task_logger.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+require "json"
+require "time"
+
+class TaskLogger
+  # Initialize logger with default parameters
+  def initialize( task, identifier, type, subtype )
```
pleary, line 8: It looks like in all cases so far you have the `subtype` set to rails. Would you ever want it to be anything else in this codebase? If not, I'd remove the parameter and hard code it. If so, it might be good to have a comment with some possible suggestions

pleary (changes requested):
Overall looks good. The one thing I think that should be addressed before merging is handling the log file path, currently configured with `CONFIG.task_logger_output`. I think it should gracefully handle the config not being defined, the configured path being incorrect, or not having the correct permissions to write to the path. It would also be nice for it to log to different files for different environments. All that could be handled by using `ActiveSupport::Logger` and not needing an extra config, similar to how we do in Logstasher.

lib/task_logger.rb
sylvain-morin: subtype is "cloudfront" for a few of them (coming from the production_schedule.rb)

lib/task_logger.rb
sylvain-morin: if I'm not using the "default" folder to task logs, I need to update the filebeat config, which is currently only monitoring this default folder (`/var/log/inaturalist-tasks/*.log`) 

I can for sure add a second path like:
```
paths:
    - /var/log/inaturalist-tasks/*.log
    - /home/inaturalist/deployment/production/current/log/task_log_*.logstash.log
```

I suppose it won't impact anything else, since Ruby filebeat config is looking at:
`/home/inaturalist/deployment/production/current/log/production.logstash.log`

is it what you had in mind?

pleary: That is what I had in mind. To me, it would be preferred that the logs generated by this application all go to the same place - `$RAILS_ROOT/logs/*.log`. I understand that other task logs will go elsewhere, but since those will be "generated" externally to this codebase, e.g. in bash or elsewhere, it makes sense they would live in a different location

lib/task_logger.rb
pleary: OK. We could add a default value here (`subtype = "rails"`) and that way in this codebase we wouldn't have to repeat that parameter when creating instances. Not a big deal - just want to have code be as obvious as possible and if other people are working here wonder why we're always sending "rails" and what other values could be sent

lib/tasks/inaturalist.rake
```diff
@@ -47,7 +47,12 @@ namespace :inaturalist do
   end
 
   desc "Delete expired updates"
-  task delete_expired_updates: :environment do
+  task :delete_expired_updates, [:log_task_name] => :environment do | _, args |
+    log_task_name = args[:log_task_name]
+    if log_task_name
+      task_logger = TaskLogger.new( log_task_name, nil, "cleanup", "rails" )
```
pleary, line 53: With `rails` being the default value for subtype it's not necessary to include this last parameter anymore

pleary (approved):
Looks good to me

PR #4247, opened 2024-10-03
fix: Change wiki date format

RichardLitt:
This modifies the template for Wikipedia suggested article content, ensuring that the date field doesn't throw an error on Wikipedia.

Cf. https://en.wikipedia.org/wiki/Help:CS1_errors\#bad_date

### DIFF
--- a/app/views/taxa/_description.html.haml
+++ b/app/views/taxa/_description.html.haml
@@ -56,7 +56,7 @@
         else
           @taxon.name
         end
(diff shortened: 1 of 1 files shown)


RichardLitt:
Let me know if there is something else I should do for this.

RichardLitt:
This continues to be a bug. Let me know what you need.

RichardLitt:
@kueda A PR review for this should take less than one minute.

pleary:
Thanks!

RichardLitt:
Sweet. Thank you.

Issue #4288, opened 2024-11-14
"Follow" button appears on obs detail page if user is logged out


tiwane:
If you're logged-out and on an obs detail page, the "Follow" button appears. No options appear if you click on it, but maybe it should be gone or grayed out if someone isn't logged in? 

![image](https://github.com/user-attachments/assets/316c594d-0bb9-497d-b9ed-424b8ce2f4c5)

PR #4294, opened 2024-11-20
#4288 not showing followButton with not logged users

jotamayo97:
#4288 
I think it’s better to simply not display the button rather than disabling it. Currently, it also shows errors in the console if you click on it.
image

This is my first time contributing to an open source project; any feedback is welcome :)


pleary:
Thanks!

Issue #4289, opened 2024-11-14
On /projects/new change link for "You can learn more about managing projects here"


tiwane:
For "You can learn more about managing projects here" we should update the "here" link to target https://help.inaturalist.org/en/support/solutions/articles/151000176472

kueda:
Fixed in 841b0882685c9e5d4d3a0f45a75ae90b32846407

PR #4297, opened 2024-11-21
#4289 change url text in translations

jotamayo97:
#4289 
I've seen that you use Crowdin for translations but in this particular case with links idk if its ok to do this from here.
Im sorry for the noise in other case.

### DIFF
--- a/config/locales/ar.yml
+++ b/config/locales/ar.yml
@@ -4548,7 +4548,7 @@ ar:
         welcome_to_projects: مرحبا بك في المشاريع!
         you_can_learn_more_html: |
           يمكنك معرفة المزيد حول إدارة المشاريع 
-          <a href="https://www.inaturalist.org/pages/managing-projects">هنا</a>
+          <a href="https://help.inaturalist.org/en/support/solutions/articles/151000176472">هنا</a>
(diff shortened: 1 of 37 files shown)


jotamayo97:
I'm also curious about the deployment process. When do you usually deploy to production?

pleary:
We only edit the en.yml file in this repo and all other changes are pulled from Crowdin. When we make a change to an existing translation, we prefer to generate a new key for it. Crowdin isn't great about highlighting changes made to content of existing keys, and by generating a new key it makes it obvious that work is needed to edit the translation.

Based on those requirements, could you please do a few things for this pull request:
* Revert all changes to files that aren't en.yml
* Change the key in en.yml from `you_can_learn_more_html` to `you_can_learn_more_html2`
* Update all uses of that key to use `you_can_learn_more_html2` instead. It seems like that might only be [here](https://github.com/inaturalist/inaturalist/blob/main/app/webpack/projects/form/components/app.jsx#L151).

If you make those changes we'll be able to merge this in. We deploy most days there are code changes. That is a few times a week at least, some days multiple times a day. Translations are merged in from Crowdin at least once a month

PR #4301, opened 2024-11-29
Admin tool

sylvain-morin:


pleary (changes requested):
I left some comments. It would be good to get the model validations implemented. It would also be great to have some tests for this functionality - model test for validations, and at least some integration tests that IPs get blocked or not for subsequent requests

app/controllers/admin/blocked_ips_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      respond_to do | format |
+        format.html { render layout: "admin" }
+      end
+    end
+
+    def block
+      blocked_ip = BlockedIp.find_or_initialize_by( ip: params[:ip] )
+
+      unless blocked_ip.valid?
+        flash[:notice] = "Failed to block IP `#{ip}`: #{blocked_ip.errors.full_messages.to_sentence}"
```
pleary, line 27: `ip` is not defined here, so when the IP is invalid and error is being thrown. `#{ip}` could be changed to `#{params[:ip]}`

app/controllers/admin/blocked_ips_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      @blocked_ips = BlockedIp.joins( :user ).all
+      @blocked_ips = @blocked_ips.where( "users.login LIKE ?", "%#{@q_blocked_by}%" ) unless @q_blocked_by.blank?
+      @blocked_ips = @blocked_ips.where( "blocked_ips.ip LIKE ?", "%#{@q_ip}%" ) unless @q_ip.blank?
+      @blocked_ips = @blocked_ips.select( "blocked_ips.*, users.login as login" )
+      respond_to do | format |
+        format.html { render layout: "admin" }
+      end
+    end
+
+    def block
```
pleary, line 23: There's a few rails model lifecycle practices that might be good to stick to here. This is a BlockedIps controller, which has a index page, and allows the creation and deletion of BlockedIps. The rails custom is use `create` for the create method (with the `POST` method), and `destroy` for the destroy method (instead of the custom `block` and `unblock` used here). Here are some changes we could make to have it act more consistently with other controllers, e.g.:

```diff
diff --git a/app/controllers/admin/blocked_ips_controller.rb b/app/controllers/admin/blocked_ips_controller.rb
index 070e8fa60..d8b6a4593 100644
--- a/app/controllers/admin/blocked_ips_controller.rb
+++ b/app/controllers/admin/blocked_ips_controller.rb
@@ -4,6 +4,7 @@ class Admin
   class BlockedIpsController < ApplicationController
     before_action :authenticate_user!
     before_action :admin_required
+    before_action :load_record, only: [:destroy]

     layout "application"

@@ -20,11 +21,11 @@ class Admin
       end
     end

-    def block
+    def create
       blocked_ip = BlockedIp.find_or_initialize_by( ip: params[:ip] )

       unless blocked_ip.valid?
-        flash[:notice] = "Failed to block IP `#{ip}`: #{blocked_ip.errors.full_messages.to_sentence}"
+        flash[:notice] = "Failed to block IP `#{params[:ip]}`: #{blocked_ip.errors.full_messages.to_sentence}"
         return redirect_back_or_default( action: "index" )
       end

@@ -34,13 +35,18 @@ class Admin
       redirect_to admin_blocked_ips_path( q_ip: params[:q_ip], q_blocked_by: params[:q_blocked_by] )
     end

-    def unblock
-      blocked_ip = BlockedIp.find_by( ip: params[:ip] )
-      blocked_ip&.destroy
+    def destroy
+      @blocked_ip.destroy

       Rails.cache.delete( "blocked_ips" )

       redirect_to admin_blocked_ips_path( q_ip: params[:q_ip], q_blocked_by: params[:q_blocked_by] )
     end
+
+    protected
+
+    def load_record( options = {} )
+      super( options.merge( klass: BlockedIp ) )
+    end
   end
 end
```

* leverage `load_record` to load instances for instance endpoints, like destroy
* customize `load_record` since it doesn't get the instance class correct by default since this controller is nested under Admin
* Update block -> create, and unblock ->  destroy
* Destroy the instance created with `load_record` (a 404 wold be generated if the instance didn't exist)

If you did want to go with this, then the routes would need to be updated as well, e.g.:

```diff
diff --git a/config/routes.rb b/config/routes.rb
index b0aed26c3..f548bf0a0 100644
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -596,11 +596,9 @@ Rails.application.routes.draw do
         get :pending
       end
     end
-    resources :blocked_ips, only: :index, controller: "admin/blocked_ips" do
+    resources :blocked_ips, only: [:index, :create, :destroy], controller: "admin/blocked_ips" do
       collection do
         get :index
-        put :block
-        delete :unblock
       end
     end
   end
```

* remove the custom methods
* specify that we only have the three defined methods, :index, :create, and :destroy

Lastly, some changes would need to be made to the view for the proper paths and method, e.g.:
```diff
diff --git a/app/views/admin/blocked_ips/index.html.haml b/app/views/admin/blocked_ips/index.html.haml
index 878513b1a..0de5d69a5 100644
--- a/app/views/admin/blocked_ips/index.html.haml
+++ b/app/views/admin/blocked_ips/index.html.haml
@@ -24,7 +24,7 @@
                 =link_to 'Reset', url_for, class: "btn btn-link"
     .col-xs-6
       #controls.well
-        =form_tag( { controller: :blocked_ips, action: :block }, method: :put ) do
+        =form_tag( { controller: :blocked_ips }, method: :post ) do
           .row
             .col-xs-3
               .text-right
@@ -56,5 +56,5 @@
                 %td= blocked_ip.login
                 %td= blocked_ip.created_at
                 %td
-                  = button_to unblock_admin_blocked_ips_path( ip: blocked_ip.ip ), method: :delete, class: "btn btn-warning" do
+                  = button_to admin_blocked_ip_path( blocked_ip ), method: :delete, class: "btn btn-warning" do
                     Unblock
```

app/models/blocked_ip.rb
```diff
@@ -0,0 +1,15 @@
...
+  belongs_to :user
+  validates_uniqueness_of :ip
+  validate :ip_format_valid
+
+  private
+
+  def ip_format_valid
+    return if IPAddr.new( ip ) rescue nil
+
+    errors.add( :ip, "IP is not a valid IP address" )
```
pleary, line 14: Pretty minor, but this error reads a little strangely because it contains the attribute name in it, and the way we are displaying it the attribute will already be displayed (sorry - I know I included this in a previous comment): `Ip IP is not a valid IP address`. It would read better as something like `errors.add( :ip, "must be valid IP address" )`.

Another really small text change would be to tell ActiveRecord that the attribute `ip` should be displayed capitalized. This can be done by adding a translation entry for it to `activerecord.attributes`, e.g.:

```diff
diff --git a/config/locales/en.yml b/config/locales/en.yml
index 25419b5d9..ee1de85c7 100644
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -9623,6 +9623,8 @@ en:
       assessment_section:
         title: Title
         body: Body
+      blocked_ip:
+        ip: IP
       blocked_user:
         # The user who has been blocked
         blocked_user: Blocked user
```

pleary (changes requested):
I added a few more suggestions. The only one that will lead to errors is the ip -> params[:ip] change needed for the error message in the controller. The rest is to make the controller and routes work more in the standard rails way, and more consistent with our existing controllers/routes

spec/requests/blocked_ip_spec.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require "spec_helper"
+
+describe BlockedIp, type: :request do
+  before( :all ) do
```
pleary, line 6: the DB will be cleaned in between each spec, so no need to have this before all block

spec/requests/blocked_ip_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  it "non blocked IP should access page" do
+    get "/ping", headers: { "REMOTE_ADDR" => "192.168.0.1" }
+    expect( response.response_code ).to eq 200
+    expect( response.body ).to include "status"
+  end
+
+  it "blocked IP should not access page" do
+    BlockedIp.create!( ip: "192.168.0.2" )
```
pleary, line 17: We can use [machinist blueprints](https://github.com/inaturalist/inaturalist/blob/main/spec/blueprints.rb) in specs, created with make/make!. If you add a blueprint for BlockedIp, e.g.:
```ruby
BlockedIp.blueprint do
  ip { Faker::Internet.ip_v4_address }
  user { User.make }
end
```
This can just be `blocked_ip = BlockedIp.make!`, and you can use the generated IP value in the header below, `blocked_ip.ip`. You can also send custom values if that's relevant to the spec. As it is now, these specs are failing due to the lack of a specified user, which would come through easily with the blueprint

spec/requests/blocked_ip_spec.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
```
pleary, line 1: This spec is in a new directory `spec/requests` which didn't exist before. We should add that to the [Github actions matrix](https://github.com/inaturalist/inaturalist/blob/main/.github/workflows/CI-build-test.yml) to make sure it gets run during CI - probably in the last entry with helpers, initializers, etc

Issue #4184, opened 2024-08-14
Add 'Last Edited' Section to Observation Fields


tiwane:
From https://forum.inaturalist.org/t/add-last-edited-section-to-observation-fields/50150/

Since anyone can edit an observation field, we should display the username of the account that last edited it in addition to who originally added it when the field is clicked on:

![image](https://github.com/user-attachments/assets/78969f7d-7ee5-4205-bf5a-92866a5b492c)

PR #4302, opened 2024-12-01
feat: add updater field in observation ofvs

jotamayo97:
fix: #4184

API PR: https://github.com/inaturalist/iNaturalistAPI/pull/480
First time with Ruby, Im sorry if I made some mistake.

This approach only works for new obfv updates. in ElasticSearch we are not persisting the updater, but it is where we are taking the rest of the data from. I couldn't find a best way to do this.
I expected to collect this data from postgres but we r using elastic for this. Why?
Thanks for the feedback :)

### DIFF
--- a/app/models/observation_field_value.rb
+++ b/app/models/observation_field_value.rb
@@ -315,6 +315,7 @@ class ObservationFieldValue < ApplicationRecord
       name_ci: observation_field.name,
       value: self.value,
       value_ci: self.value,
+      updater_id: updater_id,
       user_id: user_id
     }
     # make sure taxon_id is an integer
--- a/app/webpack/observations/show/components/observation_field_value.jsx
+++ b/app/webpack/observations/show/components/observation_field_value.jsx
@@ -134,22 +134,22 @@ class ObservationFieldValue extends React.Component {
         <div className="added-by">
           { ofv.user && (
           <div>
-          <div className="view">
-            { I18n.t( "label_colon", { label: I18n.t( "added_by" ) } ) }
-          </div>
-          <a href={`/people/${ofv.user.login}`}>
-            { ofv.user.login }
-          </a>
+            <div className="view">
+              { I18n.t( "label_colon", { label: I18n.t( "added_by" ) } ) }
+            </div>
+            <a href={`/people/${ofv.user.login}`}>
+              { ofv.user.login }
+            </a>
           </div>
           ) }
           { ofv.updater && (
           <div>
-          <div className="view">
-            { I18n.t( "label_colon", { label: I18n.t( "last_edited_by" ) } ) }
-          </div>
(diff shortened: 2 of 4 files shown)


jotamayo97:
I noticed my PRs haven't received any feedback yet, and I see the project is quite active. I just wanted to check if there’s anything I need to adjust or improve :)

Thank you for your time and consideration!

app/webpack/observations/show/components/observation_field_value.jsx
```diff
@@ -132,12 +132,26 @@ class ObservationFieldValue extends React.Component {
...
+          { ofv.updater && (
+          <div>
+          <div className="view">
+            { I18n.t( "label_colon", { label: I18n.t( "last_edited_by" ) } ) }
+          </div>
+          <a href={`/people/${ofv.updater.login}`}>
+            { ofv.updater.login }
+          </a>
+          </div>
+          ) }
```
kueda, line 154: If you've got eslint running, it should be alerting you about indentation rules here. At a minimum, the contents of these outer divs should be indented.

config/locales/en.yml
```diff
@@ -2666,6 +2666,7 @@ en:
   large: large
   last_active_with_date: "Last Active: %{date}"
   last_confirmed_observation: Last confirmed observation
+  last_edited_by: Last edited by
```
kueda, line 2669: Please add a comment explaining to translators what this means and how it's used, e.g. "label for a field listing the name of the user who last edited a record"

app/models/observation_field_value.rb
```diff
@@ -315,6 +315,7 @@ def as_indexed_json
       name_ci: observation_field.name,
       value: self.value,
       value_ci: self.value,
+      updater_id: updater_id,
```
kueda, line 318: This works, but you should also describe this new field and its type around https://github.com/inaturalist/inaturalist/blob/main/app/es_indices/observation_index.rb#L180. This removes some guesswork on the part of elasticsearch and ensures this field is present if/when the index gets rebuilt. It also documents the entire shape of the schema for other developers.

kueda (changes requested):
Thanks, @jotamayo97, this works for me, we just need a few small changes.

kueda:
> I expected to collect this data from postgres but we r using elastic for this. Why?

You could fetch the data from postgres in the api like we're doing in `ObservationPreload.observerProjectMembership()` and similar methods, but your approach here works as well. Round trip to postgres adds some performance overhead, though, so your approach of adding fields to the ES index is probably better from that perspective. It also allows searching on that field if we ever want to do that.

kueda:
@jotamayo97 are you still planning on working on this? If not we can merge and make the changes I requested ourselves.

jotamayo97:
Yeah sorry I will fix this ASAP :)
thanks for waiting for me!

PR #4303, opened 2024-12-01
Update old "Darwin Core Archives" link

frewsxcv:
The old link now redirects to this link

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -7491,7 +7491,7 @@ en:
                 rel="nofollow noreferrer noopener"
                 href="http://www.inaturalist.org/observations/gbif-observations-dwca.zip">DarwinCore
                 Archive (DwC-A) for GBIF</a>: this is the <a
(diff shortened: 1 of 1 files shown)


kueda (changes requested):
Thanks for this, @frewsxcv, but this is going to result in existing translations existing translations getting unapproved without notifying translators. Also, of late, translators have been complaining about minor changes to large chunks of text like this, so if you're up to it, I think the ideal solution would be

- [ ] break this into multiple translatable strings that minimize markup, e.g. put the `p`, `ul`, and `li` tags in the react view and focus the translatable strings on the text
- [ ] take URLs out of the translations and pass them in as variables from the react views since they don't need to be localized

If you're not up for that, the second item and renaming the translation key to `other_options2_html` would be ok.

If you're not up for that either, totally fine, but please let us know so we can close.

frewsxcv:
It may take me a few days to get to it, but I think I have capacity to do all you mentioned, yeah. I'll let you know if not

kueda:
@frewsxcv still planning on working on this?

frewsxcv:
I unfortunately have limited capacity right now, so I won't be able to get to this anytime soon. So I'm going to close this. Thanks for taking a look

PR #4307, opened 2024-12-03
#3818 fix broken links containing at sign

### LINKED ISSUE #3818: Link broken with handle in it
**Describe the bug**
I included a link in the description of an observation that included my handle. The link was broken into 3 parts, the domain, my handle and the rest as text.

**To Reproduce**
1. Post link like https://chaos.social/@davidak/110776355400985286 (Mastodon)

**Expected behavior**
Links should stay intact and should be clickable.

**Screenshots**

Screenshot from 2023-07-25 21-37-56

See https://www.inaturalist.org/observations/174912802

**Context**
 - OS: NixOS 22.11.4588.93fddcf640c
 - Browser: Chrome 114.0.5735.198

**Additional context**
Probably also happens in comments and any other text field where handles are turned into clickable links.

jotamayo97:
fix #3818
It's not the best solution, but I think it's acceptable. I would first check in Elasticsearch how many descriptions contain the characters "/@" in cases where it's not part of a link. I can't think of a reason why someone would do that, but I would check just in case..

I would do it myself, but I don't have access to the database (or I don't know how to access it). I appreciate any tips on how to populate my local environment with real data.

### DIFF
--- a/app/webpack/shared/components/user_text.jsx
+++ b/app/webpack/shared/components/user_text.jsx
@@ -80,7 +80,7 @@ const CONFIG = {
 class UserText extends React.Component {
   // Imperfect solution until we can get an API endpoint to check for the existence of these users
   static hyperlinkMentions( text ) {
-    return text.replace( /(\B)@([A-z][\\\w\\\-_]*)/g, "$1<a href=\"/people/$2\">@$2</a>" );
+    return text.replace( /(\B)(?<!\/)@([A-z][\\\w\\\-_]*)/g, "$1<a href=\"/people/$2\">@$2</a>" );
   }
 
   constructor( ) {


pleary:
Thanks! As for populating an environment with "real" data, we don't have good methods for doing that. The [Development Setup Guide](https://github.com/inaturalist/inaturalist/wiki/Development-Setup-Guide) instructs to run `ruby bin/setup` which will load some [generic data](https://github.com/inaturalist/inaturalist/blob/main/bin/setup#L61). There is also a tool designed to [load taxon information](https://github.com/inaturalist/inaturalist/blob/main/tools/development_tools/import_taxon_from_production.rb) form production using the API. Aside from those, we don't have other tooling to help developers load "real" data. This is an area we could definitely spend time on to improve the developer experience

Issue #3620, opened 2023-01-12
Pressing cancel when editing comment on journal post generates whale shark error


tiwane:
See https://forum.inaturalist.org/t/pressing-cancel-when-editing-comment-on-journal-post-generates-whale-shark-error/38357 I can replicate in Firefox. 

1. Post comment on project journal post
2. Click ‘edit’ for comment.
3. Either make some changes, or make none (error occurs regardless)
4. Click cancel
5. Whale shark error appears

PR #4308, opened 2024-12-03
#3620 fix error when cancel edition of comment in projects

jotamayo97:
It seems like a very obvious mistake, but it's my first time working with .haml. I'm concerned it might have implications I’m not aware of, as I don’t fully understand this code tbh.
I've tested it in many scenarios, and it seems to work.


pleary:
Looks good - thanks!

Issue #3755, opened 2023-05-18
Old Life List name doesn't change if username changes


tiwane:
If a user had an account before we introduced the Dynamic Life List and has since changed their username, their old life list name doesn't update to display the account's current name - it still shows the old name.

cfgrok:
I would like to begin contributing to the iNaturalist project, and this looks to be a simple place to start getting my feet wet, so please assign me to this issue -- PR to follow shortly.

PR #4317, opened 2024-12-11
fix: replace legacy life list titles (#3755)

cfgrok:
It looks to me like the most straightforward way to address this issue is to override the title attribute for lists, and substitute the :life_list translation with the current list owner for the legacy title.

Note that this will mean that anyone in a non-English locale will now see a localized version of the "xxx's Life List" text instead of the hard-coded English phrase. From looking at the [old life list code](https://github.com/inaturalist/inaturalist/blob/a77a4153b002d3d12d955299cd08597819fe9817/app/models/life_list.rb#L239-L243), I see that there was also a taxon-specific version of a life list, but I did not see an appropriate translation for this, so the substitution only replaces the "xxx's Life List" portion of the title.

Please let me know if a different approach is preferred here.

### DIFF
--- a/app/models/check_list.rb
+++ b/app/models/check_list.rb
@@ -36,7 +36,7 @@ class CheckList < List
   end
   
   def owner_name
-    self.place.name
+    place&.name
   end
   
   # Is this the default check list of its place?
--- a/app/models/list.rb
+++ b/app/models/list.rb
@@ -49,7 +49,11 @@ class List < ApplicationRecord
     return nil if new_record?
     CGI.escape("#{id}-#{title.gsub(/[\'\"]/, '').gsub(/\W/, '-')}")
   end
-  
+
+  def title
+    self[:title]&.sub(/^.+'s Life List/, I18n.t(:life_list, user: owner_name))
+  end
+
   #
   # Adds a taxon to this list and returns the listed_taxon (valid or not). 
   # Note that subclasses like CheckList may override this.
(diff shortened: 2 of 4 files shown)


cfgrok:
My apologies for missing the failing specs. At present I do not have an entirely green test suite, and evidently some of the failures are intermittent, as I actually had fewer failures **after** applying my changes, but I should have taken a closer look at which specs failed, as had I done so I would have seen the failures that resulted from this change.

At any rate, I have resolved both failing specs by using the safe navigation operator in the two models in question, which prevents the NoMethodError exception from being raised by the owner_name method call when the relevant association has not been set, as is the case with the failing specs. I have squashed this change into the existing commit, and rebased onto the current main branch, so this PR should now be ready to go.

kueda (approved):
This looks good to me. I think @pleary may have some input too.

pleary:
I had some somewhat pedantic suggestions, but I was probably over thinking it. This looks OK to me. Apologies for the delay in responding to this pull-request. Thanks @cfgrok for working on and submitting this!

kueda:
@cfgrok, you're right that this doesn't handle taxon specific life lists, e.g. [rohitmg's Life List of Butterflies and Moths](https://www.inaturalist.org/lists/81023), but your solution improves that vast majority of such lists so that seems like a step in the right direction. We can deal with the taxon-specific lists in a separate issue if that comes up. I suspect we could handle it with a new translation, e.g. `%{user}'s Life List of %{taxon}`, but that will be imperfect too due to the need for inflections on two variables.

PR #4327, opened 2024-12-18
Potential improvement for explore-taxon-addition-subtraction

pleary:
- add to the species_counts API call the [new parameter](https://github.com/inaturalist/iNaturalistAPI/commit/96126150e08c146d4b9f55a1bad0ecefe4ce119c) `include_ancestors` to prevent the need for subsequent taxon details API calls to fetch taxon ancestors
- ensure that if the user is testing API v2, taxon ancestors include the attribute `preferred_common_names` in order to display multiple common names


kueda (approved):
Thanks, Patrick! This is much simpler.

PR #4339, opened 2025-01-06
Remove 0 padding from day as this causes a CS1 error on en wiki.

mvolz:
References currently created this way causes errors in citation templates on en wiki because the template does not accept days with leading zeros (see zero-padding: https://en.wikipedia.org/wiki/Help:CS1_errors#bad_date )

%-d inserts the day with no leading 0 instead.


pleary:
Thanks!

PR #4348, opened 2025-01-13
Display higher resolution user images

kueda:
This started as an effort to make Alex look nicer, but since Abhas also wants to look nicer, why don't we just make everyone look nicer. I wasn't really sure if anyone else cared.


abhasinat:
image

PR #4366, opened 2025-01-30
fix: resend messages by unsuspended user

kueda:
This only goes part way toward WEB-367 by handling the case where the sender is suspended but their messages are not spam. I'm mostly interested in critiques of the approach using the `sent_at` column in the `messages` table. Overly complicated? Edge cases I'm missing?

Uh, there are a lot of Rubocop changes too, so it might help to look at things commit by commit.

Update: this PR now closes WEB-367.

### DIFF
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -135,7 +135,12 @@ class Message < ApplicationRecord
     true
   end
 
-  def flagged_with( flag, _options = {} )
+  def flagged_with( flag, options = {} )
     evaluate_new_flag_for_spam( flag )
+    return unless options[:action] == "resolved" && flag.flag == Flag::SPAM
+
+    Message.
+      delay( priority: USER_INTEGRITY_PRIORITY ).
+      resend_unsent_for_user( user_id )
   end
 end
--- a/app/models/message.rb
+++ b/app/models/message.rb
@@ -1,37 +1,38 @@
+# frozen_string_literal: true
+
 class Message < ApplicationRecord
-  acts_as_spammable fields: [ :subject, :body ],
+  acts_as_spammable fields: [:subject, :body],
     user: :from_user
 
-  blockable_by lambda { |message| message.to_user_id },
-    blockable_user_id: lambda { |message| message.from_user_id }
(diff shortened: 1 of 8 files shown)


app/models/message.rb
```diff
@@ -1,59 +1,89 @@
...
+  # messages that the recipient deleted.
+  RESEND_UNSENT_RELEASE_DATE = Date.parse( "2025-01-28" )
 
-  attr_accessor :html
-  attr_accessor :skip_email
+  def self.resend_unsent_for_user( user )
+    user = User.find_by_id( user ) unless user.is_a?( User )
+    return unless user
+
+    scope = user.messages.sent.
```
pleary, line 48: The fact there is a scope named `sent`, and a new method named `sent?` with different meanings is a little confusing for readability. I was about to comment - should this say `.unsent.` - but this code seems right since its using the sent scope. Just a little strange that we're using the `sent` scope, and in the block below we skip the message if its `sent?`. Maybe worth a comment in cases where we use the two in close proximity?

pleary (approved):
This all seems good to me. I think the addition of the `sent_at`, and having to be nullified on suspension, does a good job at helping resolve this issue. I can't think of any edge cases not covered. I also think its OK that this cannot be applied retroactively

app/models/message.rb
kueda: Yeah, not ideal. I started with "delivered" and then realized we were using that word to mean an email was delivered, so I went with "sent" to avoid confusion there, then realized we had that scope. How would you feel about renaming the scope to "outbox" instead? It's a scope that retrieves the user's copy of the messages they *tried* to send, not necessarily messages that were actually sent, so "outbox" might capture the meaning a little better, even if it breaks the email analogy a bit.

pleary: I'm OK with renaming the scope to `outbox`. With any scope I usually need to go back to the definition to see what it _really_ means. I think `outbox` would be much better for at a glance readability, especially now that messages may not have been `sent` even though there is a copy from the sender.

PR #4382, opened 2025-02-04
fix: allow Turkic tag reuse

kueda:
Took a little digging, but the underlying problem is that acts_as_taggable_on assumes tags are case-insensitive, and when it tries to find an existing tag, it does something like `where( "LOWER(name) = ?", new_tag_name.downcase )`. However, with the default Postgres collation for a VARCHAR column, `LOWER()` behaves differently for some Turkic characters than Ruby's `downcase` method, so in the case of the photo in the issue, the keyword "İç Anadolu Bölgesi" gets lowercased in different ways:

Postgres `LOWER()`: iç anadolu bölgesi
Ruby `downcase`: i̇ç anadolu bölgesi

(the first letters are slightly different)

When acts_as_taggable_on received that as a potential tag, it tried to find an existing tag, failed b/c of the different lowercasing behavior, tried to create a *new* tag, but that failed to pass the uniqueness constraint on `name`. Curiously, this error on the tag did not apply to the observation, so while the observation did not save, it also did not appear invalid and POST /observations suceeded without creating the observation.

We could probably solve this by patching acts_as_taggable_on, but this seemed like a simpler and easier-to-maintain approach, albeit with some performance impact during the migration.

Definitely curious about an overall critique, but also curious about performance impact and whether this requires rebuilding postgres indices or if that happens automatically (I think it did for me locally).

Closes WEB-576

### DIFF
--- a/app/controllers/observations_controller.rb
+++ b/app/controllers/observations_controller.rb
@@ -658,13 +658,19 @@ class ObservationsController < ApplicationController
 
     # check for errors
     errors = false
-    if params[:uploader]
-      @observations.compact.each { |obs|
-        obs.errors.delete(:project_observations)
-        errors = true if obs.errors.any?
-      }
-    else
-      @observations.compact.each { |obs| errors = true unless obs.valid? }
+    @observations.compact.each do | obs |
+      if params[:uploader]
+        obs.errors.delete( :project_observations )
+      end
+      if obs.errors.any?
+        errors = true
+      elsif !obs.persisted?
+        # If for some reason an observation has no errors but still isn't
+        # persisted, something is still wrong and we should not fail
+        # silently
+        errors = true
+        obs.errors.add( :base, :cannot_be_saved_for_unknown_reason )
+      end
     end
     Observation.elastic_index!(
       ids: @observations.compact.map( &:id ),
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -9969,6 +9969,10 @@ en:
           attributes:
             base:
               coordinates_cannot_be_zero_zero: "coordinates cannot be 0,0"
+              # Message that might be displayed if an observation fails to
+              # save for an unknown reason, generally indicates a bug
+              cannot_be_saved_for_unknown_reason: |
+                cannot be saved for unknown reason, please take a screenshot and report to staff
             observed_on:
               cannot_be_greater_than_date_created: |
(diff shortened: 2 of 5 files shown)


pleary (approved):
This seems OK to me. I agree it might be better if ActsAsTaggableOn used a better validation that didn't use two different approaches to lowercase a name so the comparison was using a common understanding and consistent.

This alternate collation seems to work OK. I did see these warnings when running on my local setup:
```
WARNING:  collation "und-x-icu" has version mismatch
DETAIL:  The collation in the database was created using version 153.112, but the operating system provides version 153.120.
HINT:  Rebuild all objects affected by this collation and run ALTER COLLATION pg_catalog."und-x-icu" REFRESH VERSION, or build PostgreSQL with the right library version.
```
Things seemed to work OK without running those commands. But I also did run the commands and things seemed to continue to work OK just without the warnings. I did not see these warnings on staging.

It seems like PostgreSQL [recommends reindexing](https://www.postgresql.org/docs/current/sql-altercollation.html#SQL-ALTERCOLLATION-NOTES) when changing collation. Things seemed to work without it, but reindexing the affected indexes does seem pretty fast on staging. The REINDEX command will lock the table to writes, but given it only takes a few seconds this shouldn't have much of a negative impact, especially if run during lower usage hours. e.g. 
```
REINDEX INDEX index_tags_on_name;
REINDEX INDEX index_tags_on_lower_name;
```

I'd say its probably best to add these to run when this is deployed, but I also can't right now replicate a problem we'd hit if we didn't reindex.

Overall looks OK to me to merge and deploy

kueda:
Thanks for the review, @pleary.

> I'd say its probably best to add these to run when this is deployed

Do you think it's worth putting them in the migration? Or just run them manually afterward?

pleary:
I looked at that PostgreSQL link I included, and realized that's referring to when the collation _definitions_ change, e.g. running the `ALTER COLLATION` command it said I should run locally, not about changing column collation. I looked again and saw on the ALTER TABLE [docs page](https://www.postgresql.org/docs/current/sql-altertable.html#SQL-ALTERTABLE-NOTES) a specific mention of this case:
```
...if the collation for a column has been changed, an index rebuild is always required because the new sort order might be different.
```
While I don't think we care too much about tag name sort order, and that is the only issue they explicitly mention, I say we should put the `REINDEX INDEX` commands in a migration and get it all done at the same time

PR #4389, opened 2025-02-11
fix: don't ask consent for every google oauth2 signins (only at signup)

smaftoul:
It seems to me I often need to signin again, and each time I signin again with google, the site asks for consent to access services I already consented. It seems to me, most sites on internet doesn't behave like this and I find this signin flow annoying.

I tested with a local sinatra app with omniauth and a omniauth-google-oauth2 strategy.
I first used the same settings (`:prompt => "consent"`) , authenticated with my 1st user.
Then I re-opened, re logged in , and had the consent again.
Then, I switched to `:prompt => "select_account"`, re-opened the app, and logged in, with selecting my account like previsously, but without seeing the consent again.
I then tested with another google account, with `select_account`, the 1st login prompt for selecting account and consenting to used service, and later on, no consent again.


kueda (approved):
Works for me locally.

kueda:
Just deployed this, @smaftoul, please check it out. IMO, definitely an improvement, so thank you.

smaftoul:
Agreed, just checked, and seems better ;)

PR #4392, opened 2025-02-12
fix: localize GMaps using browser locale if no region in user locale

HannesOberreiter:
Most language codes in `I18n.locale` do not include a region, therefore returning an empty string for `cctld_from_locale`.

This fix uses the request header `HTTP_ACCEPT_LANGUAGE` as a sane fallback for this cases.

Cheers
Hannes

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1504,10 +1504,14 @@ module ApplicationHelper
     request.env["HTTP_ACCEPT_LANGUAGE"]&.scan( /[a-z]{2}-[A-Z]{2}/ )&.first&.downcase
   end
 
-  # Mostly for Google API regions
+  # Specifies the region parameter for the Google API
+  # Defaults to US if not set
+  # Available regions: https://developers.google.com/maps/coverage
   def cctld_from_locale( locale )
     # return "il" if locale.to_s == "he"
     if locale.nil? || locale.to_s.split( "-" ).size < 2
+      # Most user language preferences do not include a region
+      # Therefore, we attempt to infer the region from the request
       locale = locale_from_request || ""
     end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1500,7 +1500,7 @@ module ApplicationHelper
     }.to_query
   end
 
-  def region_from_browser
+  def locale_from_request
     request.env["HTTP_ACCEPT_LANGUAGE"]&.scan( /[a-z]{2}-[A-Z]{2}/ )&.first&.downcase
   end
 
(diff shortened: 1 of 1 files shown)


kueda:
@HannesOberreiter is there a specific problem that this change solves? Loading Google Maps JS with a blank `region` doesn't prevent maps from loading. I'm seeing that some unlikely combinations like `en-DE` do alter the labeling of the Gulf of Mexico, which I'm guessing is why you're doing this, but for me it only changes it to "Gulf of Mexico (Gulf of America)" which is not much of an improvement.

HannesOberreiter:
More or less this is the reason, as example I use `de` as language in my iNaturalist settings. With this settings I get an empty region, as the language code is simply `de` without any region. Therefore the resulting map only shows "Golf von Amerika".

With my "workaround" I would get your mentioned `Golf von Mexiko (Golf von Amerika)` which I would personally absolutely prefer. As it seems that google maps will use region "US" as default if there is no region given.

image

app/helpers/application_helper.rb
```diff
@@ -1500,10 +1500,16 @@ def google_maps_loader_uri( libraries: [] )
...
+  def region_from_browser
+    request.env["HTTP_ACCEPT_LANGUAGE"]&.scan( /[a-z]{2}-[A-Z]{2}/ )&.first&.downcase
+  end
+
   # Mostly for Google API regions
   def cctld_from_locale( locale )
     # return "il" if locale.to_s == "he"
-    return if locale.to_s.split( "-" ).size < 2
+    if locale.nil? || locale.to_s.split( "-" ).size < 2
+      locale = region_from_browser || ""
```
kueda, line 1511: IMO this is actually an improvement over what's here, but the naming is a big confusing: `region_from_browser` doesn't return a locale, it returns a region. Maybe instead of this and the next line you could do something like

```ruby
region = if locale.to_s.split( "-" ).size > 1
  locale.to_s.split( "-" ).last
else
  region_from_browser || ""
end
```

kueda (changes requested):
Now that I know what you're going for here, this seems harmless and somewhat beneficial. However, since it's not obvious what effect this has, could you please add a comment explaining that this makes it less likely that Google defaults to a `US` region for users who use iNat locales? This is in addition to the change I requested inline.

app/helpers/application_helper.rb
HannesOberreiter: Below we have a mapping going on (at least the start of it and would need some more attention). I don't want to skip that part when using `region_from_browser`. I will rename the definition to `locale_from_request` which is anyway more correct and the regex returns actually `de-AT` therefore we need also the split afterwards.

kueda (approved):

kueda:
@HannesOberreiter we deployed this today. Are the maps looking any better?

HannesOberreiter:
The obvious one from our discussion looks good on my side. As visitor and user with lang preference setup. If I change the browser language settings it changes accordingly. 

So yeah at least for me it seems to work. Thanks for the code review!

PR #4394, opened 2025-02-13
docs: document need to start docker before making services

elharo:
Following the existing instructions without this added step leads to:

```
~/inaturalist$ make services
docker compose build --parallel es memcached redis pg
docker compose up -d es memcached redis pg
unable to get image 'redis:6.0.3': Cannot connect to the Docker daemon at unix:///Users/elharo/.docker/run/docker.sock. Is the docker daemon running?
make: *** [services] Error 1
```

@kueda

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,9 +12,8 @@ We will probably close pull requests that don't address open issues. Again, if y
 
 The development environment for iNaturalist is set up to run containerized services in Docker. Furthermore this configuration allows running the [API](https://github.com/inaturalist/iNaturalistAPI) alongside the services. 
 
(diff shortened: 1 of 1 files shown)


elharo:
I suppose it depends on how one works, and perhaps platform. I'm usually not running Docker unless I need it, and have to launch it.

kueda (changes requested):
How about you go with a combo approach, e.g.

> 1. Install and launch [Docker](https://www.docker.com/). Make sure your current user has access to the [docker daemon](https://docs.docker.com/engine/install/linux-postinstall/) with docker run hello-world.

elharo (commented):
PTAL

kueda (approved):

PR #4399, opened 2025-02-13
Update to Ruby 3.4.1

### LINKED ISSUE #4397: Update Ruby Version to 3.1 or later
Right now the gemfile requires Ruby 3.0.4 specifically. 3.0.x is EOL and should probably be migrated to a later supported version.

This is the message I get when installing 3.0.4:

WARNING: ruby-3.0.4 is past its end of life and is now unsupported.
It no longer receives bug fixes or critical security updates.

I'm not sure which version is best to use. Probably at least 3.2, though we might want to jump all the way to 3.4.1 if there aren't compatibility problems.

elharo:
Let's see if this works...

fixes #4397


elharo:
```
An error occurred while installing sassc (2.4.0), and Bundler cannot
continue.

In Gemfile:
  sass-rails was resolved to 6.0.0, which depends on
    sassc-rails was resolved to 2.1.2, which depends on
      sassc
```

elharo:
pg_result.c:1590:58: error: ‘rb_cData’ undeclared (first use in this function)

This seems to be a known problem with the pg gem. Looking for a solution...

kueda:
@elharo as of https://github.com/inaturalist/inaturalist/pull/4450 we have migrated to Ruby 3.3. Does that meet your needs,a and if so can we close this?

elharo:
For now that should be good. It's a supported version. Eventually it will probably need to go higher.

PR #4404, opened 2025-02-15
fix: separate & wrap metadata items better in Explore

kueda:
* Add an icon before the time to avoid confusion with other numbers like ID count
* Wrap metadata when quality grade label is too wide
* Avoid layout overlaps in list view

Closes WEB-506

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -12,11 +12,11 @@ We will probably close pull requests that don't address open issues. Again, if y
 
 The development environment for iNaturalist is set up to run containerized services in Docker. Furthermore this configuration allows running the [API](https://github.com/inaturalist/iNaturalistAPI) alongside the services. 
 
-1. Install [Docker](https://www.docker.com/)
+1. Install and launch [Docker](https://www.docker.com/). Make sure your current user has access to the docker daemon with `docker run hello-world`.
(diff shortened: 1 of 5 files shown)


HannesOberreiter:
Thank you for addressing the text direction! I completely overlooked it as I am not accustomed to programming for different text directions. There is always more to learn. :)

---

This might be more of a design question and not directly related to this pull request, but I noticed a potential improvement while working on this issue. I am not very fond of the current grid view's shortened text style for dates longer than a month or a year for non-obscured data (eg. 2 mo, 1 y ...). I prefer the style used for obscured dates, where both the month and year are displayed. I suggest changing the default shortened date text to use the month/year format for dates longer ago than a month.

kueda:
> I suggest changing the default shortened date text to use the month/year format for dates longer ago than a month

I agree "Dec '24" requires less mental math than "2mo". @abhasinat any opinion? If Abhas doesn't chime in within a few days, go ahead and make an issue or PR for that, Hannes.

HannesOberreiter:
Sure can do, will probably wait for this one to be merged to prevent any merge conflicts which may arise. 

As one should only do one feature per pull/feature branch and I believe you will do a squash merge here for a clean history.

Cheers
Hannes

PR #4409, opened 2025-02-21
feat: ✨ add showTimeAgo param to inat-calendar-date directive

HannesOberreiter:
Follow-up on the discussion in: https://github.com/inaturalist/inaturalist/pull/4404

As a result of the changes I made, there are some global updates affecting all instances where the directive is used.

## Affected Components

- `/app/assets/javascripts/ang/templates/observation_search/results_grid.html.haml`
- `/app/assets/javascripts/ang/templates/observation_search/results_table.html.haml`
- `/app/assets/javascripts/ang/templates/shared/observation.html.haml`

## Global Changes

- The title is now wrapped around both `span` elements, resulting in a hover effect over the entire title.
- Previously, the `short` flag only had an effect when used with `obscured`. With my changes, it now also affects output without the `obscured` flag. In instances where this change should not occur, I applied the same logic to `short` as for `obscured` (`short: "o.obscured && !o.private_geojson"`).

## Specific Changes

- The directive now has an optional `title` parameter, which will override the title in all cases if provided. This is used for the `created_at` instances in `shared/observation.html.haml`.
- The directive now has an optional `showTimeAgo` parameter. It will display the same string as the `am-time-ago` Angular directive for up to 30 days ago. After that, it will use the default settings.

## Discussion

Do we actually need `created_at` in the `shared/observation.html.haml` template? When looking at the map grid, it seems more logical to style it the same way as the grid view. At first glance, I did not realize that the bottom value is `created_at` and not `observed_at`. In my opinion, we should remove `observed_at` after the location string and only show `observed_at` after the icons, as in the grid view, while removing the `created_at` information.

### Map Grid

image

## View Grid

image

## Hover effect now over the whole div

image

### DIFF
--- a/app/assets/javascripts/ang/inaturalist_angular.js
+++ b/app/assets/javascripts/ang/inaturalist_angular.js
@@ -218,7 +218,9 @@ iNatAPI.directive( "inatCalendarDate", ["shared", function ( shared ) {
       timezone: "=",
       obscured: "=",
       short: "=",
-      viewersTimezone: "="
+      viewersTimezone: "=",
+      showTimeAgo: "=",
+      title: "="
     },
     // eslint-disable-next-line no-unused-vars
     link: function ( scope, elt, attr ) {
@@ -238,12 +240,14 @@ iNatAPI.directive( "inatCalendarDate", ["shared", function ( shared ) {
         var date = moment.tz( scope.time || scope.date, timezone );
         var now = moment( new Date( ) );
         var dateString;
-        if ( date.isSame( now, "day" ) ) {
+        if ( scope.showTimeAgo && date.isAfter( moment( ).subtract( 1, "month" ) ) ) {
+          dateString = date.fromNow();
+        } else if ( date.isSame( now, "day" ) ) {
           dateString = I18n.t( "today" );
         } else if ( date.isSame( now.subtract( 1, "day" ), "day" ) ) {
           dateString = I18n.t( "yesterday" );
         } else {
-          dateString = date.format( "ll" );
+          dateString = date.format( scope.short ? I18n.t( "momentjs.month_year_short" ) : "ll" );
         }
         return dateString;
       };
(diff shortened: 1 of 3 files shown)


kueda (changes requested):
Mostly looking good, but we need a few changes:

- [x] when an observation is obscured, replace the clock icon with an `icon-icn-location-obscured` icon (use `icon-icn-location-private` when private); otherwise users lose any indication that the date is being obscured vs merely shortened
- [x] when geoprivacy is obscured or private, remote the `title` attribute from the date if the viewer cannot view the true coordinates or date

kueda:
> Do we actually need created_at in the shared/observation.html.haml template?

If we start removing things *someone* will be unhappy. Explore is largely in maintenance mode so IMO removing stuff like this is probably a step too far for the time being.

HannesOberreiter:
- [x] when an observation is obscured, replace the clock icon with an icon-icn-location-obscured icon (use icon-icn-location-private when private); otherwise users lose any indication that the date is being obscured vs merely shortened --> **reused the `Observation.placeIcon()` method**
- [x] when geoprivacy is obscured or private, remote the title attribute from the date if the viewer cannot view the true coordinates or date --> **title is now removed, except if user has access to `private_geojson`**


image


> If we start removing things someone will be unhappy. Explore is largely in maintenance mode so IMO removing stuff like this is probably a step too far for the time being.

Sure no problem. Was more like an "aha" moment for myself when I realized that the exact same text row in Map and Grid view have two completely different meaning. Wonder how many people realise that one is the created time and the other one the observation one.

kueda (approved):

Kildor:
No, please! Give us ability to see full date without hovering.
This is one of the most annoying thing in modern web, all these "* hours ago", and so on and so for.

HannesOberreiter:
@Kildor I believe there may be a misunderstanding about this pull request. The change implemented was an improvement from the previous "time ago" format to at least show month/year information. This was a relatively minor update, which is why it involved fewer decision making and could be implemented rather quickly.

If you'd like to see full dates displayed (which I would also support), I'd recommend starting a new discussion thread with some mockups or examples of your proposed format. This would allow the iNaturalist team to properly evaluate and discuss the suggestion with all relevant team members, as I'm not related to iNaturalist.

Cheers
Hannes

PR #4420, opened 2025-02-26
Update to Ruby 3.0.7

elharo:
Just trying something. Let's see if 3.0.7 works.

3.0.5, 3.0.6, and 3.0.7 all included security fixes, so it seems worthwhile to update at least this far until someone manages to update to 3.1. or later.

https://www.ruby-lang.org/en/news/2024/04/23/ruby-3-0-7-released/


elharo:
@kueda It looks like we can update from Ruby 3.0.4 to 3.0.7 without anything obvious breaking. This still doesn't get iNat onto a supported Ruby version but it might be a helpful intermediate step on the way to 3.4.1 if you feel comfortable merging this.

kueda:
My personal preference would be to avoid infrastructural changes like this until we've had multiple sets of eyes on it, so I defer to @pleary and @sylvain-morin as they are more responsible for the ops implications of these kinds of changes. If they want to merge this, great.

FWIW, I think infrastructural changes are better suited to staff, not volunteers. This is why we [recommend](https://github.com/inaturalist/inaturalist/blob/main/CONTRIBUTING.md#finding-issues-to-work-on) working on issues in this repo staff have written. I realize this is just a patch upgrade and the risk is low, and for you this might be a pre-req for working on existing issues, but please to restrict future PRs to those issues.

pleary:
@elharo thanks for looking into this. I agree that a change of language version is best left for staff to manage. There may be reasons to pick one version over another related to other iNaturalist production infrastructure impacts outside of just this codebase. I agree an update would be useful for several reasons, and we are currently working on it. It is looking like Ruby 3.3.7 will be the next target, and some changes have been [made in a branch](https://github.com/inaturalist/inaturalist/compare/main...ruby-33) towards that goal.

We'll keep you posted when the main branch has been updated to reflect a ruby upgrade. We'll also do our best to keep development guides updated for various ways the codebase could be installed, calling out non-obvious steps users may need to take, for example compiling ruby with OpenSSL support which can be a common issue when not using containers

PR #4421, opened 2025-02-26
feat: remove Twitter sign in

kueda:
Twitter sign in has not worked for years, so this removes it while leaving the ProviderAuthorization records for bookkeeping (without tokens or secrets).


db/migrate/20250226225252_remove_site_twitter_sign_in.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class RemoveSiteTwitterSignIn < ActiveRecord::Migration[6.1]
+  def up
+    # Remove site preference for twitter sign in
+    execute "DELETE FROM preferences WHERE owner_type = 'Site' AND name = 'twitter_sign_in'"
+
+    # Remove potentially private data from twitter
+    execute <<-SQL
```
pleary, line 9: Any reason not to delete the authorizations entirely? Locally I had a Twitter authorization, and with the code in this branch I still see a line for it on my `Connected Accounts` user edit screen, but there is no label for it. That's because Twitter was removed from `ProviderAuthorization::PROVIDERS` and we use `PROVIDER_NAMES` to label the authorizations on that settings page. So it appears I've given authorization in the past to some unknown application

db/migrate/20250226225252_remove_site_twitter_sign_in.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class RemoveSiteTwitterSignIn < ActiveRecord::Migration[6.1]
+  def up
+    # Remove site preference for twitter sign in
+    execute "DELETE FROM preferences WHERE owner_type = 'Site' AND name = 'twitter_sign_in'"
```
pleary, line 6: This query will be quite slow due to minimal indices on the preferences table. I don't expect that to be a problem necessarily - I believe only affected rows would get locked after postgres figures out what those are. There will also only be 1 matching row. Not necessary, but a possible change would be to loop through each site's preferences separately as that would leverage the table index, e.g. `Site.all.each{ |site| site.stored_preferences.where( name: "twitter_sign_in" ).destroy_all }`

db/migrate/20250226225252_remove_site_twitter_sign_in.rb
kueda: Only reason not to delete them entirely would be for recordkeeping and/or dealing with support issues. Leaving them in place at least allows us to know that an account *was* associated with a Twitter account. @tiwane and @kvangork what do you think? It would definitely simplify things if we could simply erase any trace of connections to twitter authorizations.

db/migrate/20250226225252_remove_site_twitter_sign_in.rb
kueda: Thanks, I updated it to use the approach you recommended.

db/migrate/20250226225252_remove_site_twitter_sign_in.rb
tiwane: I honestly can't remember the last time someone has contacted support about Twitter log in not working, I suspect it's been quite a few years. I don't think there's much benefit to keeping the record for these.

kueda: Thanks for the input all. I will remove the records entirely instead of simply removing the tokens.

PR #4422, opened 2025-02-27
feat: document domain-specific terms from obs exports and DwC-A

kueda:
@pleary this is pretty rough (page text needs to be globalized, etc) but I'd love your input on a few things:

1. Does a single namespace make sense or should we be namespacing terms that get used in different places, e.g. `/schema/observations/export` vs `/schema/dwca` or something?
2. How do you feel about changing the URIs of our custom fields in the DwC-As, e.g. changing `https://www.inaturalist.org/terms/numIdentificationAgreements` to `https://www.inaturalist.org/terms/num_identification_agreements`? My thinking was that a DwC-A is somewhat self-documenting and DwC-A consumers are probably sophisticated enough to deal with these changes, while obs export users might be more sensitive to changes, but that could be way off
3. How are these definitions looking to you?


lib/darwin_core/occurrence.rb
```diff
@@ -38,7 +38,7 @@ class Occurrence
       %w(recordedByID http://rs.tdwg.org/dwc/terms/recordedByID),
       %w(identifiedBy http://rs.tdwg.org/dwc/terms/identifiedBy),
       %w(identifiedByID http://rs.tdwg.org/dwc/terms/identifiedByID),
-      %w(captive https://www.inaturalist.org/schema/terms/captive),
+      %w(captive https://www.inaturalist.org/schema/terms/captive_cultivated),
```
pleary, line 41: pretty minor - if we're already changing the URI, would it make sense to put it at `https://www.inaturalist.org/terms/captive_cultivated` (remove schema). I think this is the only place we reference `schema/terms` when using our own custom URIs. Both would resolve, but removing schema seems like it would be most consistent

config/locales/en.yml
```diff
@@ -7536,46 +7546,126 @@ en:
...
+          coordinates, if known
+        private_place_guess: |
+          Observation location as written by the observer if location obscured
+        public_positional_accuracy:
+          Maximum horizontal positional uncertainty in meters; includes
+          uncertainty added by coordinate obscuration
+        taxon_class_name: |
+          Name of the taxonomic class containing the observed taxon
+        taxon_family_name: |
+          Name of the taxonomic family class containing the observed taxon
```
pleary, line 7625: family ~~class~~

pleary (approved):

pleary:
> 1. Does a single namespace make sense or should we be namespacing terms that get used in different places, e.g. /schema/observations/export vs /schema/dwca or something?

I think a single namespace is preferred to keep it simple

> 2. How do you feel about changing the URIs of our custom fields in the DwC-As, e.g. changing https://www.inaturalist.org/terms/numIdentificationAgreements to https://www.inaturalist.org/terms/num_identification_agreements? My thinking was that a DwC-A is somewhat self-documenting and DwC-A consumers are probably sophisticated enough to deal with these changes, while obs export users might be more sensitive to changes, but that could be way off

I agree that consumers of observation export downloads are probably on average more sensitive to changes in terms. I wouldn't expect that DwC-A consumers would have resiliency for these changes built-in, but we can alert direct consumers we are aware of ahead of time in case they need to modify anything on their end. I think being consistent with case is a good thing, and a few one-time changes now to achieve that is probably best in the long run. I guess another option would be to create duplicate entries for the camel-case versions (and plain `captive`) and declare them aliases in their definitions and link to the preferred terms. I don't know that that would be any better than changing them outright

> 3. How are these definitions looking to you?

Definitions look great

PR #4441, opened 2025-03-07
fix: zoom bug in Identify images

### LINKED ISSUE #4424: Small, narrow images are not centered when zoomed in Identify
[//]: # (For feature requests and bug reports not specifically concerning code, please use https://forum.inaturalist.org/c/feature-requests/16. iNat staff will make github issues if/when an actionable solution emerges)

**Describe the bug**
In Identify, when you try to zoom in on an image that is both small and narrow, the zoomed in image is stuck to the left side of the image area.

**To Reproduce**
1. Go to https://www.inaturalist.org/observations/identify?reviewed=any&quality_grade=needs_id%2Cresearch%2Ccasual&id=254745074&place_id=any, or view another observation in identify where the photo is both small and narrow.
1. Click on the observation
1. Click on the photo to zoom
1. The zoomed photo is aligned to the left side of the image area

**Expected behavior**
The zoomed photo should be horizontally centered in the image area.

**Screenshots**
Image

**Context**
 - OS: Mac OS 15.2
 - Browser: Firefox 135.0

**Additional context**
This has been happening for at least a few months, maybe longer, maybe forever. Given that EasyZoom controls the absolute positioning of the zoomed element, this probably requires messing with https://github.com/inaturalist/EasyZoom/tree/click-toggle

jotamayo97:
fix #4424
I'm not a CSS master, so I hope I'm not falling into any bad practices, but this apparently fixes the bug.

I've also found another possible bug. Some images (in my case, it's the first one, but I think it doesn't depend on the order but rather on the image size) don't allow zooming when viewed on my ultrawide monitor, but they do on a smaller monitor. I'm investigating the reason why. Is this behavior familiar to you?

### DIFF
--- a/app/assets/stylesheets/observations/identify/observation_modal.scss
+++ b/app/assets/stylesheets/observations/identify/observation_modal.scss
@@ -409,9 +409,9 @@ body > [role="dialog"] .ObservationModal {
   }
   
   .image-gallery-content .easyzoom-flyout.short img {
-    width: 100%;
+    height: 100%;
   }
-
+  
   .image-gallery-content .easyzoom-flyout.narrow img {
     width: 100%;
   }
--- a/app/assets/stylesheets/observations/identify/observation_modal.scss
+++ b/app/assets/stylesheets/observations/identify/observation_modal.scss
@@ -407,7 +407,15 @@ body > [role="dialog"] .ObservationModal {
     opacity: 1;
     cursor: zoom-out;
   }
+  
+  .image-gallery-content .easyzoom-flyout.short img {
+    width: 100%;
+  }
 
+  .image-gallery-content .easyzoom-flyout.narrow img {
+    width: 100%;
+  }
+  
   .easyzoom-zoomed img {
     visibility: hidden;
   }
--- a/app/assets/stylesheets/shared/easyzoom.scss
+++ b/app/assets/stylesheets/shared/easyzoom.scss
@@ -89,6 +89,13 @@
   cursor: zoom-out;
 }
 
+.image-gallery-content .easyzoom-flyout.short img {
+  width: 100%;
+}
+
+.image-gallery-content .easyzoom-flyout.narrow img {
+  width: 100%;
+}
 .easyzoom-zoomed img {
   visibility: hidden;
 }
--- a/app/assets/stylesheets/shared/easyzoom.scss
+++ b/app/assets/stylesheets/shared/easyzoom.scss
@@ -90,7 +90,7 @@
 }
 
 .image-gallery-content .easyzoom-flyout.short img {
-  width: 100%;
+  height: 100%;
 }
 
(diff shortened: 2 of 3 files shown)


kueda (changes requested):
This makes it so large, landscape-oriented images do not zoom beyond the width of the container. For example, here's what https://www.inaturalist.org/observations/262988360 looks like when zoomed in with Identify in your branch

Screenshot 2025-03-10 at 11 47 00 AM

and in production

Screenshot 2025-03-10 at 11 47 34 AM

Ideally, the fix for this issue should address the case when the photo is narrow and small without changing the behavior for other situations.

kueda:
> I've also found another possible bug. Some images (in my case, it's the first one, but I think it doesn't depend on the order but rather on the image size) don't allow zooming when viewed on my ultrawide monitor, but they do on a smaller monitor. I'm investigating the reason why. Is this behavior familiar to you?

That's probably a function of window size rather than monitor size. If you have repro conditions, please file a new issue for that.

jotamayo97:
I'm really sorry, I should have tested with larger photos!

After spending a looooot of time struggling with the library [EasyZoom](https://github.com/inaturalist/EasyZoom/tree/click-toggle), this is the best solution I've found.

It's likely that a refactor could solve the issue in a better way. For example, in the files of this PR there are some occurrences of:

```
margin-left: auto !important;
margin-right: auto !important;
```

that are thought for the short or narrow classes but they don't seem to be applied, and maybe they could be removed. I could take care of that as part of this task if you think it's a good idea.

This solution shouldn't break other cases...
I really hope I'm not wrong this time! 😅

kueda (changes requested):
This is better, but I'm still seeing some behavior unrelated to the bug that differs from production

- [x] When I view the photo from https://www.inaturalist.org/photos/456366571 in Identify, the first time I click on it it zooms but can't be panned more than a few pixels. When I click on it to zoom out and click on it again, it zooms correctly and I can pan. That doesn't seem to happen in production. Here's what it looks like on that first zoom: Screenshot 2025-03-17 at 12 59 13 PM
 - [ ] Some smaller landscape images like https://www.inaturalist.org/photos/472653274 don't seem to zoom at all, they just fill the width of the image area and get locked to the top: Screenshot 2025-03-17 at 1 01 53 PM. In production it gets locked to the top too but it zooms and I can pan.

jotamayo97:
With this CSS change and this PR: https://github.com/inaturalist/EasyZoom/pull/1, the issue should now be resolved.

I've tested it with various images and everything seems to be working as expected.  
Hopefully I haven’t missed anything, I really don’t want to waste your time.

Thanks again for your attention!

kueda:
@jotamayo97 one more thing: could you please run `npm update EasyZoom` so package-lock.json is updated with the correct commit hash for your change in the easyzoom lib?

PR #4444, opened 2025-03-07
fix: remove notifications about comment when hidden

kueda:
Creating a HIDE ModeratorAction deletes all UpdateActions where the moderated item is a resource or a notifier, and prevents the creation of new UpdateActions when the resource or the notifier is hidden.

One implication is that if the offending content is later unhidden, the people it notified on creation will not be notified. That strikes me as fine, but kind of up for debate. Another approach might be to update the existing UpdateAction and clear out its subscribers from ES, but that seemed excessively complicated.


pleary (approved):
This all seems OK to me

PR #4448, opened 2025-03-12
feat: add high-priority filters for targeting announcements

kueda:
Should not be merged until associated API work is ready too, b/c some of this targeting is only relevant to the mobile placement.

I don't need to a full review ATM (though happy to receive one), but I'd like some input on the following:

1. Do we like yes/no/any strings in the db to model binary states with the option not to be selected? Could also do a nullable boolean, but this seemed more explicit about the meaning of the blank state.
2. Last obs targeting [uses ES](https://github.com/inaturalist/inaturalist/pull/4448/files#diff-dafad8f7f24127720f042c933df7c00fa4455f7ab1264df8da009069de486be7R187-R192): it's certainly more performative than postgres, but is that going to be a performance drain regardless?
3. Is this the best way to do [geoip lookup](https://github.com/inaturalist/inaturalist/pull/4448/files#diff-dafad8f7f24127720f042c933df7c00fa4455f7ab1264df8da009069de486be7R251)? I'm sort of concerned that we'll be doing the geoip lookup twice for these requests, since it will happen again when we log to logstash
4. Are we ok with [loading all countries from the db](https://github.com/inaturalist/inaturalist/pull/4448/files#diff-4029876ea95e38fa8d1aaa7a18d450a613402be7c07adcdff0f0efb75a5e3240R205-R207)?  Could also use the [countries gem](https://github.com/countries/countries) or just make a hard-coded constant somewhere
5. I've been trying to keep my commits somewhat atomic, which means several migrations; is there any reason to consolidate into a single migration before merging?

Closes WEB-622


app/models/place.rb
```diff
@@ -1006,14 +1017,18 @@ def clean_geometry
...
+    if options[:admin_level] == COUNTRY_LEVEL
+      translated_name( place_name, options )
     else
-      display_name
+      options[:display_name] || place_name
     end
   end
 
+  def localized_name
+    Place.localized_name( name, { admin_leve: admin_leve, display_name: display_name } )
```
pleary, line 1029: Maybe typo here with `admin_leve`

kueda:
Many thanks, will fix that bug and look into caching.

> I would probably generally go with a boolean with true/false/null but I don't object to yes/no/any. I dislike any when used as an API parameter, but for something like this as long as intention is clear and there is no performance impact (which there is not here) I don't object to yes/no/any

To elaborate, I agree a nullable boolean makes more sense in postgres, but it really bugs me that null becomes nil in ruby which is then falsey, which leads to bugs like `target_logged_in ? do_logged_in_thing : do_logged_out_thing` in which you accidentally `do_logged_out_thing` when you didn't mean to do any targeting.

Issue #4362, opened 2025-01-29
Allow searches by taxon ID number


tiwane:
See [https://forum.inaturalist.org/t/accept-numerical-taxon-ids-in-searches/60511](https://forum.inaturalist.org/t/accept-numerical-taxon-ids-in-searches/60511)

Request is to allow a user to enter a taxon ID number in places where one can search for taxa. So if I entered `6930` I'd get *Anas platyrhynchos* as a result.

(Added by @pleary on 2024-03-03)

When the searched tern is an integer and only an integer,

* For universal search, return any searched class type as long as its ID matches the searched integer
* For taxon autocomplete search fields, return only the taxon whose ID is equal to the searched integer

kueda:
IMO, this should only work if you *only* enter a number, so `6930 duck` should not look up the taxon. Probably needs to be supported at the API level.

jotamayo97:
Hello! I'm working on this 👍

kueda:
Thanks for your work on this, @jotamayo97!

PR #4451, opened 2025-03-13
Search by Id in autocomplete taxon and general search

jotamayo97:
fix: #4362 

There are two different approaches here. Personally, I would only modify the API, but I have also worked on part of the web approach because the task is in the web project, and this way, I learn more about the iNaturalist environment 😄 .

**API approach:** 
I believe this is the best solution:
https://github.com/inaturalist/iNaturalistAPI/pull/501
https://github.com/inaturalist/iNaturalistAPI/pull/502
I've tested it with many cases; I hope I haven't broken anything.

**Web approach**: 
I can filter before calling the API, but for a general search, this wouldn't work well. However, if for some reason you prefer this way for the taxon autocomplete, I can adapt it for all taxons inputs 👍 .

### DIFF
--- a/app/assets/javascripts/jquery/plugins/inat/taxon_autocomplete.js.erb
+++ b/app/assets/javascripts/jquery/plugins/inat/taxon_autocomplete.js.erb
@@ -189,6 +189,7 @@ $.fn.taxonAutocomplete = function( options ) {
     options.user = CURRENT_USER;
   }
   var source = function( request, response ) {
+    var isIdNumber = !isNaN(request.term);
     var apiToken = $( "meta[name='inaturalist-api-token']" ).attr( "content" );
     var headers = {};
     if ( apiToken) {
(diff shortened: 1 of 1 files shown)


kueda:
While it might be a bit weird for the API to have special behavior for integers, I agree that's probably the better approach here. What do you think, @pleary?

pleary:
With the changes made to the API in https://github.com/inaturalist/iNaturalistAPI/pull/501, where sending an integer to taxon search returns the taxon with that ID, do the changes here make any behavioral difference? This seems unnecessary to me once we merge in https://github.com/inaturalist/iNaturalistAPI/pull/501, so I'd be inclined to say let's not apply these changes. If we do want to make a change like this, we'll need a handful of changes to this PR.

Our JS asset pipeline for JS files outside `app/webpack` currently uses Uglifier as the js_compressor, and that does not currently support converting ES6 syntax. This PR uses the spread operator (...) which is not supported and must be converted to ES5-compatible syntax. That could be done in several ways, for example preparing an object then using conditionals to decide to append new attributes to that object. You can test how the file will compile in the asset pipeline by running `RAILS_ENV=production rake assets:precompile` (`rake assets:clobber` can be used to clean up the files in `public/assets` the former task will create). 

A few other things - the way the ajax `data` attribute is prepared changes behavior and it doesn't act the same when `options.useAPIv2` is used this breaks requests for APIv2 requests. The v2 request here is using `X-HTTP-Method-Override` with a `POST` method, but that isn't needed and is broken when not sending data. Finally a small formatting gripe, but some the ajax options were modified to use 4 spaces instead of 2 which isn't ideal and throws off the diff.

All that said, unless there is a compelling reason to change the JS here, I'd recommend we don't change it and let the API decide to return taxa when searching by ID instead of making a different API call here.

kueda:
Great, I think we all agree that changing the API is the better approach here. I'll close this and we'll go ahead with the API PRs. Thanks all.

Issue #4428, opened 2025-02-28
Hide "Add a Name" button on taxon pages to logged-out users


tiwane:
See https://forum.inaturalist.org/t/add-a-name-appearing-when-not-logged-in/60647 and https://forum.inaturalist.org/t/source-of-sawfly-vernacular-names/62073/4

At least two users on the forum are wondering why they see the "Add a name" button on a taxon page when they're logged out, then see it disappear when they log in (they haven't met the requirements to add names yet). We should probably just hide the "Add a name" button for logged-out users.

PR #4461, opened 2025-03-17
fix: taxon page remove name buttons elements for visitors

HannesOberreiter:
fixes https://github.com/inaturalist/inaturalist/issues/4428

Removed the whole `ul` element, as both buttons depend on a logged-in user. 

## After change

image


pleary:
Thanks!

PR #4505, opened 2025-04-08
feat: support HEIC/HEIF files in the Uploader

kueda:
This is not ready to merge but it is ready to review. IMO we should figure out how to use imagemagick 7 on the production servers before merging this so our claim that we support HEIC is true for HEIC's produced by recent iPhones.

* Uploader accepts HEIC/HEIF files and we convert them to JPGs on the server
* *All* new photos now get converted to JPG
* We needed a fork of kt-paperclip to ensure the `file_content_type` column in photos gets set to `image/jpeg` after the conversion from HEIC to JPEG happens
* Uploader shows a loading spinner while the file is being uploaded; we *could* convert to JPEG in the browser (and in fact are for the CV request), but conversion can be slow, so we'd probably need to show some kind of loading state regardless
* Computer vision demo should work with HEIC, but I bailed on including support for the eval tool

Closes WEB-326


jwidness:
Just an FYI, some platforms do not give `image/heic` or `image/heif` as the MIME type for those files. For example, on win11/chrome, heif files are given the generic MIME type `application/octet-stream`.

kueda:
@jwidness I think that's an acceptable cost. Accepting an image format that most browsers won't even render is already an edge case. Handling images that browser's can't even assign a MIME type to is the edge of the edge.

PR #4516, opened 2025-04-15
feat: link taxon change dashboard update to affected idents

### LINKED ISSUE #4363: Dashboard: link taxon change to user's IDs affected by that change
See https://forum.inaturalist.org/t/view-affected-content-after-taxon-change/29776

When notified about a taxon change in the Dashboard, eg:

Image

If the user has not opted out of accepting taxonomic changes, show them a "View my affected content" button instead of the "Update your content" button. Clicking on "View my affected content" will take the user to https://www.inaturalist.org/identifications but filtered to their IDs affected by that specific change. [See discussion](https://inaturalist.slack.com/archives/C04MF2V4ZAN/p1738107377998309?thread_ts=1738105820.297529&cid=C04MF2V4ZAN).

jotamayo97:
fix #4363 
MERGE FIRST THIS: https://github.com/inaturalist/iNaturalistAPI/pull/510
image
image
image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -6296,7 +6296,8 @@ en:
   view_more_observations_of_x_project: "View more observations from %{project} on <nobr>%{site_name} &raquo;</nobr>"
   view_more_observations_of_x_taxon: "View more observations of %{taxon} on <nobr>%{site_name} &raquo;</nobr>"
   view_more_unreviewed: "View more unreviewed"
-  view_my_affected_content: View my affected content
+  # Link to view identifications affected by a taxon change
+  view_my_affected_identifications: View my affected identifications
   view_observation: View Observation
   view_observations: View Observations
   view_observation_flags: View observation flags
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -6296,6 +6296,7 @@ en:
   view_more_observations_of_x_project: "View more observations from %{project} on <nobr>%{site_name} &raquo;</nobr>"
   view_more_observations_of_x_taxon: "View more observations of %{taxon} on <nobr>%{site_name} &raquo;</nobr>"
   view_more_unreviewed: "View more unreviewed"
+  view_my_affected_content: View my affected content
   view_observation: View Observation
   view_observations: View Observations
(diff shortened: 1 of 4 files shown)


kueda (changes requested):
You still need to do the part where `/identifications?taxon_change_id=123` only shows identifications added automatically by taxon change 123.

jotamayo97:
I’ve added the API call with taxon_change_id attribute and the other button in /taxon_changes.

config/locales/en.yml
```diff
@@ -6296,6 +6296,7 @@ en:
   view_more_observations_of_x_project: "View more observations from %{project} on <nobr>%{site_name} &raquo;</nobr>"
   view_more_observations_of_x_taxon: "View more observations of %{taxon} on <nobr>%{site_name} &raquo;</nobr>"
   view_more_unreviewed: "View more unreviewed"
+  view_my_affected_content: View my affected content
```
kueda, line 6299: Let's change this to "View my affected identifications" since the tool being linked to doesn't show *all* content affected by the change, just the identifications.

Also, please add a comment explaining to translators what this means, e.g. "Link to view identifications automatically added by a taxon change"

kueda (changes requested):
This is working well for me locally, I just want the one text change and this should be good to merge.

config/locales/en.yml
```diff
@@ -6296,6 +6296,8 @@ en:
   view_more_observations_of_x_project: "View more observations from %{project} on <nobr>%{site_name} &raquo;</nobr>"
   view_more_observations_of_x_taxon: "View more observations of %{taxon} on <nobr>%{site_name} &raquo;</nobr>"
   view_more_unreviewed: "View more unreviewed"
+  # Link to view identifications affected by a taxon change
+  view_my_affected_identifications: View my affected identifications
```
kueda, line 6300: Argh, sorry I didn't notice this before: we generally avoid using the first person in UI text like this, so please change this to "View your affected IDs"

kueda (changes requested):

kueda (approved):

Issue #4519, opened 2025-04-15
Identify: marking as unreviewed doesn't seem to work


tiwane:
See [https://forum.inaturalist.org/t/identify-mark-all-as-unreviewed-does-not-work/63547](https://forum.inaturalist.org/t/identify-mark-all-as-unreviewed-does-not-work/63547)

1. Make sure you're on a page that has observations you have reviewed. For example I went to [https://www.inaturalist.org/observations/identify?reviewed=true&order_by=random&verifiable=true&hrank=class&lrank=class&taxon_id=47119](https://www.inaturalist.org/observations/identify?reviewed=true&order_by=random&verifiable=true&hrank=class&lrank=class&taxon_id=47119)
2. Click on "Mark all as Unreviewed"
3. The thumbnails are no longer grayed out, but the number of reviewed observations doesn't change.
4. Use the modal. Observations are still marked as reviewed.

jotamayo97:
I'm working on this :)

PR #4531, opened 2025-04-18
fix: make Mark All As Unreviewed in Identify work

jotamayo97:
fix: #4519 
That was easy hehe


kueda:
Oi, that was probably my mistake. Thanks for fixing it.

PR #4536, opened 2025-04-23
feat: allow customization of projects in header

kueda:
Depends on https://github.com/inaturalist/iNaturalistAPI/pull/514

Closes WEB-509


kueda:
@pleary I realize there's a lot here, but I'm curious what you think about this approach of avoiding yet another one-off CRUD controller and API endpoint. Instead, I've added a custom setter on user that just takes an array of project IDs and does the associated create/update/destroy work. So, some small but unconventional complexity in exchange for another Rails controller, another node API controller, and another inatjs endpoint.

pleary (approved):
This all seems OK to me. Sounds fine to have a custom setter on User and avoid another endpoint. I haven't gone through the rest very thoroughly yet. From a quick test, you might look to see if we want to vertical alignment of the "Remove" button when the project name wraps onto multiple lines. Otherwise looking good to me

PR #4537, opened 2025-04-23
new admin tool to replace an account name

jotamayo97:
This is my largest PR so far, so feel free to request any changes or improvements you think are necessary.

I've opted to add this change to the moderation history as I believe such a modification should be traceable. If you have any suggestions or alternative ideas on how this could be implemented better, I'm more than happy to hear them!

image
image
image

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -68,6 +68,38 @@ class UsersController < ApplicationController
     }
   cache_sweeper :user_sweeper, :only => [:update]
 
+  def replace_username
+    @user = User.find(params[:id])
+    new_username = nil
+    i = 1
+    loop do
+      candidate = "naturalist#{i.to_s.rjust(4, '0')}"
+      unless User.exists?(login: candidate)
+        new_username = candidate
+        break
+      end
+      i += 1
+    end
+    if new_username
+      old_username = @user.login
+      begin
+        ModeratorAction.create!(
+          user: current_user,
+          resource: @user,
+          action: ModeratorAction::REPLACEUSERNAME,
+          reason: "Username changed from #{old_username} to #{new_username}",
+          resource_content: new_username
+        )
+
+        flash[:notice] = "Username successfully replaced with #{new_username}"
+      rescue ActiveRecord::RecordInvalid => e
+        flash[:alert] = flash[:alert] =  e.message
+      end
+    else
+      flash[:error] = "Failed to generate a unique username"
+    end
+    redirect_to user_detail_admin_path(id: params[:id])
+  end
   def suspend
     if @user.suspended?
       flash[:error] = "You cannot suspend someone who is already suspended"
--- a/app/models/moderator_action.rb
+++ b/app/models/moderator_action.rb
@@ -2,11 +2,13 @@
 
 class ModeratorAction < ApplicationRecord
   HIDE = "hide"
+  REPLACEUSERNAME = "replaceUsername"
   UNHIDE = "unhide"
   SUSPEND = "suspend"
   UNSUSPEND = "unsuspend"
   ACTIONS = [
     HIDE,
+    REPLACEUSERNAME,
     SUSPEND,
     UNHIDE,
     UNSUSPEND
@@ -27,6 +29,7 @@ class ModeratorAction < ApplicationRecord
   validates :reason, length: { minimum: MINIMUM_REASON_LENGTH, maximum: MAXIMUM_REASON_LENGTH }
   validate :only_curators_and_staff_can_hide, on: :create
   validate :only_staff_can_make_private
+  validate :only_staff_can_replace_username, on: :create
   validate :only_hidden_content_can_be_private
   validate :only_staff_and_hiding_curator_can_unhide, on: :create
   validate :check_accepted_actions, on: :create
@@ -91,6 +94,13 @@ class ModeratorAction < ApplicationRecord
     errors.add( :base, :only_staff_and_hiding_curators_can_unhide )
   end
 
+  def only_staff_can_replace_username
+    return unless action == REPLACEUSERNAME
+    return if user&.is_admin?
+
+    errors.add( :base, :only_staff_can_replace_username )
+  end
+
   def cannot_suspend_staff
     return unless resource.is_a?( User ) && resource.is_admin? && action == SUSPEND
 
@@ -150,6 +160,7 @@ class ModeratorAction < ApplicationRecord
 
   def set_resource_content
     return unless resource
+    return if action == REPLACEUSERNAME
 
     self.resource_content = Flag.instance_content( resource )
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -16,7 +16,7 @@ class User < ApplicationRecord
   acts_as_voter
   acts_as_spammable fields: [:description],
     comment_type: "signup"
-  has_moderator_actions %w(suspend unsuspend)
+  has_moderator_actions %w(suspend unsuspend replaceUsername)
 
   # If the user has this role, has_role? will always return true
   JEDI_MASTER_ROLE = "admin"
@@ -1590,6 +1590,11 @@ class User < ApplicationRecord
       self.spammer = false
       self.suspended_by_user = nil
       unsuspend!
+    elsif moderator_action.action == ModeratorAction::REPLACEUSERNAME
+      new_login = moderator_action.resource_content.to_s.strip
+      return if new_login.blank? || User.exists?(login: new_login)
+      self.login = new_login
+      save!
     end
   end
 
--- a/app/views/admin/user_detail.html.haml
+++ b/app/views/admin/user_detail.html.haml
@@ -37,10 +37,15 @@
       } );
     } );
 = content_for :extracss do
+  :javascript
+    $('#replaceusername').on('shown.bs.modal', function () {
+      $('.modal-footer .btn-danger', this).focus()
+    })
   :css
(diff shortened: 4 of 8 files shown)


app/models/moderator_action.rb
```diff
@@ -2,11 +2,13 @@
 
 class ModeratorAction < ApplicationRecord
   HIDE = "hide"
+  REPLACEUSERNAME = "replaceUsername"
```
kueda, line 5: IMO it's best to think of these action names as short verbs that might be reused in other contexts, so here I would go with `RESET` or `RENAME`, because you could imagine needing the same action for a project or a journal post with an objectionable title.

app/models/user.rb
```diff
@@ -1590,6 +1590,11 @@ def moderated_with( moderator_action )
       self.spammer = false
       self.suspended_by_user = nil
       unsuspend!
+    elsif moderator_action.action == ModeratorAction::REPLACEUSERNAME
+      new_login = moderator_action.resource_content.to_s.strip
+      return if new_login.blank? || User.exists?(login: new_login)
+      self.login = new_login
+      save!
```
kueda, line 1597: This is where you want to use `User.suggest_login`. I guess we could use `resource_content` to do message passing like this, but it's more of a bookkeeping tool.

app/views/admin/user_detail.html.haml
```diff
@@ -379,3 +386,17 @@
...
+#replaceusername.modal.fade
+  .modal-dialog{ role: "document" }
+    .modal-content
+      = form_tag replace_username_admin_user_path(@display_user), method: :put do
+        .modal-header
+          %button.close{ type: "button", "data-dismiss" => "modal" } ×
+          %h4.modal-title= t(:moderator_actions_replaceUsername)
+        .modal-body
+          %p
+            = t(:replace_user_name_admin_explanation)
```
kueda, line 398: Nothing on the admin page needs to be localized, so don't bother making this translatable. This just adds work for the volunteer translators that no one will ever see.

app/views/admin/user_detail.html.haml
```diff
@@ -379,3 +386,17 @@
...
+            %code naturalistXXXX
+        .modal-footer
+          %button.btn.btn-default{ type: "button", "data-dismiss" => "modal" }= t(:cancel)
+          %button.btn.btn-danger{ type: "submit" }= "Yes, replace"
```
kueda, line 402: Newline at the end of the file would be nice. If you're just adding a plain text node, you can just do this

```haml
%button.btn.btn-danger{ type: "submit" } Yes, replace
```

config/locales/en.yml
```diff
@@ -4956,6 +4958,10 @@ en:
   repair_external_photos: Repair External Photos
   replace_boundary_with_kml: Replace boundary with KML
   replace_parentesis: (replace)
+  # Label for a button that confirms the replacement of the userName
+  replace_user_name_admin_confirmation: Yes, replace
+  # Message displayed in the modal of replacement of the userName as explanation of the action
+  replace_user_name_admin_explanation: Are you sure you want to replace this user's username with an anonymized one? This cannot be undone. The new username will look like
```
kueda, line 4964: Again, we don't want this text to be localized, but if we did, we want to avoid sentence fragments like "The new username will look like". To give translators more control over ordering you would use a variable if the value was genuinely dynamic, but here you would just need to make a translatable string with some HTML.

kueda (changes requested):
This works, I think it just needs some refactoring. Main things are

1. Using a ModeratorAction is good, but the name of that action should be more generic and the effect should happen in the `moderated_with` callback
2. We don't need to add new translatable text in the admin interface b/c it's exclusively used by English-speakers

app/controllers/users_controller.rb
```diff
@@ -68,6 +68,38 @@ class UsersController < ApplicationController
...
+    end
+    if new_username
+      old_username = @user.login
+      begin
+        ModeratorAction.create!(
+          user: current_user,
+          resource: @user,
+          action: ModeratorAction::REPLACEUSERNAME,
+          reason: "Username changed from #{old_username} to #{new_username}",
+          resource_content: new_username
```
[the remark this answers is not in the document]
jotamayo97: I struggled a lot with this :joy:. I need the new name at the moment of creating the ModerationAction (because it's in the reason), so I cannot define it in the callback. However, I still need it in the callback, so either I grab it from the reason (which I don't like) or I send it using another param. That's why I used resource_content. Otherwise, this value would be null in the database, so I don't think I'm 'overwriting' anything. 
Any better ideas are welcome., maybe I'm just getting confused.

app/controllers/users_controller.rb
```diff
@@ -68,6 +68,38 @@ class UsersController < ApplicationController
...
+        )
+
+        flash[:notice] = "Username successfully replaced with #{new_username}"
+      rescue ActiveRecord::RecordInvalid => e
+        flash[:alert] = flash[:alert] =  e.message
+      end
```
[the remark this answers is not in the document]
jotamayo97: I tried to draw the analogy with Java, I see it didn’t work out well. hehe

jotamayo97:
I think I’ve applied everything. If something is missing, please feel free to ask me for any necessary changes!
I can create another PR if this is too messy

app/views/admin/user_detail.html.haml
jotamayo97: why newline?

jotamayo97:
I think I’ve applied everything. If something is missing, please feel free to ask me for any necessary changes!
I can create another PR if this is too messy

app/views/admin/user_detail.html.haml
kueda: At this point mostly just style, but there are some [historical reasons](https://www.baeldung.com/linux/files-end-with-newlines).

app/controllers/users_controller.rb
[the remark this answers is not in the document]
kueda: You don't need it in the reason. The reason the name is getting reset is because it was inappropriate. We don't really care what the new name is b/c it's essentially random. So you could just set the reason to "Inappropriate username: #{old_username}" and that would provide the kind of bookkeeping you're trying to achieve with a `ModeratorAction`.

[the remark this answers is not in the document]
jotamayo97: Oky, I agree, I will change that!

app/controllers/admin_controller.rb
```diff
@@ -145,6 +145,17 @@ def update_user
...
+    if params[:reset_name]
+      @moderator_action = ModeratorAction.new(
+        user: current_user,
+        resource: user,
+        action: ModeratorAction::RENAME,
+        reason: "Inappropriate username: #{user.login} was renamed",
+      )
+      if @moderator_action.save
+        flash[:notice] = "Username successfully renamed"
+      end
```
kueda, line 157: You need to handle the case where the ModeractorAction does *not* save.

app/models/moderator_action.rb
```diff
@@ -150,6 +160,7 @@ def set_resource_user_id
 
   def set_resource_content
     return unless resource
+    return if action == RENAME
```
kueda, line 163: The benefit of performing this task with a ModeratorAction is to preserve some history, so you should probably set `resource_content` to something like `resource.try_methods( :name, :title )` here.

kueda (changes requested):

spec/models/moderator_action_spec.rb
```diff
@@ -262,7 +262,30 @@
...
+
+      it "should rename for a user and persist the action when admin performs it" do
+        user.update!( login: new_login )
+        action
+        user.reload
+        expect( user.login ).to eq( new_login )
+        expect( action ).to be_persisted
+        expect( action.action ).to eq( ModeratorAction::RENAME )
+        expect( action.reason ).to eq( generic_reason )
+      end
```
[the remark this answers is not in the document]
jotamayo97: Here in `moderator_action_spec.rb`, I'm testing things like granular access control, specifically with `only_staff_can_rename`, which lives in the `ModeratorAction `.

That said, it's true that the final consequence of creating a `ModeratorAction `of type `RENAME `relies on the callback in the `User `model, so I will add a test for that behavior in `user_spec.rb.`

If you still think it would be better to move all related tests to `user_spec.rb` for consistency, let me know and I'll adjust accordingly.

Thanks for being patient with so many errors... If you prefer to avoid too much noise in this PR, I can create a new one.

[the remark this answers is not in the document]
kueda: The test below this asserts that only staff can rename a user, which I agree is a good thing to test and this is the file to put it in. *This* test is essentially testing code in user.rb, and given that you added a test to user_spec.rb, this one now seems redundant, so please remove it.

And no worries, these aren't errors, mostly style and future maintenance concerns.

spec/models/user_spec.rb
```diff
@@ -648,6 +648,19 @@
...
   end
 
+  describe "moderated_with" do
+    it "renames the user when given a RENAME action" do
+      user = User.make!( login: "old_login" )
+      moderator_action = ModeratorAction.new(
+        action: ModeratorAction::RENAME,
+        user: User.make!,
+        resource: user
+      )
```
kueda, line 658: Not critical, but if you use the FactoryBot factory you can omit the user since it will create one by default, e.g.

```ruby
moderator_action = build :moderator_action, action: ModeratorAction::RENAME, resource: user
```

kueda (changes requested):

kueda (approved):

Issue #4540, opened 2025-04-24
User settings goes blank after deleting the content of email and username


kueda:
**Describe the bug**
When you delete both the username and email fields, the section goes blank

**To Reproduce**
1. Go to https://www.inaturalist.org/users/edit
1. Delete your username
1. Delete your email
1. Everything in the Profile section is blank

**Expected behavior**
The screen should not blank out. The default HTML required notice might appear, otherwise deleting the content of these inputs shouldn't do anything.

**Screenshots**
<img width="1189" alt="Image" src="https://github.com/user-attachments/assets/7456e78a-e825-4fb1-9575-eba9eef1fab1" />

**Context**
 - OS: MacOS 15.2
 - Browser: Firefox 137.0.2

**Additional context**
Feels like an error boundary not doing its job. FWIW I've only tested with accounts with confirmed email addresses.

jotamayo97:
I'll take this one 👀

PR #4542, opened 2025-04-25
fix: prevent user settings from going blank after deleting email & username

jotamayo97:
fix: #4540
image


kueda (approved):
Looks good to me. FYI we're holding off on merging anything but critical bug fixes until City Nature Challenge is over, so we might not get to this until next week.

jotamayo97:
Wow it’s s cool to feel like “I’m part of that” hahaha

PR #4546, opened 2025-05-02
fix: ensure license changes happen in a delayed job

kueda:
Previously if a user changed their license preference and requested that it be retroactive, the task was happening in the request, which caused it to time out if the user had a lot of observations.

Closes WEB-682

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -675,15 +675,12 @@ class User < ApplicationRecord
   end
 
   def update_photo_licenses
-    puts "update_photo_licenses, preferred_photo_license: #{preferred_photo_license}"
     license_number = Photo.license_number_for_code( preferred_photo_license )
-    puts "update_photo_licenses, license_number: #{license_number}"
     return true unless license_number
 
     Photo.where( "user_id = ? AND type != 'GoogleStreetViewPhoto'", id ).
       update_all( license: license_number, updated_at: Time.now )
     index_observations
-    puts "update_photo_licenses, enqueing photo bucket moving jobs"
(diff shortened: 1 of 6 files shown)


kueda:
@pleary I'm curious if you think the database query to update the licenses needs to be split up. For a user with a lot of IDs like me, it took a while (30 mins?) to complete that one query on staging, which may not be acceptable. The overall process is still running for me after 2+ hours, but most of that is the reindexing work, and I'm guessing there's no way to optimize that.

pleary (approved):
Interesting. Yes I suppose for some users on the extremes with lots of observations or media, if these queries are taking a long time it would be worthwhile splitting them up. Maybe something like fetching the record IDs at once, but chunking them into batches (1k? 10k?), then updating records by batch.

Indexing is currently being done by a single process. We could also split that up into multiple delayed jobs, so if there are multiple delayed jobs workers available, the indexing could be done in parallel by multiple workers. That could happen in a similar way - fetching all IDs, and creating separate jobs for each batch of say 1k observations

I'd say since this code hasn't been touched in a while the splitting up of the update query doesn't need to happen here in this PR as we haven't been noticing cases of performance problems related to it yet. If it doesn't happen here, it'd be great to open a separate ticket for batch updating since we're now aware of it. Splitting up the indexing is a nice-to-have and is in my opinion purely optional as it won't be affecting performance, just total time for these changes to be reflected assuming we have free workers

kueda:
Many thanks. I made an issue for the follow-up work at https://linear.app/inaturalist/issue/WEB-698/split-user-license-updates-into-separate-jobs

Issue #4538, opened 2025-04-24
Account Setting - Content and Display: change link in projects section


tiwane:
On the Content and Display section of Account Settings, under Projects, the last sentence reads "You can read more [here](https://www.inaturalist.org/blog/15450-announcing-changes-to-projects-on-inaturalist)." Let's update the link to our help doc, https://help.inaturalist.org/en/support/solutions/articles/151000176472

jotamayo97:
done!

PR #4549, opened 2025-05-05
feat: link project help in setting to official help docs

jotamayo97:
feat: #4538 
image

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -9528,11 +9528,11 @@ en:
           # lexicon settings that we currently allow
           the_maximum_number_of_lexicons_have_been_added:
             The maximum of 3 common name lexicons have already been added
-        this_only_applies_to_traditional_projects: |
+        this_only_applies_to_traditional_projects2: | 
           This only applies to traditional projects. You can't exclude observations
           from collection or umbrella projects, which are essentially saved searches.
           You can read more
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -9528,11 +9528,11 @@ en:
           # lexicon settings that we currently allow
           the_maximum_number_of_lexicons_have_been_added:
             The maximum of 3 common name lexicons have already been added
-        this_only_applies_to_traditional_projects: |
+        this_only_applies_to_traditional_projects2: |
           This only applies to traditional projects. You can't exclude observations
           from collection or umbrella projects, which are essentially saved searches.
           You can read more
-          <a href="/blog/15450-announcing-changes-to-projects-on-inaturalist">here</a>.
+          <a href="https://help.inaturalist.org/en/support/solutions/articles/151000176472">here</a>.
```
kueda, line 9535: This is super minor, but could you change this to `https://help.inaturalist.org/support/solutions/articles/151000176472`? I should have caught this in the issue, but removing the locale from the URL allows FreshDesk to redirect the user to a localized version of that page based on their browser language *if* that page has been translated into that language. For example, if your browser is using Spanish, it should show a Spanish version of that page.

kueda (changes requested):

config/locales/en.yml
jotamayo97: sure!

PR #4550, opened 2025-05-05
feat: require Community and Observation taxon to match for RG

kueda:
WEB-633

Closing the issue will require reassessing all existing RG obs, with the anticipated impact being ~0.25% of RG obs.

### DIFF
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -1560,10 +1560,10 @@ class Observation < ApplicationRecord
       else
         NEEDS_ID
       end
-    elsif community_taxon_at_species_or_lower?
+    elsif community_taxon_at_species_or_lower? && community_taxon == taxon
       RESEARCH_GRADE
     elsif community_taxon_id && voted_out_of_needs_id?
-      if community_taxon_below_family?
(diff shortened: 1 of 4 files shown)


jotamayo97:
I'm really sorry for the noise, I'm just trying to understand this PR. With this change, the business rules are being modified, and we can now encounter situations like this:
image
where the community agrees on the genus but not on the species-level ID. I suppose all these texts would need to be updated as well, and maybe the "Has ID supported by two or more" item shouldn't be marked as completed in such cases.

I'm just asking to better understand the purpose of this, I'm not judging the PR itself.

kueda:
Yeah, updating the DQA was something I was working on yesterday, just pushed that commit. "Supported by two or more" is still accurate, there's just an additional requirement that the Observation Taxon match the Community Taxon. Currently we're blocked b/c we don't understand some of the ways the Observation Taxon gets set, so we need Scott to weigh in.

loarie:
@pleary - I closed https://github.com/inaturalist/inaturalist/pull/4673 because it looks like this indeed does the same thing plus adds a nice addition to the DQA UI. I tested this branch on staging and this is the behavior we want

PR #4565, opened 2025-05-13
refactor: support ES6 / ES2020 Javascript syntax

kueda:
Switches from uglifier to terser to support modern Javascript syntax. Uglifier supports ES6, but pretty much ends there, so if we want things like optional chaining, we need a different compressor. FWIW, [this seems to be what Rails itself uses these days](https://github.com/rails/sprockets/pull/713).

Closes WEB-704

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -40,7 +40,7 @@ Rails.application.configure do
   # Compress JavaScripts and CSS
   # Choose the compressors to use (if any)
   config.assets.compress = true
-  config.assets.js_compressor = Uglifier.new( mangle: false )
+  config.assets.js_compressor = :terser
   config.assets.css_compressor = :yui
 
   # Don't fallback to assets pipeline if a precompiled asset is missed
--- a/Gemfile
+++ b/Gemfile
@@ -109,7 +109,7 @@ gem "savon" # allow to consume soap services with WSDL
 gem "soundcloud"
 gem "sprockets"
 gem "terrapin"
-gem "uglifier"
+gem "terser"
 gem "utf8-cleaner"
 gem "watu_table_builder", require: "table_builder"
 gem "whenever", require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -340,7 +340,7 @@ GEM
       base64
       simpleidn
     erubi (1.13.1)
-    execjs (2.8.1)
+    execjs (2.10.0)
     exifr (1.3.9)
     exiftool (1.2.4)
       json
@@ -703,13 +703,13 @@ GEM
     temple (0.8.2)
     terrapin (1.0.1)
       climate_control
+    terser (1.2.5)
+      execjs (>= 0.3.0, < 3)
     thor (1.3.2)
     tilt (2.0.10)
     timeout (0.4.3)
     tzinfo (2.0.6)
       concurrent-ruby (~> 1.0)
-    uglifier (4.2.0)
-      execjs (>= 0.3.0, < 3)
     unf (0.1.4)
       unf_ext
     unf_ext (0.0.8)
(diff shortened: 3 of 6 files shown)


config/environments/production.rb
```diff
@@ -40,7 +40,7 @@
   # Compress JavaScripts and CSS
   # Choose the compressors to use (if any)
   config.assets.compress = true
-  config.assets.js_compressor = Uglifier.new( mangle: false )
+  config.assets.js_compressor = :terser
```
pleary, line 43: We can add additional config here, I think with the options listed at https://github.com/ahorek/terser-ruby?tab=readme-ov-file#usage, eg.

```
  config.assets.terser = {
    mangle: false
  }
```

pleary (approved):
I'm OK to try this out - seems like terser will be a better tool going forward. I'm just not 100% sure on the configuration we should start with. By default it does look like terser will mangle class, function, and variable names which we were disabling with Uglifier. We can disable that option.

Terser will also do various code compression by default (remove unused variables for example) and its hard to say if any of the compression features will modify code in such a way that will cause problems.

I guess I'd suggest disabling mangling for the initial release to minimize known changes, and if we spot errors we'll be more confident they are related to compression techniques, or something else in the gem switch. Then if we find things are working well with terser, we could try enabling mangling again. We should check we don't have code that depends on checking method/class/variable names. We should also check to make sure mangled names are consistent if we run the process on different instances, and each instance ends up with the same contents (same resulting fingerprinted JS). I think this last point is among the reasons we disabled it in the past, but that was back when different sites had different instances and potentially different values for some JS variables, which is no longer the case.

There is also no transpiling happening, so if we start to use more modern syntax we may see JS errors in older browsers. We should be mindful of that and maybe set targets and have some documentation on syntax we prefer not to use as it will cause errors in browser versions we would like to support.

config/environments/production.rb
kueda: I'm happy to add this, but are you actually seeing evidence of mangling with the current config and if so where? For example, I just deployed this branch to staging and when I look at `public/assets/messages/new-c645c7e70d9f53c338efe6e792ad9297bd60f7274513042379910073f01e0b16.js` I see

```
$(document).ready((()=>{$("#new_message .user_name").length>0&&$("#new_message .user_name").userAutocomplete({resetOnChange:!1,idEl:$("#new_message .user_name"),afterSelect:function(e){$("#new_message .user-field img").attr({src:e.item.icon}),$("#message_to_user_id").val(e.item.user_id)},afterUnselect:function(){$("#new_message .user-field img").attr({src:$("#new_message .user-field img").data("original-src")}),$("#message_to_user_id").val("")}}),$("#new_message").submit((()=>{$("#new_message").data({submitted:!0})})),window.onbeforeunload=()=>$("#new_message").data("submitted")?null:!_.isEmpty($("#message_subject").val())||!_.isEmpty($("#message_body").val())||null}));
```

which looks completely un-mangled.

kueda:
> There is also no transpiling happening, so if we start to use more modern syntax we may see JS errors in older browsers. We should be mindful of that and maybe set targets and have some documentation on syntax we prefer not to use as it will cause errors in browser versions we would like to support.

This is a good point. Most of the features we want have been widely supported by browsers for almost 10 years at this point, so I'm not too concerned, but we should keep our ears open for edge cases. We've never formally declared any kind of compatibility window, so that might be something to consider too, e.g. we try to support these browsers for this amount of time.

config/environments/production.rb
kueda: Correction: mangling was definitely happening, and with Patrick's suggestion it does not. Also helps to clear out assets with `rake assets:clobber && rake tmp:clear`

PR #4570, opened 2025-05-14
feat: allow filtering observations by whether there are disagreements

kueda:
Adds `disagreements=true` param to the Identify UI to allow people to search for observations that have some disagreement. Requires a new `identification_disagreements_count` attribute in the observations index, so we'll need to reindex ~20M observations, which I estimate will take about 2 weeks with a single process.

In terms of review, I'm never sure if I'm changing the index in the right way or not, so just want to double check. I imagine I'll need to update the production index before deploying like this:

```ruby
Observation.__elasticsearch__.client.indices.put_mapping(
  index: Observation.index_name,
  body: {
    properties: {
      identification_disagreements_count: { type: "integer" }
    }
  }
)
```

Identify integration depends on https://github.com/inaturalist/iNaturalistAPI/pull/517

WEB-508

### DIFF
--- a/app/webpack/observations/identify/components/filters_button.jsx
+++ b/app/webpack/observations/identify/components/filters_button.jsx
@@ -90,15 +90,7 @@ class FiltersButton extends React.Component {
         config.currentUser.curator_projects,
         p => [p.id, p.slug].includes( params.project_id )
       );
-    const FilterCheckboxWrapper = checkbox => (
-      <FilterCheckbox
-        {...Object.assign( {}, checkbox, {
-          defaultParams,
-          params,
-          updateSearchParams
-        } )}
-      />
-    );
+
     const visibleRanks = [
       "kingdom",
       "phylum",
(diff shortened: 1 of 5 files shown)


app/webpack/observations/identify/components/filters_button.jsx
```diff
@@ -185,6 +177,28 @@ class FiltersButton extends React.Component {
...
+        label: I18n.t( "your_observations" ),
+        checked: CURRENT_USER.id
+      }
+    ];
+    if ( viewerIsAdmin ) {
+      lastColCheckboxProps.splice( 2, 0, {
+        param: "disagreements",
+        key: "show-filter-disagreements",
+        label: I18n.t( "disagreements" ),
+        className: "admin"
```
pleary, line 199: The :before label the admin class is adding isn't displaying quite right for me. The admin label overlaps the checkbox label due to its absolute positioning

pleary (approved):
Looks OK to me. Yes that is a correct way to update the mapping, and we'd want to run that before deploying this code so we are sure the index has the mapping type we want.

app/webpack/observations/identify/components/filters_button.jsx
kueda: Yeah, it sort of looked messed up no matter what. At least it makes it really clear it's only meant for testing.

PR #4593, opened 2025-05-25
fix: pluralize leaderboard identifications title for previous month correctly

HannesOberreiter:
fixes: https://forum.inaturalist.org/t/minor-leaderboard-shows-identification-for-month-year-instead-of-actual-date/65554

The change corrects the pluralization of the word "identification" to "identifications", which fixes the wrong translation variable naming. Additionally adjusted the syntax for consistency (consistent key-value pairs with colons instead of hash rockets).

## Example Image (EN)

image

Cheers
Hannes


kueda:
Thanks!

PR #4662, opened 2025-08-19
Canada gaps

loarie:
Pilot showing list of species not in the model observed nearby/same season as participants last obs location to Canada users with over 1k IDs to encourage gap-filling

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1677,7 +1677,7 @@ en:
     # button asking to participate in a pilot
     pilot_participate: "Yes! I'd like to participate."
     # state of button asking to participate in a pilot after participating
-    pilot_stop: "stop participating"
+    pilot_stop: "Stop Participating"
     # link to a blog post
     pilot_read_more: "Read more on our blog."
     # title for a pilot asking users to help fill species gaps by identifiying needs_id observations of species not in the computer vision model
@@ -1685,7 +1685,7 @@ en:
     # description of a pilot explaining the instructions introducing 2 conditions for observations to be included in the list
     gaps_id_pilot_intro: "Help us fill gaps by participating to recieve daily link to observations:"
     # description of a pilot explaining the instructions, the first condition for observations to be included in the list
-    gaps_id_pilot_condition_1: "Not yet included in the Computer Vision model"
+    gaps_id_pilot_condition_1: "Not yet included in the Computer Vision Model"
     # description of a pilot explaining the instructions, the second condition for observations to be included in the list
     gaps_id_pilot_condition_2: "We think you can identify"
     # instructions once people have participated explaining the list of observations
@@ -1709,7 +1709,7 @@ en:
     # instructions once people have participated explaining the list of observations introducing 2 conditions for observations in the list
     gaps_obs_pilot_current_intro: "Here’s your current list of species that are:"
     # instructions once people have participated explaining the list of observations, the first condition for observations in the list
-    gaps_obs_pilot_current_condition_1: "Not yet included in the Computer Vision model"
+    gaps_obs_pilot_current_condition_1: "Not yet included in the Computer Vision Model"
     # instructions once people have participated explaining the list of observations, the second condition for observations in the list
     gaps_obs_pilot_current_condition_2: "Observed near your last location this season"
     # instructions showing the benefit of participating in the pilot
@@ -1717,9 +1717,9 @@ en:
     # label of link to explore species view
     gaps_obs_pilot_browse: "Browse species"
     # instructions once people have participated if the list of observaitons is not ready
-    gaps_obs_pilot_check_back: "Check back tomorrow for a personalized list of species not yet in the Computer Vision model, observed this season. The personalized list will use the location of your most recent observation."
+    gaps_obs_pilot_check_back: "Check back tomorrow for a personalized list of species not yet in the Computer Vision Model, observed this season. The personalized list will use the location of your most recent observation."
     # instructions once people have participated if the list of observaitons is not ready explaining why
-    gaps_obs_pilot_check_back_body: "If you don't see a link there might not be any species matching these criteria near you"
+    gaps_obs_pilot_check_back_body: "If you don't see a link, there might not be any species matching these criteria near you."
     # title for a pilot asking users to help onboard new users by identifying their obs
     identification_pilot_title: "Identification Pilot: Onboarding New Users"
     # instructions explaining the pilot


app/assets/javascripts/users/dashboard.js
```diff
@@ -243,4 +243,27 @@ $( function( ) {
     $( "#participate_section" ).show( );
     $( "#close_needs_id_pilot_panel" ).show( );
   });
+
+  $( "#close_gaps_obs_pilot_panel" ).on( "click", function( e ) {
+    e.preventDefault( );
+    $( "#gaps_obs_pilot_panel" ).hide( );
+    updateSession( { prefers_gaps_obs__pilot: false } );
```
pleary: There's a double underscore here in `prefers_gaps_obs__pilot`, causing this to be an unknown setting, and preventing `prefers_gaps_obs_pilot` from being updated - i.e the user can't hide the panel in a way that persists after page reload

app/models/observation_accuracy_experiment.rb
```diff
@@ -1079,4 +1082,164 @@ def self.top_identifiers
...
+
+  def self.user_eligible_for_gaps_obs_pilot?( user )
+    return false if !gaps_obs_pilot || user.nil? || user.prefers_gaps_obs_pilot == false
+    return true if user.prefers_gaps_obs_pilot == true || user.is_admin?
+
+    user.site.try( :name ) == "iNaturalist Canada"
+  end
+
+  def self.get_modeled_taxa
+    path = "/disk/mnt/nfs/ml_models/models/latest/taxonomy.csv"
```
pleary: This should be moved to a config entry in config.yml. Will need to coordinate with Ops to have that implemented in config templates

app/views/users/dashboard.html.haml
```diff
@@ -152,6 +152,52 @@
                 %p Check back tomorrow for a link to observations that match your identification skills.
               %p.text-right
                 = link_to "stop participating", "#", id: "stop_participating_link", class: "btn-link"
+      - if ObservationAccuracyExperiment.user_eligible_for_gaps_obs_pilot?( current_user )
+        #gaps_obs_pilot_panel.panel.panel-default
+          .panel-heading
+            %a.close{ "data-panel-type" => "gaps_obs_pilot_panel", id: "close_gaps_obs_pilot_panel", style: ( current_user.prefers_gaps_obs_pilot == true ? "display:none;" : "" ) } &#215;
+            %h3.panel-title= "Observation Gaps Pilot: Nearby Wanted Species"
```
pleary: There's no translations here. I see there also aren't with the previous experiments, so it's consistent with that. But if this is going out to users in Canada, is it likely some of them will have a locale of French and would appreciate translations?

app/models/observation_accuracy_experiment.rb
```diff
@@ -1079,4 +1082,164 @@ def self.top_identifiers
...
+
+      modeled_taxon_ids << row[1].to_i
+    end
+
+    modeled_taxon_ids
+  end
+
+  def self.get_observer_location( user_id )
+    observations = INatAPIService.
+      observations( user_id: user_id, order_by: "observed_on", per_page: 1, verifiable: true )
```
pleary: There's an edge case where the observation has a location but has a geoprivacy of private, and this would return nil coordinates. Or the coordinates could be obscured and the obscured location a little far away from the actual location of the latest observation. You could add `obscuration=none` to make sure to grab the latest observation with an unobscured location for a little more precision

app/models/observation_accuracy_experiment.rb
```diff
@@ -1079,4 +1082,164 @@ def self.top_identifiers
...
+
+    last_observation["geojson"]["coordinates"]
+  end
+
+  def self.get_nearby_gaps( lat, lng, month, radius, modeled_taxon_ids )
+    params = { month: month, lat: lat, lng: lng, radius: radius, verifiable: true, rank: "species" }
+    response = INatAPIService.get( "/observations/taxonomy", params )
+
+    raw = response["results"].select {| r | r["rank"] == "species" }
+    ids = raw.map {| r | r["id"] } - modeled_taxon_ids
```
pleary, line 1153: This seems like it will indeed return IDs of species not in the model. Good to keep in mind though it may return species where the genus is a leaf node, but the species isn't yet in the model (though its photos are being used to train at genus). Something to keep in mind for messaging, or in case we want to call those out separately in some way

app/views/users/dashboard.html.haml
```diff
@@ -152,6 +152,52 @@
                 %p Check back tomorrow for a link to observations that match your identification skills.
               %p.text-right
                 = link_to "stop participating", "#", id: "stop_participating_link", class: "btn-link"
+      - if ObservationAccuracyExperiment.user_eligible_for_gaps_obs_pilot?( current_user )
+        #gaps_obs_pilot_panel.panel.panel-default
+          .panel-heading
+            %a.close{ "data-panel-type" => "gaps_obs_pilot_panel", id: "close_gaps_obs_pilot_panel", style: ( current_user.prefers_gaps_obs_pilot == true ? "display:none;" : "" ) } &#215;
```
pleary, line 157: There's [JS that will attempt](https://github.com/inaturalist/inaturalist/blob/main/app/assets/javascripts/users/dashboard.js#L129) to set a user preference to record the user hid this panel, if the user has such a preference. In this case it attempts to set `prefers_hide_gaps_obs_pilot_panel_onboarding`, but that preference doesn't exist. This is offset by the JS added to handle click on `close_gaps_obs_pilot_panel` setting a different preference.

Nothing needs to change, I was just confused by why two settings were trying to be modified and one didn't exist (well both don't exist but only because of the typo adding an extra `_` in `prefers_gaps_obs__pilot`). A little bit of inconsistency and duplication in setting updates

app/models/observation_accuracy_experiment.rb
```diff
@@ -1079,4 +1082,164 @@ def self.top_identifiers
...
+
+    obs_ids = observer.observation_accuracy_samples.pluck( :observation_id )
+    return if obs_ids.blank?
+
+    params = {
+      place_id: "any",
+      id: obs_ids.join( "," ),
+      view: "species"
+    }
+    return if INatAPIService.observations(
```
pleary: Running this API query for each dashboard load will slow things down a little. Not sure if we could do something different like run this when updating the ObservationAccuracySamples and set `validation_count` on the Validator to an integer if there are observations at the time of calculation. Then check `validation_count` here so dashboard loads don't require API calls

app/models/observation_accuracy_experiment.rb
```diff
@@ -1079,4 +1082,164 @@ def self.top_identifiers
...
+    )
+    return unless observer
+
+    obs_ids = observer.observation_accuracy_samples.pluck( :observation_id )
+    return if obs_ids.blank?
+
+    params = {
+      place_id: "any",
+      id: obs_ids.join( "," ),
+      view: "species"
```
pleary, line 1116: This works great as-is. If we're considering future improvements, if we had a place to store the lat/lng used to search for user observations, it might be nice to construct a search using taxon_ids and the lat/lng/radius. Not sure if URLs could get too long here with observation_ids, but there will be fewer and shorter taxon_ids. But it also allows the user to see the centroid and radius used, but also then clear that filter and see where these taxa exist across the world if the commonly observe it different locations.

app/assets/javascripts/users/dashboard.js
loarie: thanks fixed

app/models/observation_accuracy_experiment.rb
loarie: thanks made an issue https://linear.app/inaturalist/issue/OPS-210/add-support-for-configandmodel-taxonomy

app/views/users/dashboard.html.haml
loarie: will add

app/models/observation_accuracy_experiment.rb
loarie: added

app/models/observation_accuracy_experiment.rb
loarie: I think thats ok

app/views/users/dashboard.html.haml
loarie: I added some code so that if a specific handler exists, skip the generic one

app/models/observation_accuracy_experiment.rb
loarie: done

app/models/observation_accuracy_experiment.rb
loarie: agreed - leveraging the ObservationAccuracyExperiment infrastructure for this - but agreed it would be a better fit to use taxon_ids and lat/lng/radiu rather than obs_ids

app/models/observation_accuracy_experiment.rb
pleary: that's completed

app/models/observation_accuracy_experiment.rb
```diff
@@ -1079,4 +1080,320 @@ def self.top_identifiers
...
+      params = {
+        place_id: "any",
+        id: obs_ids.join( "," ),
+        view: "species"
+      }
+      next unless INatAPIService.observations(
+        params.merge( per_page: 0, viewer_id: user_id )
+      ).total_results.zero?
+
+      observer.validation_count = nil
```
pleary: Should `observer.validation_count` be getting set to the count of the previous query? There's a few things here setting `validation_count` to nil, but nothing setting it to anything other than nil that I can see for these two experiments.

app/models/observation_accuracy_experiment.rb
```diff
@@ -1079,4 +1080,320 @@ def self.top_identifiers
...
+        reviewed: "false",
+        quality_grade: "needs_id",
+        place_id: "any",
+        id: obs_ids.join( "," )
+      }
+      next unless INatAPIService.observations(
+        params.merge( per_page: 0, viewer_id: user_id )
+      ).total_results.zero?
+
+      identifier.validation_count = nil
```
pleary: Same comment as above - should `identifier.validation_count` be getting set to the count of the previous query?

config/config.yml.example
```diff
@@ -148,6 +148,8 @@ base: &base
             default: 1234
     help_email: help@inaturalist.org
 
+    model_taxonomy: /disk/mnt/nfs/ml_models/models/latest/taxonomy.csv
```
pleary: can you modify the path in this example? Maybe just make it `taxonomy.csv` or something more generic?

loarie: done

app/models/observation_accuracy_experiment.rb
loarie: these dashboard pilots are leveraging some infrastructure originally build for the ObservationAccuracyExperiments.
The true ObservationAccuracyExperiments make use of validation_counts, but these dashboard pilots don't so its always 0. I'm hijacking it here for that purpose we discussed using nil as an indicator, but it could be anything - would you prefer 1?

pleary: ok. I see the default value is `0`. The `process_iders` and `process_observers` observers methods can re-use validators. So if their `validation_count` was set to nil initially, and those methods were re-run, will anything change `validation_count` to anything other than nil if they have records? It's also a little confusing for `count=0` to mean they have records. I'd say it'd be preferred to just set `validation_count` to the count of the query which allows it to get updated to something other than `nil`, and gives it some meaningful value

PR #4666, opened 2025-08-21
update improving identifications link

cobbettn:
I tested these changes locally and confirmed that the link now points to the current help page. I also tested by supplying the locale query param in the page URL to view locales other than my browser default.

@pleary any further testing you would recommend?


pleary:
I appreciate the attempt to be thorough, but we should only ever edit the `en.yml` file. All other translation files are updated when we sync changes from our translation service CrowdIn, and not modified directly in this codebase. Any changes here to files other than `en.yml` would be overwritten the next time we sync translations. You are welcome to volunteer to create an account to [modify translations in CrowdIn](https://crowdin.com/project/inaturalistweb), otherwise translators will be made aware of the changes once the `en.yml` is modified. Thanks!

cobbettn:
Gotcha, thanks for the explanation! I reverted changes in all other files except en.yml.

pleary:
Looks great - thanks!

PR #4669, opened 2025-08-27
Withdrawn state

loarie:
linear ticket
https://linear.app/inaturalist/issue/WEB-795/add-withdrawn-status-to-taxon-changevs
functional spec:
https://docs.google.com/document/d/194OQCzS946Sb8mYqzU4KhO2DbaCv3nCEmEEpepQ0qB0/edit?tab=t.0

### DIFF
--- a/app/controllers/taxon_changes_controller.rb
+++ b/app/controllers/taxon_changes_controller.rb
@@ -12,7 +12,7 @@ class TaxonChangesController < ApplicationController
 
   def index
     filter_params = params[:filters] || params
-    # NEW: status param for 4-state enum; keep old @committed for legacy URLs/UI
+    # NEW: status param for 3-state enum; keep old @committed for legacy URLs/UI
     @status    = filter_params[:status].presence
     @committed = filter_params[:committed]
     @types = filter_params[:types]
@@ -419,7 +419,6 @@ class TaxonChangesController < ApplicationController
   end
 
   def withdraw
-    @taxon_change = TaxonChange.find( params[:id] )
     if @taxon_change.status_withdrawn!
       flash[:notice] = "Taxon change was withdrawn."
     else
--- a/app/controllers/taxon_changes_controller.rb
+++ b/app/controllers/taxon_changes_controller.rb
@@ -12,6 +12,8 @@ class TaxonChangesController < ApplicationController
 
   def index
     filter_params = params[:filters] || params
+    # NEW: status param for 4-state enum; keep old @committed for legacy URLs/UI
+    @status    = filter_params[:status].presence
     @committed = filter_params[:committed]
     @types = filter_params[:types]
     @types ||= %w(split merge swap stage drop).map{|t| filter_params[t] == "1" ? t : nil}
@@ -31,9 +33,19 @@ class TaxonChangesController < ApplicationController
     @taxon_schemes = TaxonScheme.limit(100).sort_by{ |ts| ts.title }
     
     scope = TaxonChange.all
-    if @committed == 'Yes'
+    # Prefer new status filter if provided
+    if @status.present?
+      allowed_statuses = TaxonChange.statuses.keys
+      if allowed_statuses.include?(@status)
+        scope = scope.where(status: @status)
+      else
+        # ignore unknown status values
+        @status = nil
+      end
+    # Backward-compat: support old committed=Yes/No URLs
+    elsif @committed == "Yes"
       scope = scope.committed
-    elsif @committed == 'No'
+    elsif @committed == "No"
       scope = scope.uncommitted
     end
     scope = scope.types(@types) unless @types.blank?
@@ -71,6 +83,7 @@ class TaxonChangesController < ApplicationController
         taxon_options = { only: [:id, :name, :rank] }
         render json: @taxon_changes.map{|tc|
           json = tc.as_json( methods: [:type ] )
+          json[:status] = tc.status
           json[:input_taxa] = tc.input_taxa.as_json( taxon_options ).compact
           json[:output_taxa] = tc.output_taxa.as_json( taxon_options ).compact
           json
@@ -404,7 +417,27 @@ class TaxonChangesController < ApplicationController
       format.json { render json: response[:json], status: response[:status] }
     end
   end
+
+  def withdraw
+    @taxon_change = TaxonChange.find( params[:id] )
+    if @taxon_change.status_withdrawn!
+      flash[:notice] = "Taxon change was withdrawn."
+    else
+      flash[:error] = "Could not withdraw taxon change."
+    end
+    redirect_to edit_taxon_change_path( @taxon_change )
+  end
   
+  def restore
+    @taxon_change = TaxonChange.find(params[:id])
+    if @taxon_change.update(status: :draft)
+      flash[:notice] = "Taxon change was restored to draft."
+    else
+      flash[:error] = "Could not restore taxon change."
+    end
+    redirect_to taxon_change_path(@taxon_change)
+  end
+
   private
   def load_taxon_change
     render_404 unless @taxon_change = TaxonChange.where(id: params[:id] || params[:taxon_change_id]).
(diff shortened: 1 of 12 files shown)


app/models/taxon_change.rb
```diff
@@ -211,7 +217,7 @@ def commit
         skip_taxon_framework_checks: true
       )
     end
-    update_attribute(:committed_on, Time.now)
+    update_columns(committed_on: Time.current, status: :committed)
```
pleary: I'd double check this change. `update_attribute` [skips validations but runs callbacks](https://apidock.com/rails/v6.1.7.7/ActiveRecord/Persistence/update_attribute) and also updates `updated_at`, `update_columns` [skips both validations and callbacks](https://apidock.com/rails/v6.1.7.7/ActiveRecord/Persistence/update_columns), and does not update `updated_at`. I suspect we want the callbacks to run, namely `commit_records_later` which will queue a job to perform most of the logic of committing a taxon change

loarie: thanks - I think thats what was causing the rspec failures I was looking into - rerunning them now

app/controllers/taxon_changes_controller.rb
```diff
@@ -12,6 +12,8 @@ class TaxonChangesController < ApplicationController
 
   def index
     filter_params = params[:filters] || params
+    # NEW: status param for 4-state enum; keep old @committed for legacy URLs/UI
```
pleary: Does status have 4 states, or just 3 (draft, committed, withdrawn)?

app/controllers/taxon_changes_controller.rb
```diff
@@ -404,7 +417,27 @@ def analyze_ids
...
+    else
+      flash[:error] = "Could not withdraw taxon change."
+    end
+    redirect_to edit_taxon_change_path( @taxon_change )
+  end
   
+  def restore
+    @taxon_change = TaxonChange.find(params[:id])
+    if @taxon_change.update(status: :draft)
+      flash[:notice] = "Taxon change was restored to draft."
```
pleary, line 432: I think we should add a model validation to TaxonChange so that if committed_on is not null, it cannot have a status other than `committed`. Otherwise this endpoint could be called to make an already committed taxon change a draft, and then later the UI is only checking the status column and not the committed_on column, so the UI offers the ability to commit it again. Or the above endpoint can mark a committed change as withdrawn. Since only `status` is being checked for some things and `committed_on` not checked, we should have model validations to ensure the two attributes stay in sync

app/views/taxon_changes/_show_change_taxon.html.haml
```diff
@@ -15,23 +15,31 @@
           = link_to t(:update_your_content), taxon_change_commit_for_user_path(taxon_change), :class => "btn btn-default"
       - if taxon_change.editable_by?(current_user)
         = link_to t(:edit), edit_taxon_change_path(taxon_change), :class => "btn btn-default"
-        - if show_commit_button && !taxon_change.committed? && taxon_change.committable_by?( current_user )
```
pleary, line 18: This is what I was referencing above. We should ensure there is no way to make a TaxonChange with a non-null value of committed_on have a status of draft (or anything other than committed)

app/controllers/taxon_changes_controller.rb
```diff
@@ -404,7 +417,27 @@ def analyze_ids
       format.json { render json: response[:json], status: response[:status] }
     end
   end
+
+  def withdraw
+    @taxon_change = TaxonChange.find( params[:id] )
```
pleary: It isn't necessary to look up the taxon change here or in `restore` below, as it's already defined by the `load_taxon_change` before_action that will be called for these two methods

config/locales/en.yml
```diff
@@ -9276,10 +9276,28 @@ en:
...
       move_children_desc: |
         Children of the input taxon that are genera or coarser will be moved to
         the output; species, subspecies, varieties, etc. will be swapped into
         equivalents with the new genus or species name.
+      # act of changing a taxon change status from withdrawn back to draft meaning reinstated as a draft
+      restore: restore
+      # in process of changing a taxon change status from withdrawn back to draft meaning reinstated as a draft
+      restoring: restoring
+      # act of changing a taxon change status from draft to withdrawn meaning abandoned
+      withdraw: withdraw
```
pleary: This text is also used in a button and should be capitalized

config/locales/en.yml
```diff
@@ -9276,10 +9276,28 @@ en:
...
+      # warning when changing taxon change status from withdrawn back to draft
+      are_you_sure_you_want_to_restore_this_taxon_change: Are you sure you want to restore this taxon change?
+      # warning when changing taxon change status from draft to withdrawn
+      are_you_sure_you_want_to_withdraw_this_taxon_change: Are you sure you want to withdraw this taxon change?
       move_children_desc: |
         Children of the input taxon that are genera or coarser will be moved to
         the output; species, subspecies, varieties, etc. will be swapped into
         equivalents with the new genus or species name.
+      # act of changing a taxon change status from withdrawn back to draft meaning reinstated as a draft
+      restore: restore
```
pleary: This word is used in two places, both buttons. We generally capitalize text in buttons, so this should also be capitalized

app/views/taxon_changes/edit.html.haml
```diff
@@ -3,7 +3,12 @@
     =t :editing
     =t @taxon_change.class.name.underscore
     = surround '(', ')' do
-      = @taxon_change.committed? ? t(:committed_on, date: l( @taxon_change.committed_on ) ) : t(:draft)
+      - if @taxon_change.status_committed?
+        = t(:committed_on, date: l(@taxon_change.committed_on))
+      - elsif @taxon_change.status_withdrawn?
+        = t(:withdrawn)
```
pleary: There is no root-level I18n key of `:withdrawn` - there's one at `views.taxon_changes.withdrawn`

pleary (changes requested):
I left some suggestions. The most important is probably some validations when `status` is set on TaxonChange so that its value stays consistent with the value of `committed_on`, e.g. ensuring committed changed can't get a status other than `committed`

app/controllers/taxon_changes_controller.rb
loarie: just 3, changed

app/views/taxon_changes/edit.html.haml
loarie: changed

config/locales/en.yml
loarie: capitalized

config/locales/en.yml
loarie: capitalized

app/controllers/taxon_changes_controller.rb
loarie: removed

app/views/taxon_changes/_show_change_taxon.html.haml
loarie: see above

app/controllers/taxon_changes_controller.rb
loarie: I added some validations and tests

app/controllers/taxon_changes_controller.rb
```diff
@@ -404,7 +417,26 @@ def analyze_ids
...
+    if @taxon_change.status_withdrawn!
+      flash[:notice] = "Taxon change was withdrawn."
+    else
+      flash[:error] = "Could not withdraw taxon change."
+    end
+    redirect_to edit_taxon_change_path( @taxon_change )
+  end
   
+  def restore
+    @taxon_change = TaxonChange.find(params[:id])
```
pleary: Not a big deal, but this lookup is also unnecessary

loarie:
great thanks - I fixed that minor thing and will merge in when checks pass - no urgency to deploy this

PR #4673, opened 2025-09-03
RG bug

loarie:
fixes https://linear.app/inaturalist/issue/WEB-633/observation-can-be-labeled-research-grade-at-wrong-taxonomic-level
by keeping obs at Needs ID
if the obs.taxon is an infraspecies (e.g. ssp) and there is not community support


loarie:
I'm closing this as it does the same thing as https://github.com/inaturalist/inaturalist/pull/4550 but is missing a nice addition to the DQA UI that that one has

PR #4675, opened 2025-09-05
Logged out header donate button

loarie:
donate button on logged out header for iNat site only
https://linear.app/inaturalist/issue/WEB-804/add-donate-button-to-logged-out-header


pleary (approved):
Looks good to me

PR #4677, opened 2025-09-06
Exclude ip countries and duplicate button for announcements

loarie:
Linear ticket here
https://linear.app/inaturalist/issue/WEB-806/exclude-country-ip-from-announcements

### DIFF
--- a/app/controllers/announcements_controller.rb
+++ b/app/controllers/announcements_controller.rb
@@ -3,9 +3,9 @@
 class AnnouncementsController < ApplicationController
   before_action :authenticate_user!, except: [:active]
   before_action :site_admin_required, except: [:active, :dismiss]
-  before_action :load_announcement, only: [:show, :edit, :update, :destroy, :dismiss]
-  before_action :load_sites, only: [:new, :edit, :create]
-  before_action :load_oauth_applications, only: [:new, :edit, :create]
+  before_action :load_announcement, only: [:show, :edit, :update, :destroy, :dismiss, :duplicate]
+  before_action :load_sites, only: [:new, :edit, :create, :duplicate]
+  before_action :load_oauth_applications, only: [:new, :edit, :create, :duplicate]
 
   layout "bootstrap"
 
@@ -107,6 +107,31 @@ class AnnouncementsController < ApplicationController
       format.any { head :no_content }
       format.html { redirect_back_or_default( dashboard_path ) }
     end
+  end
+
+  def duplicate
+    original = @announcement
+    @announcement = original.dup
+
+    # Explicitly dup array columns so edits don’t mutate the original
+    @announcement.clients                                 = original.clients&.dup
+    @announcement.locales                                 = original.locales&.dup
+    @announcement.ip_countries                            = original.ip_countries&.dup
+    @announcement.exclude_ip_countries                    = original.exclude_ip_countries&.dup
+    @announcement.include_observation_oauth_application_ids = original.include_observation_oauth_application_ids&.dup
+    @announcement.exclude_observation_oauth_application_ids = original.exclude_observation_oauth_application_ids&.dup
+    @announcement.include_virtuous_tags                   = original.include_virtuous_tags&.dup
+    @announcement.exclude_virtuous_tags                   = original.exclude_virtuous_tags&.dup
+
+    # Don’t carry over dismissals; set creator to current user
+    @announcement.dismiss_user_ids = []
+    @announcement.user = current_user
+
+    # Copy site associations
+    @announcement.sites = original.sites
+
+    # Show the form with all fields prefilled (unsaved record)
+    render :new
   end
 
   private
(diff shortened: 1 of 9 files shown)


app/controllers/announcements_controller.rb
```diff
@@ -109,6 +109,31 @@ def dismiss
     end
   end
 
+  def duplicate
+    original = @announcement
+    @announcement = original.dup
+
+    # Explicitly dup array columns so edits don’t mutate the original
+    @announcement.clients                                 = original.clients&.dup
```
pleary: Are these dup lines necessary? If there is a use case for that, does using `deep_dup` instead of `dup` above address it? I think these can all go away

app/controllers/announcements_controller.rb
```diff
@@ -109,6 +109,31 @@ def dismiss
     end
   end
 
+  def duplicate
+    original = @announcement
+    @announcement = original.dup
```
pleary: It would be great if all the duplication logic was in the model, leaving the controller to do just controller things (accept request, pass to models for logic, prepare data for view to be rendered). So this controller method could be something like:
```
  def duplicate
    @announcement = @announcement.duplicate_as_user( current_user )

    # Show the form with all fields prefilled (unsaved record)
    render :new
  end
```

and the model method could be something like:
```
class Announcement < ApplicationRecord
  def duplicate_as_user( duplicate_user )
    announcement = deep_dup

    # Don’t carry over dismissals; set creator to specified user
    announcement.dismiss_user_ids = []
    announcement.user = duplicate_user

    # Copy site associations and preferences
    announcement.sites = sites
    announcement.stored_preferences = stored_preferences

    announcement
  end
end
```

app/controllers/announcements_controller.rb
```diff
@@ -109,6 +109,31 @@ def dismiss
...
+    @announcement.exclude_observation_oauth_application_ids = original.exclude_observation_oauth_application_ids&.dup
+    @announcement.include_virtuous_tags                   = original.include_virtuous_tags&.dup
+    @announcement.exclude_virtuous_tags                   = original.exclude_virtuous_tags&.dup
+
+    # Don’t carry over dismissals; set creator to current user
+    @announcement.dismiss_user_ids = []
+    @announcement.user = current_user
+
+    # Copy site associations
+    @announcement.sites = original.sites
```
pleary: Stored preferences should also be duplicated as there are some filters stored that way: `target_staff`, `target_unconfirmed_users`, `exclude_monthly_supporters`

app/views/announcements/show.html.haml
```diff
@@ -82,6 +86,8 @@
                     - val = @announcement.send( a )
                     - if a == "ip_countries"
                       = val.map { | code | Place::COUNTRIES_BY_NAME.values.detect { | country | country[:code] == code }.try(:[], :name) }.sort.join( ", " )
+                    - elsif a == "exclude_ip_countries"
```
pleary: This is the same logic as the line above, right? Can the above conditional be expanded to capture both attributed? e.g. `if a == "ip_countries" || a == "exclude_ip_countries"`

app/views/announcements/show.html.haml
```diff
@@ -66,7 +67,10 @@
     .col-xs-12
       %span.breadcrumbs
         = link_to t(:back), announcements_path, class: "back crumb"
-      = link_to edit_announcement_path(@announcement), class: "btn btn-primary pull-right" do
+      = link_to duplicate_announcement_path( @announcement ), class: "btn btn-default pull-right" do
+        %i.fa.fa-copy
+        = t :duplicate
```
pleary: There is no translation config for this string

app/views/announcements/show.html.haml
```diff
@@ -66,7 +67,10 @@
     .col-xs-12
       %span.breadcrumbs
         = link_to t(:back), announcements_path, class: "back crumb"
-      = link_to edit_announcement_path(@announcement), class: "btn btn-primary pull-right" do
+      = link_to duplicate_announcement_path( @announcement ), class: "btn btn-default pull-right" do
```
pleary: Minor thing, but these buttons don't have any spacing between them. Usually we'll have a little (see the buttons at the bottom to create/submit/cancel

app/controllers/announcements_controller.rb
loarie: removed

app/controllers/announcements_controller.rb
loarie: refactored

app/controllers/announcements_controller.rb
loarie: added

app/views/announcements/show.html.haml
loarie: added

app/views/announcements/show.html.haml
loarie: restyled

app/views/announcements/show.html.haml
loarie: combined

PR #4678, opened 2025-09-06
Minor direct taxon name changes

loarie:
https://linear.app/inaturalist/issue/WEB-805/trivial-changes-to-species-names

1. allow curators (not just staff) to directly edit taxon.name under strict conditions (no taxa coarser than species, and only minor changes to species epithets for species, minor changes to infraspecies epithets for infraspecies, and changes to the hybrid symbol for hybrids.
2. automatically update taxon_names when changing taxon.name
3. made rubocop happy with taxon.rb and taxon_controller.rb

### DIFF
--- a/app/controllers/taxa_controller.rb
+++ b/app/controllers/taxa_controller.rb
@@ -151,7 +151,12 @@ class TaxaController < ApplicationController
         pagination_headers_for @taxa
         Taxon.preload_associations(
           @taxa,
-          [{ taxon_photos: { photo: :user } }, :taxon_descriptions, { taxon_names: :place_taxon_names }, :iconic_taxon]
+          [
+            { taxon_photos: { photo: :user } },
+            :taxon_descriptions,
+            { taxon_names: :place_taxon_names },
+            :iconic_taxon
+          ]
         )
         options = Taxon.default_json_options
         options[:include].merge!(
--- a/app/models/taxon.rb
+++ b/app/models/taxon.rb
@@ -2634,6 +2634,118 @@ class Taxon < ApplicationRecord
     Taxon.elastic_index!( ids: taxon_ids )
   end
 
+  def self.normalize_ws( str )
+    str.to_s.strip.tr( "\u00A0", " " ).gsub( /\s+/, " " )
+  end
+
+  # Detect hybrid *marker* between Genus and species in verbatim string.
+  # Returns :times (Unicode ×), :x (ASCII x), or :none.
+  def self.hybrid_marker_kind( str )
+    cleaned_str = str.to_s.strip
+    return :times if cleaned_str =~ /\A[A-Z][a-z\.-]*\s+×\s+[a-z-]+/
+    return :x     if cleaned_str =~ /\A[A-Z][a-z\.-]*\s+x\s+[a-z-]+/
+
+    :none
+  end
+
+  # True hybrid *formula*: "Genus species × (Genus) species"
+  # (NOT a named nothospecies like "Citrus × aurantium")
+  def self.hybrid_formula?( str )
+    cleaned_str = Taxon.normalize_ws( str )
+    cleaned_str = cleaned_str.gsub( /\s*×\s*/, " × " ).gsub( /(?<=\s)[xX](?=\s)/, " × " )
+    cleaned_str.sub!( /\A([A-Z][A-Za-z\.-]+)\s+\([A-Z][A-Za-z\.-]+\)\s+/, "\\1 " )
+    /\A[A-Z][A-Za-z\.-]+ [a-z-]+ × (?:[A-Z][A-Za-z\.-]+ )?[a-z-]+(?:\b|\z)/.match?( cleaned_str )
+  end
+
+  # Prefer GNparser, but fall back gracefully if it can’t provide clean tokens/stems.
+  # Returns { genus:, species:, infra:, species_stem:, infra_stem:, ok: true/false }
+  def self.extract_parts( name )
+    parsed = ::Biodiversity::Parser.parse( name.to_s )
+    simple = parsed.dig( :canonical, :simple ).to_s
+    stem   = parsed.dig( :canonical, :stem ).to_s
+
+    tokens = simple.split( /\s+/ )
+    stems  = stem.split( /\s+/ )
+
+    parts = {
+      genus: tokens[0],
+      species: tokens[1],
+      infra: tokens[2],
+      species_stem: stems[1],
+      infra_stem: stems[2]
+    }
+
+    ok = parts[:genus].present? && parts[:species].present? && parts[:species_stem].present?
+    return parts.merge( ok: true ) if ok
+
+    fb = Taxon.fallback_parts( name )
+    parts[:genus]        ||= fb[:genus]
+    parts[:species]      ||= fb[:species]
+    parts[:infra]        ||= fb[:infra]
+    parts[:species_stem] ||= Taxon.fallback_stem( fb[:species] )
+    parts[:infra_stem]   ||= Taxon.fallback_stem( fb[:infra] )
+    parts.merge( ok: false )
+  end
+
+  # Fallback tokenizer if GNparser fails or returns blanks. { genus:, species:, infra: }
+  def self.fallback_parts( name )
+    cleaned_name = Taxon.normalize_ws( name )
+    cleaned_name.sub!( /\A([A-Z][a-z-]+)\s+\([A-Z][a-z-]+\)\s+/, "\\1 " )
+    cleaned_name.gsub!( /\s+[×x]\s+/, " " ) # treat marker as a space
+    parts = cleaned_name.split( /\s+/ )
+    { genus: parts[0], species: parts[1], infra: ( parts[-1] if parts.length >= 3 ) }
+  end
+
+  def self.name_tokens( name )
+    cleaned_name = Taxon.normalize_ws( name )
+    cleaned_name.sub!( /\A([A-Z][A-Za-z\.-]+)\s+\([A-Z][A-Za-z\.-]+\)\s+/, "\\1 " )
+    # normalize only the standalone marker token (won't touch 'max', 'plexippus', etc.)
+    cleaned_name = cleaned_name.gsub( /\s*×\s*/, " " ).gsub( /(?<=\s)[xX](?=\s)/, " " )
+    cleaned_name.split( /\s+/ )
+  end
+
+  # Crude stemmer good enough for small-typo/gender-variant checks
+  def self.fallback_stem( str )
+    str.to_s.downcase.gsub( /[^a-z-]/, "" ).sub( /(ius|us|a|um|is|e|er|es|ii|i)\z/, "" )
+  end
+
+  # “Similar” = identical stems or Damerau-Levenshtein <= 2 (and not blank)
+  def self.epithet_similar_by_stem?( old_stem, new_stem )
+    return false if old_stem.blank? || new_stem.blank?
+    return true  if old_stem == new_stem
+
+    DamerauLevenshtein.distance( old_stem, new_stem ) <= 2
+  end
+
+  # Marker-only normalization exactly between Genus and species; tail must match.
+  def self.marker_only_between_genus_species?( old_str, new_str )
+    old = Taxon.normalize_ws( old_str )
+    new = Taxon.normalize_ws( new_str )
+
+    genus    = "[A-Z][A-Za-z\\.-]+"
+    subgenus = "\\([A-Z][A-Za-z\\.-]+\\)"
+    species  = "[a-z-]+"
+
+    with_marker = /\A(#{genus})\s+(?:#{subgenus}\s+)?(?:×|[xX])\s+(#{species})(.*)\z/
+    no_marker   = /\A(#{genus})\s+(?:#{subgenus}\s+)?(#{species})(.*)\z/
+
+    cmp = lambda do | m, n |
+      g1, sp1, tail1 = m.captures
+      g2, sp2, tail2 = n.captures
+      g1.casecmp?( g2 ) && sp1.casecmp?( sp2 ) && tail1 == tail2
+    end
+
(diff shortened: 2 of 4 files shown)


app/controllers/taxa_controller.rb
```diff
@@ -312,85 +335,110 @@ def new
...
+      if @taxon.errors.added?( :name, :not_similar ) || @taxon.errors.added?( :name, :genus_changed )
+        @taxon.name = @taxon.name_was
+      end
+      flash.now[:error] = if @taxon.errors.of_kind?( :name, :not_similar )
+        t( "views.taxa.name_must_be_similar" )
+      else
+        @taxon.errors.full_messages.to_sentence
+      end
+
+      # Rebuild ivars needed by the edit view
```
pleary: Are these the same variable assignments as in the edit method? Could they be extracted into some private method in the controller that both the `edit` and `update` actions call to avoid repetition?

app/models/taxon.rb
```diff
@@ -2245,17 +2240,17 @@ def detect_clashes( narrowed_parents, child_ids )
         params = { id: id_obs[0..499] }
         obs = INatAPIService.get( "/observations", params ).results
         id_taxa = []
-        obs.each do | o |
+        obs.map do | o |
```
pleary: I see the rubocop warning this was addressing - `rubocop: warning
Style/MapIntoArray - Use 'map' instead of 'each' to map elements into an array.`. The change here doesn't do what they are recommending which is something like `id_taxa = obs.map{ }`. `.each` would still be best here if there is still assignment inside the loop. Same with line 2253/2248 below

app/models/taxon.rb
```diff
@@ -2636,4 +2631,159 @@ def self.index_taxa( taxa )
...
+  def normalize_epithet( str )
+    str.to_s.downcase.gsub( /[^a-z-]/, "" )
+  end
+
+  def gender_variant?( left, right )
+    variants = [%w(us a), %w(us um), %w(a um), %w(is e), %w(er a), %w(er um), %w(es is), %w(i ii)]
+    variants.any? {| from, to | left == right.sub( /#{to}\z/, from ) || right == left.sub( /#{from}\z/, to ) }
+  end
+
+  def levenshtein( source, target )
```
pleary: This might be better located in a shared library and not in the Taxon model as it doesn't relate directly to taxa. There's also a [levenshtein](https://rubygems.org/gems/levenshtein/versions/0.2.2?locale=en
) gem, and a [damerau-levenshtein](https://rubygems.org/gems/damerau-levenshtein) gem. The [damerau-levenshtein](https://en.wikipedia.org/wiki/Damerau%E2%80%93Levenshtein_distance) algorithm treats transposed characters as a single change, something I've used before when comparing names as that is a common error and often helpful to treat that as a single change.

app/models/taxon.rb
```diff
@@ -151,6 +154,9 @@ def set_uuid
   validate :active_parent_if_active
   validate :ancestry_and_active_if_taxon_framework
   validate :cannot_edit_parent_during_content_freeze
+  validate :restrict_name_changes_by_rank,
+    on: :update,
+    unless: -> { validation_context == :bypass_name_restrictions }
```
pleary, line 160: It might be helpful to note here that `bypass_name_restrictions` is only ever used in specs

app/models/taxon.rb
```diff
@@ -2636,4 +2631,159 @@ def self.index_taxa( taxa )
   end
 
   # /Static #################################################################
+
+  private
+
+  # Keeps TaxonName rows in sync with a Taxon name change for the Scientific Names lexicon:
+  def sync_scientific_taxon_names_for_name_change
+    lexicon = "Scientific Names"
```
pleary, line 2755: the code in this method would read more clearly if this variable was named more closely to what it is - `sci_name` or something else reflecting the value not the concept

app/models/taxon.rb
```diff
@@ -2636,4 +2631,159 @@ def self.index_taxa( taxa )
...
+
+      new_row = taxon_names.where( name: name, lexicon: lexicon ).order( :id ).first
+      if new_row
+        new_row.update_columns( is_valid: true, updated_at: Time.current ) unless new_row.is_valid?
+      else
+        new_row = taxon_names.create!( name: name, lexicon: lexicon, is_valid: true )
+      end
+
+      if old_valid && old_valid.id != new_row.id && old_valid.is_valid?
+        old_valid.update_columns( is_valid: false, updated_at: Time.current )
```
pleary: the conditional above validates `old_valid`, but then its being updated with `update_columns` which skips validations and callbacks. Do they need to be skipped? If not, I'd recommend using `.update` instead

app/models/taxon.rb
```diff
@@ -2636,4 +2631,159 @@ def self.index_taxa( taxa )
...
+    old_valid = taxon_names.where( is_valid: true, lexicon: lexicon ).order( :id ).first
+    return if old_valid&.name.to_s == name.to_s
+
+    TaxonName.transaction do
+      # re-fetch inside the transaction to avoid drift
+      old_valid = taxon_names.where( is_valid: true, lexicon: lexicon ).order( :id ).first
+
+      new_row = taxon_names.where( name: name, lexicon: lexicon ).order( :id ).first
+      if new_row
+        new_row.update_columns( is_valid: true, updated_at: Time.current ) unless new_row.is_valid?
```
pleary: The conditional here validates `new_row`, but then its being updated with update_columns which skips validations and callbacks. Do they need to be skipped? If not, I'd recommend using .update instead

I'm also a little confused about the name `new_row`, when its being assigned to an existing name. Is there a more descriptive variable name that can be used to reflect what is is - seeming a TaxonName that matches the Taxon's `name` attribute?

app/models/taxon.rb
```diff
@@ -2636,4 +2631,159 @@ def self.index_taxa( taxa )
...
+    if rank_level && rank_level > 10
+      errors.add( :name, :not_similar, message: "name changes are not allowed for ranks coarser than species" )
+      return
+    end
+
+    old_name, new_name = name_change
+    old_marker = hybrid_marker_kind( old_name ) # :none, :times, :x
+    new_marker = hybrid_marker_kind( new_name )
+
+    o = parse_name_parts( old_name )
```
pleary: We should avoid single-letter variables. Variable names should be easily readable so their intention is clear. `old_name_parts` seems to fit. Doesn't matter how long it is - it is most important for code to be readable and understandable

app/models/taxon.rb
```diff
@@ -2636,4 +2631,159 @@ def self.index_taxa( taxa )
...
+    s = str.to_s.strip
+    return :times if s =~ /\A[A-Z][a-z-]*\s+×\s+[a-z-]+\b/
+    return :x     if s =~ /\A[A-Z][a-z-]*\s+x\s+[a-z-]+\b/
+
+    :none
+  end
+
+  # Parse name into parts, ignoring hybrid marker (×/x) and optional subgenus
+  # Returns { genus:, species:, infra: }
+  def parse_name_parts( str )
```
pleary: A former colleague of mine made a nice scientific name parsing gem called [biodiversity](https://github.com/GlobalNamesArchitecture/biodiversity). Might be worth looking into. In my experience there can be great variation in name strings that produce a lot of edge cases. I know that gem handles a lot of them. If continuing to use this method - I'd check a lot of names to make sure this is capturing name parts correctly on some of the stranger variations

As one example, this is inactive, but the rank of `nothosubsp` on https://www.inaturalist.org/taxa/1478720-Glycine-max-nothosubsp-gracilis is being captured as an infraspecific name, but it's a rank.

We might also want to skip this entirely for virus names which don't follow the same structure as other clades, e.g. https://www.inaturalist.org/taxa/1470951-Emaravirus-sorbi

app/models/taxon.rb
```diff
@@ -2636,4 +2631,159 @@ def self.index_taxa( taxa )
...
+    end
+
+    # ❌ Never allow genus to change
+    unless o[:genus].casecmp?( n[:genus] )
+      errors.add( :name, :genus_changed, message: "genus cannot change" )
+      return
+    end
+
+    case rank_level
+    when 10 # species
```
pleary: This is allowing be to completely change the second epithet in hybrid species names. For example on https://www.inaturalist.org/taxa/780846-Danaus-plexippus---gilippus you can change `gilippus` to anything you want, or even remove `× gilippus` entirely and the change is saved

app/controllers/taxa_controller.rb
```diff
@@ -16,186 +17,194 @@ class TaxaController < ApplicationController
...
         if params[:q].blank? && params[:taxon_id].blank? && params[:place_id].blank? && params[:names].blank?
           @taxa = Taxon::ICONIC_TAXA
         end
         pagination_headers_for @taxa
-        Taxon.preload_associations(@taxa, [
-          { taxon_photos: { photo: :user } }, :taxon_descriptions,
-          { taxon_names: :place_taxon_names }, :iconic_taxon ] )
+        Taxon.preload_associations(
+          @taxa,
+          [{ taxon_photos: { photo: :user } }, :taxon_descriptions, { taxon_names: :place_taxon_names }, :iconic_taxon]
```
pleary: Generally speaking, we've preferred in this codebase to have objects like this be pretty-formatted and span multiple lines. We can see if we can add a Rubocop rule to validate that going forward

app/models/taxon.rb
```diff
@@ -2636,4 +2629,159 @@ def self.index_taxa( taxa )
...
+      # Allow adding/removing Unicode ×; disallow adding ASCII 'x' from none.
+      if old_marker == :none && new_marker == :x
+        errors.add( :name, :not_similar, message: "cannot add 'x' as a hybrid marker between genus and species" )
+        return
+      end
+      # proceed with minor-change rule for specific epithet
+      unless epithet_similar?( o[:species], n[:species] )
```
pleary: Do we want to allow changing the first to characters of a species? Those generally seem more important. For example you could change `dimorpha` to `morpha`. Not sure if we think that is a "minor" change

app/controllers/taxa_controller.rb
loarie: refactored

app/models/taxon.rb
loarie: incorporated the damerau-levenshtein gem

app/models/taxon.rb
loarie: added comment

app/models/taxon.rb
loarie: renamed old_valid variable to sci_name

app/models/taxon.rb
loarie: changed to update

app/models/taxon.rb
loarie: renamed new_row to taxon_name_matching_name

app/models/taxon.rb
loarie: renamed to old_name_parts new_name_parts

app/models/taxon.rb
loarie: incorporated biodiversity gem and refactored

app/models/taxon.rb
loarie: added tests for this case and made modifications to code to catch it

app/models/taxon.rb
loarie: fixed

app/models/taxon.rb
```diff
@@ -2636,4 +2635,310 @@ def self.index_taxa( taxa )
...
+      unless epithet_similar_by_stem?( old_infra_stem, new_infra_stem )
+        errors.add(
+          :name, :not_similar,
+          message: I18n.t( "#{i18n_base}.infra_epithet_similar" )
+        )
+      end
+    end
+  end
+
+  def normalize_ws( str )
```
pleary: I think all the methods from here on aren't relevant to an instance - they accept parameters and act on the parameters, but do not reference anything on the instantiated instance. They should ideally be made into class methods (e.g. `self.normalize_ws`), and invoked on the class (e.g. `Taxon.normalize_ws( str )`)

app/models/taxon.rb
```diff
@@ -2636,4 +2635,310 @@ def self.index_taxa( taxa )
...
+    cleaned_str = normalize_ws( str )
+    cleaned_str = cleaned_str.gsub( /\s*×\s*/, " × " ).gsub( /(?<=\s)[xX](?=\s)/, " × " )
+    cleaned_str.sub!( /\A([A-Z][A-Za-z\.-]+)\s+\([A-Z][A-Za-z\.-]+\)\s+/, "\\1 " )
+    /\A[A-Z][A-Za-z\.-]+ [a-z-]+ × (?:[A-Z][A-Za-z\.-]+ )?[a-z-]+(?:\b|\z)/.match?( cleaned_str )
+  end
+
+  # Prefer GNparser, but fall back gracefully if it can’t provide clean tokens/stems.
+  # Returns { genus:, species:, infra:, species_stem:, infra_stem:, ok: true/false }
+  def extract_parts( name )
+    parsed = Biodiversity::Parser.parse( name.to_s )
```
pleary: At least for me locally in a non-production environment, I saw Ruby complain it couldn't find `Biodiversity` as it relates to the Taxon class. If we prefix with `::`, e.g. `::Biodiversity::Parser.parse( name.to_s )`, it resets to the top-level namespace to look for the `Biodiversity` class. Not sure exactly when that's necessary as I couldn't replicate it later, but it doesn't hurt to add it

loarie: added

app/models/taxon.rb
loarie: changed to class methods

app/models/taxon.rb
loarie: added new check and specs

app/controllers/taxa_controller.rb
loarie: pretty formatted

PR #4679, opened 2025-09-06
pending other curator approval state for plant taxon changes

loarie:
[for plants, t](https://linear.app/inaturalist/issue/WEB-802/pending-and-approved-state-for-taxonchanges)
1. for plants, taxon changes enter pending state on commit, awaiting approving votes from other curators in order to be approved to commit
2. satiated rubocop on taxon_changes.rb and taxon_changes._controller.rb

### DIFF
--- a/app/models/taxon_change.rb
+++ b/app/models/taxon_change.rb
@@ -51,7 +51,7 @@ class TaxonChange < ApplicationRecord
 
   acts_as_votable
   SUBSCRIBABLE = false
-  VOTE_APPROVAL_THRESHOLD = 2
+  VOTE_APPROVAL_THRESHOLD = 5
 
   enum status: {
     draft: "draft",


app/models/taxon_change.rb
```diff
@@ -17,181 +17,221 @@ class TaxonChange < ApplicationRecord
...
   class ActiveChildrenError < StandardError; end
   class RankLevelError < StandardError; end
 
+  def requires_approval?
+    if is_a?( TaxonDrop ) || is_a?( TaxonStage )
+      return false
+    end
+
+    i_t = is_a?( TaxonMerge ) ? input_taxa.first : input_taxon
+    plant_id = Taxon::ICONIC_TAXA.find {| i | i.name == "Plantae" }&.id
```
pleary: We also have `Taxon::ICONIC_TAXA_BY_NAME["Plantae"]`

app/models/taxon_change.rb
```diff
@@ -17,181 +17,221 @@ class TaxonChange < ApplicationRecord
...
   class PermissionError < StandardError; end
   class ActiveChildrenError < StandardError; end
   class RankLevelError < StandardError; end
 
+  def requires_approval?
+    if is_a?( TaxonDrop ) || is_a?( TaxonStage )
+      return false
+    end
+
+    i_t = is_a?( TaxonMerge ) ? input_taxa.first : input_taxon
```
pleary: Just to harp on variable naming and readability a little more - given the next line, I was confusing this below with thinking it meant iconic_taxon. If you do rename it to be more readable, don't assign it to `input_taxa` or `input_taxon` since those attributes already exist

app/models/taxon_change.rb
loarie: changed

app/models/taxon_change.rb
loarie: changed

Issue #3990, opened 2024-01-25
For "Based on the evidence..." DQA question, display community taxon


tiwane:
**Note**: @loarie  says we first need to fix how this behaves, he'll be making an issue. 

From https://forum.inaturalist.org/t/say-what-the-community-taxon-is-in-the-dqa/8836

Currently, this DQA question says

> Based on the evidence, can the Community Taxon still be confirmed or improved? 

You'll need to scroll up on obs detail or change tabs in the Identify modal to see the community taxon. Let's have add a second line below it that displays the Community Taxon, so it shoud look like:

> Based on the evidence, can the Community Taxon still be confirmed or improved? 
> 
> Current Community Taxon: Insects (Class Insecta)

PR #4682, opened 2025-09-19
show current community taxon under DQA (#3990)

kfc35:
I’m not fully sure how internationalization works for the project (is this OK to merge as is? or do translations need to be added first / does a process need to be kicked off?), but here’s the requested change. I’ve tested the changes locally and I’ve provided screenshots for what it looks like on my end (please ignore the broken map; I’ve had to comment out the mapnik stuff for my dev env if that is related at all).

Before DQA is available:
Screenshot 2025-09-18 at 10 28 49 PM
Screenshot 2025-09-18 at 10 28 59 PM

After DQA is available:
Screenshot 2025-09-18 at 10 30 58 PM
Screenshot 2025-09-18 at 10 31 05 PM

Note: The current community taxon is NOT a link currently, but it can be updated to link to `/taxa/${taxon.id}` or whatever else if desired. Just let me know if it should be the whole line or just the taxon name

### DIFF
--- a/app/webpack/observations/show/components/quality_metrics.jsx
+++ b/app/webpack/observations/show/components/quality_metrics.jsx
@@ -147,46 +147,58 @@ class QualityMetrics extends React.Component {
         />
       )
       : null;
-    const currentCommunityTaxon = observation.communityTaxon 
-      ? <>
+    const currentUser = config?.currentUser;
+    const currentCommunityTaxon = observation.communityTaxon
+      ? (
+        <>
           <br />
           <span className="current-community-taxon">
             {
               inatreact.t( "current_community_id", {
-                taxon: <SplitTaxon taxon={observation.communityTaxon} /> 
-              })
+                taxon: <SplitTaxon
+                  taxon={observation.communityTaxon}
+                  user={currentUser}
+                />
+              } )
             }
           </span>
         </>
+      )
       : null;
     return (
       <tr className={`improve${needsIDVoteDisabled ? " disabled" : ""}`}>
         <td className="metric_title" colSpan={3}>
-          <i className="fa fa-gavel" />
-          { I18n.t( "based_on_the_evidence_can_id_be_improved" ) }
-          { currentCommunityTaxon }
-          <div className="inputs">
-            <div className="yes">
-              { checkboxYes }
-              <label
-                htmlFor="improveYes"
-                className={needsIDInfo.mostAgree ? "bold" : ""}
-              >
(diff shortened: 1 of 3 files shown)


app/webpack/observations/show/components/quality_metrics.jsx
```diff
@@ -145,11 +147,24 @@ class QualityMetrics extends React.Component {
         />
       )
       : null;
+    const currentCommunityTaxon = observation.communityTaxon 
+      ? <>
+          <br />
+          <span className="current-community-taxon">
+            {
+              inatreact.t( "current_community_id", {
+                taxon: <SplitTaxon taxon={observation.communityTaxon} />
```
pleary: The `currentUser` should be passed to SplitTaxon so the formatting reflects the users's settings, e.g. show the common before or after the scientific name

app/webpack/observations/show/components/quality_metrics.jsx
```diff
@@ -145,11 +147,24 @@ class QualityMetrics extends React.Component {
         />
       )
       : null;
+    const currentCommunityTaxon = observation.communityTaxon 
+      ? <>
```
pleary: A bit of feedback I got from our designer. It would be best if the `Based on the evidence` text, as well as the community taxon string should both indent to align with `Based` if they span multiple lines. For example on narrower browser windows, and if the taxon name is long or has multiple/long common names. That's a new style change for the `Based on the evidence` text, that will also be applied to this new text

pleary (changes requested):
Requesting a few smaller changes. I also agree that it makes sense to update `communityTaxonAtLeastSubfamily` to reference `observation.community_taxon` in this PR if you want to make that change as well

app/webpack/observations/show/components/quality_metrics.jsx
```diff
@@ -145,33 +147,58 @@ class QualityMetrics extends React.Component {
...
-              <label
-                htmlFor="improveNo"
-                className={needsIDInfo.mostDisagree ? "bold" : ""}
-              >
-                { I18n.t( "no_its_as_good_as_it_can_be" ) }
-              </label>
-              { " " }
-              { votesAgainstCount }
+            <div className="column">
+              { I18n.t( "based_on_the_evidence_can_id_be_improved" ) }
```
kfc35, line 176: nothing below is changed (besides the added `currentCommunityTaxon` line of course); it’s just now wrapped by a `div.column` for the desired styling

Issue #3375, opened 2022-02-07
Make a dedicated detail page for sounds, similar to photo detail pages


tiwane:
From https://forum.inaturalist.org/t/create-sound-detail-page/29108

Each photo has its own page on iNat, eg https://www.inaturalist.org/photos/178482672 which you can get to by clicking the "i" button on obs detail. 

Sounds, however, are set adrift in the ether with no home page. The request is to make a place for each sound, similar to photo pages, and to add an "i" button on obs detail for each sound. The sound page would include

The sound itself, playable, of course
Attribution with sound license
Uploaded by
Associated observations
Length of sound
Filename (visible only to observer)
File format (e.g. audio/mp4)
File URL (e.g. https://static.inaturalist.org/sounds/229366.m4a?1621685446)
Link to flag the sound

kfc35:
I would also like to take a shot at this if y’all would not mind! It’s a little meatier than the last one I took, but I think I can have something up for review within a week.

PR #4684, opened 2025-09-22
Implement a sounds detail page (#3375)

kfc35:
Sound duration is NOT included — it’s not easy metadata, and I think it probably requires some dependency to analyze in order to fetch, so I left it out. Instead, I put file size as a proxy for this.

A lot of the code has been taken and adapted from equivalent PhotosController / show.html.erb for Photos. It might be useful to cross reference these files as you view this pull request.

What a sound looks like (non owner view):
Screenshot 2025-09-21 at 9 16 08 PM

What a hidden sound looks like (non-owner admin view)
Screenshot 2025-09-21 at 9 17 18 PM

After successfully updating the license of a sound (owner view):
Screenshot 2025-09-21 at 9 41 19 PM

Observation information icon has been added to sounds
Screenshot 2025-09-21 at 9 50 11 PM

Note: I have not tested with Soundcloud sounds


app/assets/stylesheets/shared/media_show.scss
```diff
@@ -0,0 +1,75 @@
+th, td {
+  vertical-align: top;
+}
+.mediacell {
```
kfc35, line 4: this file is just moved from `photos/show.css`, and the only thing i changed was the class used here, from `.photocell` -> `.mediacell`. i updated photo’s `show.html.erb` to reflect this change

app/controllers/sounds_controller.rb
```diff
@@ -1,13 +1,23 @@
 # frozen_string_literal: true
 
 class SoundsController < ApplicationController
-  before_action :doorkeeper_authorize!, only: [:create],
+  before_action :doorkeeper_authorize!, only: [:create, :update],
```
kfc35, line 4: these actions are mostly the same as their equivalents in `PhotosController` (except for `:return_here` for `:show` in PhotosController… but it seems `ApplicationController` already defines it so I did not feel the need to repeat it here)

app/controllers/sounds_controller.rb
```diff
@@ -1,13 +1,23 @@
...
+  before_action :curator_required, only: [:hide]
+
+  prepend_around_action :enable_replica, only: [:show]
 
   def show
-    redirect_to @sound.observations.first
+    respond_to do | format |
+      format.html do
+        @observations = @sound.observations.limit( 100 )
+        @flags = @sound.flags
```
kfc35, line 18: afaict `Sound` doesn’t have a taxa field. `SoundcloudSound` does but `LocalSound` does not, so the `taxa` section that is in `PhotosController#show` is not included in `Sound`’s detail page

app/views/sounds/show.html.erb
```diff
@@ -0,0 +1,162 @@
...
+<h2><%= @title %></h2>
+
+<% unless @flags.blank? -%>
+  <div class="notice box">
+    <h3><%=t :heads_up_this_sound_has_been_flagged %></h3>
+    <%= render :partial => "flags/flag", :collection => @flags %>
+    <% if logged_in? && current_user.is_curator? -%>
+      <h4 class="upstacked"><%=t :curators %></h4>
+      <p class="ui">
+        <%=t 'views.sounds.show.view_original_html', :url => sound_url %>
```
kfc35: i noticed that `photos/show.html.erb` has some way to get the original_url on the photo object in its equivalent `flagged` section. just putting this note here this flagged section here on the sound details page deviates from that and just links directly to what the `File URL` would be later on in this page.

app/views/sounds/show.html.erb
```diff
@@ -0,0 +1,162 @@
...
+            <td><%= @sound.file_file_name %></td>
+          </tr>
+        <% end %>
+        <% unless @sound.file_content_type.blank? -%>
+          <tr>
+            <th><%=t :file_format %></th>
+            <td><%= @sound.file_content_type %></td>
+          </tr>
+        <% end %>
+        <% unless sound_url.blank? -%>
```
pleary: This section should work more similarly to [photos/show.html.erb](https://github.com/inaturalist/inaturalist/blob/025a6e7b696730e09e1d07920f2fdd2173fa3dbc/app/views/photos/show.html.erb#L125). If the sound is hidden and not viewable by the current_user, the URL should not be displayed. If the sound is hidden and viewable but the current user, there should be a stub link with no href and a class like `hidden-image` (no `data-size` attribute needed for sounds). There should be an equivalent to [photos/show.js](https://github.com/inaturalist/inaturalist/blob/025a6e7b696730e09e1d07920f2fdd2173fa3dbc/app/assets/javascripts/photos/show.js) that will intercept clicks on that link, make an API call to fetch a temporary URL for the hidden content if the user has access, and open a new window using that temporary URL.

This allows the URLs of hidden content to be consistently hidden if users do not access. It also allows users with access to get temporary access to the hidden content so they can review it and maybe decide it can be un-hidden

pleary (changes requested):
It would be nice to add to the content moderation hiding page, some version of the sound player like we show some [representation of the photo](https://github.com/inaturalist/inaturalist/blob/025a6e7b696730e09e1d07920f2fdd2173fa3dbc/app/views/moderator_actions/hide_content.html.haml#L21), if not hidden. This can help a curator get context for the content they are hiding as they are adding the reason why they are hiding it. If hidden, we might just show a similar `content-hidden` as is in the Photo section.

When adding translations, we do like to add comments with all newer strings. These are visible to translators and helps provide additional context with the translation which can be very helpful for non-native English speakers when the English text may be ambiguous or confusing. For example `Updated sound` may not be obvious what is trying to convey. Indeed `Updated photo` has no such context, but that's an older string and if we were adding it today I'd recommend that it should have a comment.

Aside from these things, and the comment I left inline, I think this is generally looking great. I'll run it by some other folks internally to see what they think. But if these changes are made, we can likely merge this soon

spec/views/sounds/show.html.erb_spec.rb
```diff
@@ -0,0 +1,269 @@
...
+            href: hide_sound_path(s) )
+        end
+
+        it "can hide if content is unhidden and user is admin" do
+          sign_in make_admin
+          
+          render
+
+          expect( rendered ).not_to have_tag( "a", 
+            text: t( :flag_this_sound ),
```
kfc35: note to self: fix this test from its copypasta

spec/controllers/moderator_actions_controller_spec.rb
```diff
@@ -43,79 +43,104 @@
     end
   end
 
-  describe "resource_url" do
```
kfc35, line 31: the git diff makes this hard to read but i basically extracted out common logic for the resource_url tests into a `shared_examples`.  Now the tests are run as a suite with different types of media (localphoto, localsound, and soundcloudsound)

app/views/photos/show.html.erb
```diff
@@ -5,7 +5,7 @@
   <%= stylesheet_link_tag "photos/show" %>
 <%- end -%>
 <%- content_for( :extrajs ) do -%>
-  <%= javascript_include_tag "photos/show" %>
+  <%= javascript_include_tag "media/show" %>
```
kfc35, line 8: extracted out `photos/show.js` into `media/show.js` that is used by both photos and sounds. `hidden-image` below has been changed to `hidden-media-link` as a consequence

app/views/sounds/show.html.erb
```diff
@@ -4,10 +4,13 @@
...
   <%= stylesheet_link_tag "sounds/show" %>
 <%- end -%>
+<%- content_for( :extrajs ) do -%>
+  <%= javascript_include_tag "media/show" %>
+<%- end -%>
 <%-
   sound_hidden = @sound.hidden?
   last_moderator_action = @sound.moderator_actions.last
-  sound_url = if @sound.is_a?( SoundcloudSound ) 
+  direct_sound_url = if @sound.is_a?( SoundcloudSound )
```
kfc35, line 13: i realized `sound_url` is a rails url helper function, so I renamed this variable to differentiate

kfc35:
@pleary I’ve incorporated all of your requested changes, so this is ready for review again

Hide content page for sound has the sound player available
Screenshot 2025-09-25 at 10 05 19 AM

Unhide content page for sound has a content hidden blockquote. The link redirects to the sound’s detail page
Screenshot 2025-09-25 at 10 05 35 AM

A hidden sound viewed from the public
Screenshot 2025-09-25 at 10 33 36 AM

Owner view of the hidden sound (with actions visible)
Screenshot 2025-09-25 at 10 07 39 AM

app/views/sounds/show.html.erb
```diff
@@ -113,10 +116,18 @@
...
         <% end %>
-        <% unless sound_url.blank? -%>
+        <% unless sound_hidden && !@sound.hidden_content_viewable_by?( current_user ) -%>
           <tr>
             <th><%=t :file_url %></th>
-            <td><%= link_to sound_url, sound_url %></td>
+            <% if sound_hidden -%>
+              <td>
+                <a href="" class="hidden-media-link" data-moderator-action-id=<%= last_moderator_action.id %>>
+                  <%= direct_sound_url %>
```
kfc35, line 125: figured it makes sense to still show the `direct_sound_url` as the file url, even though clicking it now will use the `presigned_url` from that ajax call to `"/moderator_actions/" + $( this ).data( "moderatorActionId" ) + "/resource_url”`… but it could feel weird since clicking this url is technically taking you to a different url. If you have a better idea for what the content of this anchor tag should be in spite of this, let me know

app/assets/javascripts/media/show.js
```diff
@@ -1,13 +1,19 @@
 $( function ( ) {
-  $( "a.hidden-image" ).on( "click", function ( e ) {
+  $( "a.hidden-media-link" ).on( "click", function ( e ) {
     e.preventDefault( );
     // open a new window now in the context of the user action so it doesn't get popup-blocked
     var resourceWindow = window.open( );
+    var resource_data;
```
pleary: I would recommend enabling automated linting in the whatever editor you are using. We have an [.eslintrc](https://github.com/inaturalist/inaturalist/blob/a6ed12941087f39d446bb6b585912944d440d9ee/app/assets/javascripts/.eslintrc) config for ESLint specifying some preferred ways to format JS. There are a few ways the changes here conflict with those rules. If your editor cannot implement in-line linting (this is a very useful feature that I'd highly recommend), you can manually run eslint with e.g. `node_modules/.bin/eslint app/assets/javascripts/media/show.js`. Could you please address the errors ESLint is reporting?

app/controllers/flags_controller.rb
```diff
@@ -133,6 +133,7 @@ def new
     @flag = Flag.new(params[:flag])
     @object = find_object
     @object = @object.becomes(Photo) if @object.is_a?(Photo)
+    @object = @object.becomes(Sound) if @object.is_a?(Sound)
```
pleary: Similar to ESLint, we use Rubocop for Ruby code formatting validation. There is a [.rubocop.yml](https://github.com/inaturalist/inaturalist/blob/a6ed12941087f39d446bb6b585912944d440d9ee/.rubocop.yml) specifying preferred Ruby code formatting rules. Editors can be configured to show rubocop validation errors inline (again, highly recommended), or it can be run manually, e.g. `rubocop app/controllers/flags_controller.rb`. We have not yet taken the time to address all ruby validation errors in this codebase, so you will see errors not related to code you added or modified. There is no need to update those - we will hopefully soon do a pass on the entire codebase so we can automate this with commit hooks going forward.

But it would be great if any new code was compliant with rubocop rules. Could you please check the code you added and update formatting to be compliant?

pleary (changes requested):
Thanks for making those changes. I think this is the last bit - I am requesting some formatting updates to make JS and Ruby code compliant with our validation rules. Thanks!

kfc35:
@pleary Thanks for letting me know about the linting / validation. Hopefully will no longer be an issue going forward :)

I went through all the files and fixed warnings related to this code. It should be all ready now, but I did notice that the `application_helper` test was failing in the most recent run? It’s not failing when I’m running it locally so I’m not completely sure what the problem is, but I’m on standby if I have to look deeper into it

app/webpack/observations/show/components/photo_browser.jsx
```diff
@@ -471,6 +471,11 @@ class PhotoBrowser extends React.Component {
                   </button>
                   )
                 }
+                <a href={`/sounds/${sound.id}`} title={I18n.t( "view_details" )} label={I18n.t( "view_details" )}>
```
kfc35, line 474: added the label and title due to eslint’ing for accessibility. it seemed like the most appropriate title to put for this icon

kfc35:
after running the tests locally now with my most recent commit, this is good to go!

app/views/shared/_license_form.html.erb
```diff
@@ -1,9 +1,12 @@
 <%-
-  record ||= @observation || @photo
-  record = record.becomes(Photo) if record.is_a?(Photo)
+  record ||= @observation || @photo || @sound
+  record = record.becomes( Phot ) if record.is_a?( Photo )
```
pleary: Typo here introduced in a recent commit. `record.becomes( Phot )` should be `record.becomes( Photo )`

pleary (changes requested):
One more requested change

app/views/shared/_license_form.html.erb
kfc35: 🤦‍♂️ fixed

pleary:
Thanks for all the work on this!

Issue #4326, opened 2024-12-17
Taxon page photo modal - for CC0 photos, credit the account that posted the photo


tiwane:
See https://forum.inaturalist.org/t/cc0-observations-could-should-be-also-credited-in-taxon-view/58325

If a taxon photo is licensed CC0, we display this on the photo's modal:

![image](https://github.com/user-attachments/assets/898569c1-9a62-4aa5-955e-fe9dc9e2f158)

If the photo is CC-BY or more restrictive, we display this, which credits the photographer:

<img width="384" alt="image" src="https://github.com/user-attachments/assets/937ced55-5038-4610-8ae5-7e3ab8778085" />

The ask is to credit the photographer for CC0 taxon photos, so instead of "no rights reserved (CC0)", display 

>by [name or username] - no rights reserved (CC0)

pleary:
This should be resolved going forward. The fix requires reindexing all taxa with CC0 photos, which is currently ongoing. Should be done within the next 24 hours

PR #4685, opened 2025-09-23
fix: taxa grid, updates userName extraction

nobeeakon:
This PR updates the `photo.attribution` user name extraction. 

Added 2 new cases: extract when the string contains "uploaded by", or when it is under a "public domain" licencing

Original report https://forum.inaturalist.org/t/cc0-photos-authorship-showing-up-as-unknown/70568/2


pleary:
A similar problem was [reported on taxon pages](https://github.com/inaturalist/inaturalist/issues/4326), which was addressed in https://github.com/inaturalist/inaturalist/pull/4553/files. One issue with the photo attribution string is that it is localized, mixing the structured data we need, e.g. the user name, with other natural language. In these views, we are then attempting to extract out the user name to change the presentation of the rest of the attribution. A change made in #4553 (and the related API commits [8a984bf7ab4b02830f3beb315efbf34ac01913c2](https://github.com/inaturalist/iNaturalistAPI/commit/8a984bf7ab4b02830f3beb315efbf34ac01913c2) and validation fix in [86486677e4d0fbef3226266c7c3b6dbf1aaf27e5](https://github.com/inaturalist/iNaturalistAPI/commit/86486677e4d0fbef3226266c7c3b6dbf1aaf27e5)), was to store the photo attribution name separately so it can be used directly when attribution needs to be generated when not using `photo.attribution` directly.

I should have realized explore would suffer from the same problem and applied the same change there. I have no objection with the implementation here. I would recommend though that we add a check for `photo.attribution_name` before the block modified here, similar to how [‎app/webpack/taxa/shared/util.js](https://github.com/inaturalist/inaturalist/pull/4553/files#diff-8c023a63dca1fd5f6c09b9b6ac5881b796cf91aa42891a2dbe3fa1ea6df3214a) was modified in the above referenced commit. This will allow us to fetch the photo `attribution_name` directly without any parsing, if it is available. In order for this to work correctly for explore, we'll also need to specify this field needs to be returned by API v2 by adding `attribution_name` to the [taxon fields requested](https://github.com/inaturalist/inaturalist/blob/e1b4c0037b820d63fef01f92fddc19146849cc76/app/assets/javascripts/ang/factories/observations.js.erb#L73).

With that addition in place, we should be able to merge and release this. Thanks!

app/assets/javascripts/ang/models/taxon.js.erb
```diff
@@ -65,9 +65,15 @@ iNatModels.Taxon.prototype.localizedPhotoAttribution = function ( opts ) {
     userName = user.name || user.login || userName;
   }
   if ( userName.length === 0 && photo.attribution ) {
```
pleary, line 67: Requested adding a block above this one, similar to a previous commit in [‎app/webpack/taxa/shared/util.js](https://github.com/inaturalist/inaturalist/pull/4553/files#diff-8c023a63dca1fd5f6c09b9b6ac5881b796cf91aa42891a2dbe3fa1ea6df3214a)

nobeeakon: done, added the check and update the taxon requested fields

Issue #4012, opened 2024-02-20
Prevent link alteration in translations


kueda:
- [ ] Add a validation to our translation spec that ensures the value of an `href` attribute in HTML matches the source value exactly
- [ ] Alter any translation that accepts a URL variable but allows the translator to edit the `<a>` tag so that the variable is the entire tag (there are several existing examples of this style, usually in the form of `start_tag` and `end_tag` variables)
- [ ] Consider altering the `t` ruby and js methods to insert `rel="noopener"` into all <a> tags. This seems like it might be hard and not worth it, so maybe time box it

elysiumplain:
JS Regex expression should handle item#1 to break into fragments with link exclusion.

For item#2 can you clarify demo input and desired output?

kueda:
> For item#2 can you clarify demo input and desired output?

Sure. If the translation looks like

```yaml
some_key: This is <a href="%{url}">a link</a>
```

and gets used like `t( "some_key", { url: "https://www.inaturalist.org" } )`, we'd want to replace it with

```yaml
some_key: This is %{start_tag}a link%{end_tag}
```

which would get used more like `t( "some_key", { start_tag: "<a href='https://www.inaturalist.org'>", end_tag: "</a>" } )`

> Shall we append the solution to the existing spec audit for locales in TranslationsController? 

No. That controller is just for listing available locales. This would be a part of https://github.com/inaturalist/inaturalist/blob/1b8b27493b2fa9938dda5bfe72c9959638528907/spec/lib/translations_spec.rb

elysiumplain:
Thanks, will tackle this in a bit after reading up on Crowdin and architecture.

For item#1:
- Are there any special markup tags in the pipeline that I may need to account for if I were to apply a universal tag and attribute processor?
  - e.g., preprocessor with component/variable expansion
 
Otherwise I propose tokenizing standardized DOM tags, calling t(), and handling attribute translation separately. (for example, translating attribute-embedded text such as `alt=`, etc...)

Per your example above, this should facilitate moving away from the rigid key/value parameter call of t()?

kueda:
> Are there any special markup tags in the pipeline that I may need to account for if I were to apply a universal tag and attribute processor? 

Hm, I don't think we have special XML-ish tags, but there are inflections, e.g. `@gender{m:un|f:une|n:un(e)} %{resource}`. FWIW, I'm not sure you need to write some kind of custom HTML parser in the spec. Nokogiri will probably work fine.

Issue #3660, opened 2023-02-18
Add hard-coded equivalents of untralsnatable translation keys


kueda:
The activity notification notification text is some of the hardest text to translate on the site, and in many (most?) cases it is impossible to translate. For example, a key like `user_added_a_x_to_noun_by_user_html: "%{user} added a %{x} to %{noun} by %{by}"` has _four_ variables. `%{noun}` potentially comes with an article, `%{user}` and `%{by}` represent people whose gender we don't know, and there's no way for translators to accommodate things like vowel / consonant inflections for any of those variables. Even though this code is wretchedly complicated, the reason these strings exist is largely for developer convenience, and to accommodate new kinds of activity updates, but frankly the latter almost never happens, and the impossibility of translation trumps convenience. I've been waiting for an overhaul of notifications to fix these problems, but it's not clear if/when that's going to happen, so let's just try to improve this, if not fix it.

- [ ] Review `user_added_*` translation keys and add hard-coded equivalents whenever possible, e.g. `%{user} added an identification to an observation by %{by}` instead of `%{user} added an %{x} to %{noun} by %{by}`
- [ ] Remove the `activity_snipped_resource_with_indefinite_article` and associated code; this is the cause of most of the trouble
- [ ] Refactor `activity_snippet` so it either doesn't use the `noun` param or only uses it as a fallback
- [ ] Try to test as many kinds of notifications texts as possible

Apologies in advance for anyone attempting this, including future me. The notification text code is not pleasant.

PR #4692, opened 2025-09-29
Added Comments and Tests around notf translation code to Refactor

kfc35:
This probably has lower priority than other things the team is working on. Review at your leisure; there are some questions that are probably non-trivial to answer so I will not be expecting answers right away. I may look into something else in the mean time.

Link to issue: https://github.com/inaturalist/inaturalist/issues/3660

This pull request contains :
- code comments clarifying logic in the `update_tagline_for` function 
- tests around logic that kueda expressed concern about 
- a potential, up-for-discussion plan for how most hard-coded equivalents of the more templated strings would be created

Once this is refined and merged, I will proceed with whatever decisions are made regarding the refactoring.

The points up for discussion: 
- ~~**Do we still need to support `update_tagline_for` with a `:skip_links` option?** It is only called with that option in what I would guess is an un-used non-partial view called `new_updates.html.erb`. If the view is still in use or needed, I think the calls can be replaced with the appropriate links anyway since they are wrapped in separate link generating logic. **If we don’t need to support it, I can simplify the hard-coded equivalents I need to make, removing the translations without links.**~~ ~~This seems to actually be in use when a user is looking at their notifications… which I think makes things complicated if we want to have different translation text with an anchor tag embedded. Any ideas on how to consolidate so that we aren’t increasing the amount of strings we need to translate two-fold?~~ Actually I think I will just go with the approach of extracting out the start and end tag to conform with this desire from issue #4012 , and no-link versions can just have empty strings for the start and end tags. So, there will not need to be two forms of each string to support no-link
- **Omitting 'in a comment': Bug or Feature?** Currently, as I understand it, when user A mentions user B in a comment under User C’s content (let’s say a post), it will appear in B’s update feed as “User A mentioned you in a post by User C.” This is something documented in the tests I wrote. However, it’s more accurate to say that "User A mentioned you in a comment to a post by User C.” We have an opportunity to make this more accurate statement with the refactoring. Is this something we want to address?
- **The genitive (? I think genitive?) case translation: 1 string, 2 strings, or 3 strings?** The `activity_snippet` code appends the authorship for the `noun` depending on a few factors, resulting in three different possible texts. For example: “mentioned you in a post by Some User Login” / “mentioned you in a post by you” / “mentioned you in a post”. I am not a language expert, but I would think there is an increase in translation accuracy if you separate these possible authorship situations into these three different strings (genitive by second person, genitive by third person, and omitting the genitive phrase). However, that’s a lot of strings! Any input confirmation on how accurate we want these cases translated is welcome. It could be one string like “mentioned you in a post${optional_by}”, two strings with “mentioned you in a post” / “mentioned you in a post by {you_or_owner}”, or the aforementioned three strings. For what it's worth, currently it is broken out into three strings.

There are some other smaller questions in the comments I made throughout the pull request that’d be nice to have answered (i.e. what if the resource model’s user and notifier model’s user match, but they’re not the same model (when figuring out the `noun_by_x` text)? why is that `is_admin` if block there? can some code blocks be removed that I suspect are not executed?).

Also, if there’s no appetite to do some of the refactoring proposed in the issue now that some details are more clearly laid out, that is good to know as well.

Thanks for taking a look!

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -6334,17 +6334,11 @@ en:
   user1_user2_and_x_others_faved_this_observation: "%{user1}, %{user2}, and %{x_others} faved this observation"
   user_added_a_comment_html: "%{user} added a comment"
   # Notification text to be seen by subscribers on their Updates feed for a specific circumstance
-  user_added_a_comment_to_assessment_section_by_you_html: '%{user} added a comment to an <a href="${url}" rel="nofollow">assessment section (${title})</a> by you'
+  user_added_a_comment_to_assessment_section_by_you_html: '%{user} added a comment to an %{start_tag}assessment section (%{title})%{end_tag} by you'
   # Notification text to be seen by subscribers on their Updates feed for a specific circumstance
-  user_added_a_comment_to_assessment_section_by_you_no_link_html: '%{user} added a comment to an assessment section (${title}) by you'
+  user_added_a_comment_to_assessment_section_by_owner_html: '%{user} added a comment to an %{start_tag}assessment section (%{title})%{end_tag} by %{owner}'
   # Notification text to be seen by subscribers on their Updates feed for a specific circumstance
-  user_added_a_comment_to_assessment_section_by_owner_html: '%{user} added a comment to an <a href="${url}" rel="nofollow">assessment section (${title})</a> by %{owner}'
-  # Notification text to be seen by subscribers on their Updates feed for a specific circumstance
-  user_added_a_comment_to_assessment_section_by_owner_no_link_html: '%{user} added a comment to an assessment section (${title}) by %{owner}'
-  # Notification text to be seen by subscribers on their Updates feed for a specific circumstance
-  user_added_a_comment_to_assessment_section_html: '%{user} added a comment to an <a href="${url}" rel="nofollow">assessment section (${title})</a>'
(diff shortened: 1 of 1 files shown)


config/locales/en.yml
```diff
@@ -3225,6 +3225,18 @@ en:
   # As in members of a project
   members: Members
   membership_model: Membership Model
+  # Notification text to be seen by a user on their Updates Feed after being mentioned in a specific circumstance
```
kfc35, line 3228: i wrote these just as a short illustration of the strings i would need to write. i’d have to write many combinations for comments if desired

config/locales/en.yml
```diff
@@ -3225,6 +3225,18 @@ en:
...
+  mentioned_you_in_a_comment_on_an_assessment_section_by_you_html: '%{user} mentioned you in a comment to an <a href="${url}" rel="nofollow">assessment section (${title})</a> by you'
+  # Notification text to be seen by a user on their Updates Feed after being mentioned in a specific circumstance
+  mentioned_you_in_a_comment_on_an_assessment_section_by_you_no_link_html: '%{user} mentioned you in a comment to an assessment section (${title}) by you'
```
kfc35, line 3231: this is the more complex “mentioned you in a comment to an assessment section” text as opposed to “mentioned you in an assessment section” that it would normally be now

PR #4695, opened 2025-10-09
add citation recommendation to new taxon flag form and bootsrapify

loarie:
https://linear.app/inaturalist/team/ML/cycle/active
Screenshot 2025-10-09 at 3 46 09 PM
Screenshot 2025-10-09 at 3 46 02 PM

### DIFF
--- a/app/views/flags/new.html.haml
+++ b/app/views/flags/new.html.haml
@@ -11,7 +11,7 @@
     .col-xs-12
       %ol.breadcrumb
         %li
-          = link_to t(:back_to_x, noun: @object.try_methods( :to_plain_s, :to_s ) ), @object
+          = link_to t(:back_to_x, noun: @object.try_methods( :to_plain_s, :to_s ) ), @object, class: "back crumb"
       %h2
         = t :new_flag_for
         = link_to @object.to_plain_s, @object


app/views/flags/show.html.haml
```diff
@@ -25,9 +25,9 @@
 .container
   .row
     .col-xs-12
-      .breadcrumbs
-        %strong
-          = link_to t(:back_to_flags_html), url_for_referrer_or_default( @object )
+      %ol.breadcrumb
+        %li
+          = link_to t(:back), url_for_referrer_or_default( @object )
```
pleary: pretty minor, but if this link is updated given `, class: "back crumb"`, then the CSS will show the left angled quotes like it used to have embedded in its translation

app/controllers/flags_controller.rb
pleary: `{ render layout: "bootstrap" }` could be removed now that the default layout is "bootstrap"

pleary: { render layout: "bootstrap" } could be removed now that the default layout is "bootstrap"

app/views/flags/new.html.haml
```diff
@@ -1,3 +1,5 @@
+- content_for :title do
+  = @title = "#{t :new_flag_for} #{@object.to_plain_s}"
 - content_for :extracss do
   = stylesheet_link_tag "flags/new"
```
pleary: The "flags/new" stylesheet has been removed, so we should remove this `extracss` block

pleary (changes requested):
One small requested change - to remove the inclusion of a CSS file that no longer exists. The other comments are minor and optional. The rest looks good to me, so I'd say we can merge and deploy once this fix is made

app/views/flags/show.html.haml
loarie: added

app/views/flags/new.html.haml
loarie: removed

app/controllers/flags_controller.rb
loarie: removed

loarie: removed

PR #4701, opened 2025-10-17
feat: add color banding on DQA for accessibility - small change

kfc35:
Seemed small to add and is based off of a forum post… I don’t think this will get much pushback but I leave it up to the team to decide whether to include.

https://forum.inaturalist.org/t/add-row-banding-to-the-data-quality-assessment/71680

implementing for annotations seems a little more complicated so I’m leaving that aside for now

Screenshot 2025-10-17 at 4 54 45 PM

### DIFF
--- a/app/assets/stylesheets/shared/quality_metrics.scss
+++ b/app/assets/stylesheets/shared/quality_metrics.scss
@@ -33,7 +33,6 @@
     }
   }
   tr.improve {
-    background: $light-grey;
     .inputs {
       margin-top: 5px;
       .yes, .no {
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/shared/quality_metrics.scss
```diff
@@ -33,7 +33,6 @@
     }
   }
   tr.improve {
-    background: $light-grey;
```
kfc35, line 36: superceded by the banding

pleary:
Looks good. Thank you!

PR #4708, opened 2025-10-28
feat: run rubocop during CI, WEB-825

pleary:
Currently very limited in scope to `./app/controllers/users ./app/controllers/admin ./app/es_indices`. We should expand this value over time as more code is made compliant with rubocop rules


app/es_indices/taxon_photo_index.rb
```diff
@@ -101,8 +101,8 @@ def self.prune_batch_for_index( batch )
       # and the same photo and version, no need to reindex
       next unless indexed_doc
       next if indexed_doc["embedding"].blank?
-      next unless indexed_doc["ancestor_ids"] &&
-        indexed_doc["ancestor_ids"].sort == taxon_photo&.taxon&.self_and_ancestor_ids&.sort
```
FLGMwt, line 105: Asking because this is a quasi-functional change: is this being removed because they're already asserted on https://github.com/inaturalist/inaturalist/pull/4708/files#diff-612ecf98c550e4b3492aae13ab8106020b71524a2b05fde850c49a0d523536afR97 ?

app/es_indices/taxon_index.rb
```diff
@@ -184,38 +190,39 @@ def as_indexed_json(options={})
...
+      json[:statuses] = conservation_statuses.map( &:as_indexed_json )
+      json[:extinct] = conservation_statuses.any? do | cs |
+        cs.place_id.blank? && cs.iucn == Taxon::IUCN_EXTINCT
       end
-      json[:statuses] = conservation_statuses.map(&:as_indexed_json)
-      json[:extinct] = conservation_statuses.select{|cs| cs.place_id.blank? && cs.iucn == Taxon::IUCN_EXTINCT }.size > 0
     end
     # indexing originating from Taxa
     unless options[:for_observation] || options[:no_details]
-      flag_counts = Hash[flags.group_by{ |t| t.resolved }.map{ |k,v| [k, v.length] }]
```
FLGMwt, line 210: oooh, was this an automated fix? nice.

app/es_indices/controlled_term_index.rb
```diff
@@ -44,24 +48,25 @@ class ControlledTerm < ApplicationRecord
...
     else
-      fields_to_remove << "blocking"
+      "blocking"
     end
     # splatten out the array with *
-    json = self.attributes.except(*fields_to_remove)
-    if values.length > 0
-      json[:values] = values.select(&:active).map{ |v| v.as_indexed_json(is_value: true) }
+    json = attributes.except( *fields_to_remove )
+    unless values.empty?
```
FLGMwt, line 66: is this one style-rule enforced? 100% personal preference but this seems more confusing than the original

.github/workflows/CI-build-test.yml
```diff
@@ -93,5 +93,8 @@ jobs:
         TRAVIS_CI: true
         CI: true
 
+    - name: Run restricted rubocop
+      run: bundle exec rubocop ./app/controllers/users ./app/controllers/admin ./app/es_indices
```
FLGMwt, line 97: Since this is going to grow, can we move the explicit allow-list to the `.rubocop.yml` config? https://docs.rubocop.org/rubocop/configuration.html#includingexcluding-files

```yml
AllCops:
  Include:
    - 'app/controllers/users/**/*'
    - 'app/controllers/admin'
    - 'app/es_indices'
```

FLGMwt (commented):
can't speak authoritatively as I'm just getting back up to speed with Ruby, just wanted to ask some clarifying questions here.

As far as I can tell though and to the minimal degree that it helps, I don't notice any functional changes

app/es_indices/taxon_photo_index.rb
pleary: This link doesn't seem to be bringing me to the right place, so I'm not sure what you're referencing. The error bring addressed was that the safe navigation (`&.`) chain was too long. The original functionality was skip to the next iteration unless the indexed doc has ancestor IDs, and their sorted values are the same as the taxon photos taxon sorted ancestors (which would be nil if there was no taxon photo, taxon photo taxon, or taxon photo taxon ancestors).

The change made was to just split up the safe navigation chain and check for taxon photo taxon first, then its ancestry next. I don't think there's any functionality change here as I don't think there's any condition where the behavior would change (if the indexed doc ancestor ids are nil, it'll immediately `next` - if the taxon photo taxon is nil, it'll `next`, and if the doc ancestors don't match the taxon photo taxon ancestors, it'll `next`)

app/es_indices/controlled_term_index.rb
pleary: You're saying `unless values.empty?` is more confusing than `if values.length > 0`. This is rule enforced, and is pretty typical syntax for Ruby code. We can always change the rules. I've always personally struggled a little with `unless`, but it's common in Ruby and used extensively in this codebase

.github/workflows/CI-build-test.yml
pleary: I still want my local IDE to run rubocop on all files, and I want to make it easy for people to manually run rubocop on all files while developing. I only want _CI_ to run rubocop on these files, so I think this is the appropriate place for this allow list. Eventually we'll update syntax on all files and we will no longer need an allow list

app/es_indices/taxon_index.rb
pleary: This was a manual fix. I checked, and rubocop can make this change with using `-A` for safe and unsafe autocorrects

PR #4713, opened 2025-11-01
feat: adds place filter for top identifiers in obs/show

### LINKED ISSUE #3644: Obs show: in the Top Identifier section, add a place filter

kfc35:
This new place filter appears for any observation with an observation taxon that is as specific as or more specific than “class” and does not have a “private” geoprivacy. 

I am aware that the GitHub issue asks:
> The request is that, for observations identified from family to genus, we add a place filter 

But I think having the freedom to choose between continent and global at any allowed taxonomic rank gives enough flexibility for that requirement to be waived… And I am inclined to agree with Hussell’s response...  But I am not a taxonomist so if there is a strong reason to enforce this for whatever reason, just let me know.

You can filter by Global (the previous default) and by any standard place defined on the Observation. The default filtered place is by continent if defined, and global if a continent is not defined (I assume that would not happen in production, but in my local test data, I only have countries and states so… I DID test the default `placeId` functionality by stubbing in the country admin_level value, so I am confident it would work for continents)

I am open to any input from your team designer. Please mind that I placed the filters within the collapsible panel since it does not make sense for users to fiddle with it when the panel is closed.

Closes #3644 

Screenshots! I basically made an observation for a made up species: one based in China and one based in the US. I made the one in China research grade with two identifications, so those same identifiers show up in the Global & China filters, but not for US and the US-State filter.

Under the observation in China:
Screenshot 2025-10-31 at 11 34 59 PM

Screenshot 2025-10-31 at 11 35 09 PM

Screenshot 2025-10-31 at 11 35 12 PM

Screenshot 2025-10-31 at 11 35 17 PM

Under the observation in the US:

Screenshot 2025-10-31 at 11 58 12 PM

Screenshot 2025-10-31 at 11 58 20 PM

Screenshot 2025-10-31 at 11 58 24 PM

Screenshot 2025-10-31 at 11 58 28 PM

When I change the US observation to private, it reverts to the regular view (global identifiers)

Screenshot 2025-10-31 at 11 58 55 PM

Screenshot 2025-10-31 at 11 58 59 PM

### DIFF
--- a/app/webpack/observations/show/components/identifiers.jsx
+++ b/app/webpack/observations/show/components/identifiers.jsx
@@ -1,36 +1,122 @@
 import _ from "lodash";
 import React from "react";
 import PropTypes from "prop-types";
-import { Panel } from "react-bootstrap";
+import { Dropdown, MenuItem, Panel } from "react-bootstrap";
 import UserWithIcon from "./user_with_icon";
+import util from "../util";
 
 class Identifiers extends React.Component {
   constructor( props ) {
     super( props );
     const currentUser = props.config && props.config.currentUser;
     this.state = {
-      open: currentUser ? !currentUser.prefers_hide_obs_show_identifiers : false
+      open: currentUser ? !currentUser.prefers_hide_obs_show_identifiers : false,
+      placeId: null
     };
   }
 
   componentDidMount( ) {
-    this.loadIdentifiers( );
+    this.loadIdentifiers( this.state.placeId );
   }
 
-  componentDidUpdate( ) {
-    this.loadIdentifiers( );
+  componentDidUpdate( prevProps ) {
+    const { observation, observationPlaces } = this.props;
+    if ( !observation || !observation.taxon || observation.taxon.rank_level > 50 || observation.geoprivacy === "private" ) {
+      return;
+    }
+    /* In the constructor, observationPlaces is not defined yet (can be an empty array).
+      We wait to set an initial value for placeId when props.observationPlaces is properly init'd.
+      This react hook activates when that happens, and should execute this if-block only once */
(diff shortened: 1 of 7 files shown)


app/webpack/observations/show/components/identifiers.jsx
```diff
@@ -1,36 +1,122 @@
...
+    }
+  }
+
+  renderPlaceSelector( ) {
+    const { observation, observationPlaces, fetchTaxonIdentifiers } = this.props;
+    if ( !observation || !observation.taxon || observation.taxon.rank_level > 50 || observation.geoprivacy === "private" ) {
+      return ( <span /> );
+    }
+    const standardPlaces = !observationPlaces ? []
+      : _.filter( observationPlaces, p => p.admin_level !== null )
```
kfc35, line 57: Should this also filter out any places that are more specific than County? (e.g. Town and Park)

PR #4722, opened 2025-11-17
chore: update syntax for helpers, related to WEB-825

pleary:


app/helpers/observations_helper.rb
```diff
@@ -55,20 +61,22 @@ def observation_place_guess(observation, options = {})
...
     osm_coords_url += "&zoom=#{observation.map_scale}" unless observation.map_scale.blank?
-    osm_coords_link = link_to("OSM", osm_coords_url, :target => "_blank", rel: "noopener noreferrer")
-    
-    if coordinate_truncation = options[:truncate_coordinates]
-      coordinate_truncation = 6 unless coordinate_truncation.is_a?(Integer)
-      display_lat = display_lat.to_s[0..coordinate_truncation] + "..." unless display_lat.blank?
-      display_lon = display_lon.to_s[0..coordinate_truncation] + "..." unless display_lon.blank?
+    osm_coords_link = link_to( "OSM", osm_coords_url, target: "_blank", rel: "noopener noreferrer" )
+
+    if ( coordinate_truncation = options[:truncate_coordinates] )
```
kfc35, line 74: I’m surprised this is valid to have in an `if` statement. In `irb` I get a warning about doing something like this (e.g. `if (x = 6) then puts “hello” end`

app/helpers/observations_helper.rb
```diff
@@ -77,77 +85,76 @@ def observation_place_guess(observation, options = {})
...
+        "YYYY-MM-DD or YYYY-MM-DD HH:MM:SS"
+      when "latitude"
+        "dd.dddd (latitude) or ddddddd (northing)"
+      when "longitude"
+        "dd.dddd (longitude) or ddddddd (easting)"
+      when "boolean"
+        "yes or no"
+      when "list"
+        "limited set of options, usually alphanumeric"
+      when "numeric"
```
kfc35, line 146: removing ‘number’ here is intentional?

pleary: Yes. The `when 'number'` case had no value, thus was returning nil. That was also the `else` case return, and also the default value for no match. Both `when 'number'` and `else` are unnecessary as the default return is nil, so both can be removed

app/helpers/observations_helper.rb
pleary: This is valid, though maybe not great syntax. Some older code in this codebase does this a lot - assign a variable in an if statement and assess the result of the assignment at the same time. We could separate the assignment on an earlier line, and assess the assigned variable separately, but the end result will be the same.

app/helpers/observations_helper.rb
kfc35: ah, i see. i misread and thought `when “number”` and `when “numeric”` were intended to be handled the same and would both return “positive whole number” with the lack of a `break` or anything after the “number” case… my java habits are showing (where you need explicit breaks)

pleary: This may have been the original intent of the original author. Admittedly I did not explore further to see if there were bugs related to it. My intention was to update syntax without changing current functionality. It may be worth a little extra time to dig into this further to see if returning "positive whole number" for "number" might be more appropriate. Though it was worked as-is for 11 years now

PR #4724, opened 2025-11-20
Provisional names

loarie:
Allow provisional names for Fungi - more details here
https://linear.app/inaturalist/issue/WEB-655/add-provisional-name-bool-for-fungi
and here
https://docs.google.com/presentation/d/1Xb3KwjNFK-UhDrt1Mk_7FE1ps6Z35dGvKdDYScN3FTU/edit?slide=id.g38dcdb3fd7f_0_58#slide=id.g38dcdb3fd7f_0_58

corresponding API changes here https://github.com/inaturalist/iNaturalistAPI/pull/548

Admin only for now


app/controllers/taxa_controller.rb
```diff
@@ -1795,6 +1795,10 @@ def presave
         params[:taxon][:photos_locked] != @taxon.photos_locked
       params[:taxon].delete( :photos_locked )
     end
+    if !current_user.is_admin? && params[:taxon][:provisional] &&
+        params[:taxon][:provisional] != @taxon.provisional
+      params[:taxon].delete( :provisional )
```
pleary, line 1798: Anyone that can't create or edit provisional taxa should not be able to include this parameter at all. Right now, it seems that only admins have that privilege. Instead of hard-coding `current_user.is_admin?`, I'd create a static method on Taxon like `Taxon.user_can_edit_provisional_taxa?( current_user )`, and have that do the logic (e.g. return false unless given user is an admin). That method can be re-used in the view, where it is also hard coded to use `current_user.is_admin?`. That way if the group of people who can edit provisional taxa is expanded, we only need to update the logic in that method once, and all uses of that method will inherit the change.

I'd also make it so that for anyone without permission, we delete the provisional parameter regardless of its value (e.g. don't check if it has a value or compare it to what the current value is). The value could be `false` in which case the block as it stands would be skipped, which isn't what we want (same can be said of `photos_locked` above, but that's out of scope)

app/models/taxon.rb
```diff
@@ -132,18 +132,24 @@ def set_uuid
   validates_uniqueness_of :name, scope: %i[ancestry is_active],
     unless: lambda {
       # If a complex, allow duplicate sibling name for species...
-      ancestry.blank? || !is_active || complex?
+      # Skip validation only if currently provisional AND not changing to non-provisional
+      skip_for_provisional = provisional && ( !will_save_change_to_provisional? || new_record? )
```
pleary, line 136: Why not validate uniqueness of new provisional names? I can't imagine we want any taxa with duplicate names, provisional or not

app/models/taxon.rb
```diff
@@ -132,18 +132,24 @@ def set_uuid
...
+      skip_for_provisional || ancestry.blank? || !is_active || complex?
     },
     message: "already used as a child of this taxon's parent"
   validates_uniqueness_of :name, scope: %i[ancestry is_active rank],
-    if: -> { complex? }, # ... but do not allow duplicate sibling complex name
+    if: lambda {
+      complex? && !( provisional && ( !will_save_change_to_provisional? || new_record? ) )
```
pleary, line 142: Same here - why not validate uniqueness of new provisional names in this other context? If we do want this, the message needs to be updated as it is specific to taxa of rank `complex`

app/models/taxon.rb
```diff
@@ -132,18 +132,24 @@ def set_uuid
...
+    }, # ... but do not allow duplicate sibling complex name
     message: "already used as a child complex of this taxon's parent"
   validates :name,
     format: { with: TaxonName::SCIENTIFIC_NAME_FORMAT, message: :bad_format },
-    if: proc {| t | t.new_record? || t.name_changed? }
+    unless: proc {| t | t.provisional && ( !t.will_save_change_to_provisional? || t.new_record? ) },
```
pleary: I highly recommend some validation of provisional names. There are many character we _don't_ want in names, and I think very few additions we might want to allow for provisional names. I'm OK skipping this specific validation for provisional names, but I think we should add a new validation only for provisional names. Perhaps the only needed addition to name format is allowing numbers within words

app/models/taxon.rb
```diff
@@ -1129,6 +1132,23 @@ def taxon_cant_be_its_own_ancestor
     errors.add( :base, "can't be its own ancestor" )
   end
 
+  def provisional_only_for_fungi
+    return unless provisional
+    return if new_record? && provisional_was.nil? # allow for new records
+    return unless provisional && ( new_record? || will_save_change_to_provisional? )
```
pleary: The first line already checks if the name is provisional, so that extra condition here is unnecessary

app/models/taxon.rb
```diff
@@ -1129,6 +1132,23 @@ def taxon_cant_be_its_own_ancestor
     errors.add( :base, "can't be its own ancestor" )
   end
 
+  def provisional_only_for_fungi
+    return unless provisional
+    return if new_record? && provisional_was.nil? # allow for new records
```
pleary: What is `provisional_was.nil?` doing here? I don't think it will ever be nil because the default is `false`. Can this just check for `new_record?`

app/models/taxon.rb
```diff
@@ -1129,6 +1132,23 @@ def taxon_cant_be_its_own_ancestor
...
+    return unless provisional && ( new_record? || will_save_change_to_provisional? )
+
+    # Find the Fungi taxon
+    fungi_taxon = Taxon.find_by( name: "Fungi" )
```
pleary: I think it would be preferred to use `Taxon::ICONIC_TAXA_BY_NAME["Fungi"]` here rather than looking up the taxon in real time, and only looking it up by name

app/models/taxon_name.rb
```diff
@@ -32,8 +32,13 @@ class TaxonName < ApplicationRecord
   SCIENTIFIC_NAME_FORMAT = /\A([A-z]|\s|-|×)+\z/
   validates :name,
     format: { with: SCIENTIFIC_NAME_FORMAT, message: :bad_format },
+    unless: proc {| tn |
```
pleary, line 40: I similarly feel pretty strong we should have some validation of provisional taxon_names, which can happen in a separate validation

app/models/taxon_name.rb
```diff
@@ -387,14 +391,14 @@ def species_common_name_cannot_match_taxon_name
   end
 
   def valid_scientific_name_must_match_taxon_name
-    return unless is_valid? && is_scientific_names? && taxon && name != taxon.name
+    return unless is_valid? && is_scientific_names? && taxon && name != taxon.name && !taxon.provisional
```
pleary: Do you have an example of why this method needs to be modified? Seems to me this could still hold true for provisional taxa. This raises a question I have - the names of provisional taxa are being recorded as "scientific names". Is that what we want, or should we have a "provisional name" type? How do we want the data record if a provisional taxa later gets named? Do we want the original provisional name and the accepts scientific name to both be called "scientific names"?

app/views/taxa/_form.html.erb
```diff
@@ -215,6 +215,14 @@
     <div class="description">
       <%= t :inactive_taxon_concepts_are %>
     </div>
+    <% if current_user.is_admin? %>
```
pleary: I'd replace with e.g. `Taxon.user_can_edit_provisional_taxa?( current_user )`

app/webpack/taxa/shared/ducks/taxon.js
```diff
@@ -374,6 +374,7 @@ export function fetchTaxon( taxon, options = { params: { } } ) {
     if ( testingApiV2 ) {
       params.fields = {
         ...CORE_TAXON_FIELDS,
+        provisional: true,
```
pleary, line 377: There may be other UIs that similarly need to request this field to properly display provisional names (basically anything where a taxon name is displayed - explore, observation pages, projects, Identify, lifelists, etc)

pleary (changes requested):
I added some comments to get them out sooner. I'll continue to review a bit.

Didn't we also want to restrict our DwC-A exports at least to not include observations of provisional taxa? We could add an analogous [block like this one](https://github.com/inaturalist/inaturalist/blob/main/lib/darwin_core/archive.rb#L736) for excluding observations of provisional taxa. If we don't do that, we should probably at least add fields associated with taxa to indicate which taxa are provisional (as it stands they will appear as "valid" taxa with "valid" "scientific names" which might be confusing).

It would be great to have tests for all the logic changes. There are several new or modified model validations, and changed controller behavior. Those are pretty easily testable, and could form as good documentation for the cases they are meant to address

app/es_indices/taxon_index.rb
```diff
@@ -65,6 +65,7 @@ class Taxon < ApplicationRecord
         indexes :url, type: "keyword", index: false
       end
       indexes :extinct, type: "boolean"
+      indexes :provisional, type: "boolean"
```
pleary, line 68: New Elasticsearch mappings are not added automatically. Please coordinate release of this branch with other engineers to ensure consistent ES mappings

app/webpack/taxa/shared/ducks/taxon.js
loarie: The intent is to render the `[provisional]` decoration only on taxa/show - similar to how [extinct] is only rendered there as opposed to, e.g. https://www.inaturalist.org/observations?subview=table&taxon_id=21670
should I still make the field available to other UIs even if the current intent is not to use it there?

app/controllers/taxa_controller.rb
pleary: I'd still remove `&& params[:taxon][:provisional]`. It doesn't matter what the value is. If users don't have permission to change it, it should not be included. Same above with `photos_locked` (delete `&& params[:taxon][:photos_locked]`) since that was also modified in this PR.

Usually, or maybe always, sent values of `false` will turn into strings, and so `&& params[:taxon][:provisional]` will be true even if the value is "false", but that's a confusing way to go about it. I still think we should remove the extra conditional, but if we didn't want to, then I'd at least change it to `&& params[:taxon].key?( :provisional )` to check if the key exists which is really I think the intention of that extra conditional (though still unnecessary)

app/models/taxon.rb
```diff
@@ -1129,6 +1147,69 @@ def taxon_cant_be_its_own_ancestor
...
   end
 
+  def provisional_only_for_cortinariaceae
+    return unless provisional
+    # Only validate if provisional flag is being set/changed
+    return unless new_record? || will_save_change_to_provisional?
+
+    # Find the Cortinariaceae taxon (ID 48705)
+    cortinariaceae_taxon_id = 48_705
+    cortinariaceae = Taxon.find_by( id: cortinariaceae_taxon_id )
```
pleary: We avoid hard coding any IDs as they may change in different environments (e.g during tests). I'd change this to a dynamic lookup. Maybe something like `Taxon.where( name: "Cortinariaceae", iconic_taxon: Taxon::ICONIC_TAXA_BY_NAME["Fungi"] ).first`

app/models/taxon.rb
```diff
@@ -1129,6 +1147,69 @@ def taxon_cant_be_its_own_ancestor
...
+    return unless cortinariaceae # if Cortinariaceae doesn't exist, allow provisional (edge case)
```
pleary: Could we make it so if we hit this edge case, the validation fails instead? Could use the same error message as below. Otherwise someone could make a one-letter change to that taxon (which we now allow) and the would make provisional taxa allowed anywhere

app/models/taxon.rb
```diff
@@ -1129,6 +1147,69 @@ def taxon_cant_be_its_own_ancestor
...
+
+    # Check if this taxon descends from Cortinariaceae
+    # For new records, build the ancestry chain from parent
+    is_cortinariaceae_descendant = if new_record? && parent
+      parent.ancestor_ids.include?( cortinariaceae_taxon_id ) ||
```
pleary: If using my suggested lookup, then `cortinariaceae_taxon_id` won't be defined and the few instances of it in this method need to be updated to `cortinariaceae.id`. In either case that, that seems like a preferred way of referencing that taxon's ID

app/models/taxon.rb
```diff
@@ -1129,6 +1147,69 @@ def taxon_cant_be_its_own_ancestor
...
+  end
+
+  def provisional_name_format
+    return unless provisional
+    return if name.blank?
+
+    # Format: Genus sp. 'epithet-phrase'
+    # Example: Aureonarius sp. 'callisteus-infucatus' or Calonarius sp. 'AK01'
+    # Genus can contain letters or hyphens, epithet can contain letters (upper/lower), numbers, or hyphens
+    provisional_format = /\A[A-Z][a-z-]+\s+sp\.\s+'[A-Za-z0-9-]+'\z/
```
pleary: perhaps this should be moved to TaxonName::PROVISIONAL_NAME_FORMAT to match TaxonName::SCIENTIFIC_NAME_FORMAT. This regex is used at least twice in different models, which makes me think there should be a singular definition of it that gets reused, and future modifications should just be made once to that singular definition

app/models/taxon.rb
```diff
@@ -772,6 +786,14 @@ def capitalize_name
   def self.capitalize_scientific_name( name, rank )
     return name.capitalize if rank.blank?
 
+    # Check if this is a provisional name format (Genus sp. 'epithet')
+    # Preserve case within single quotes for provisional taxa
+    if name =~ /\A([A-Za-z][a-z-]*)\s+sp\.\s+'([A-Za-z0-9-]+)'\z/i
```
pleary, line 791: This potentially could also use a `TaxonName::PROVISIONAL_NAME_FORMAT`. If it is formatted as `PROVISIONAL_NAME_FORMAT = /\A([A-Z][a-z-]+)\s+sp\.\s+'([A-Za-z0-9-]+)'\z/` in TaxonName...

[note the inclusion of the parens for capturing components, the swapping of `[a-z-]*` here with `[a-z-]+` used in the other validations, and the swapping of `\A([A-Za-z]` here with `\A([A-Z]` used in the other validations)]

...then this could be written as `if name =~ Regexp.new( TaxonName::PROVISIONAL_NAME_FORMAT.source, Regexp::IGNORECASE )`. That will allow us to reuse the same singularly defined provisional name format, but extend it to ignore case as is desired here, and to capture the name parts needed here

app/models/taxon.rb
```diff
@@ -732,6 +743,9 @@ def set_rank_level
   end
 
   def remove_rank_from_name
+    # Skip rank removal for provisional taxa as they need "sp." in their name
+    return true if provisional
```
pleary: FYI, before Rails 5, returning false in a before_ callback would stop the callback chain. But since rails 5, the return values from validation callbacks don't matter and the only way to stop the chain is to raise in error. Essentially, this can just be `return if provisional` and there is no need to return `true` at the end. Not a big deal as the return value is ignored. I just know we have a lot of places in this codebase where validations return a particular boolean, but that's just for legacy reason and wanted to point out that its an old pattern that does not apply to the version of Rails we are currently using

app/models/taxon.rb
```diff
@@ -156,8 +164,11 @@ def set_uuid
   validate :cannot_edit_parent_during_content_freeze
   validate :restrict_name_changes_by_rank,
     on: :update,
-    # bypass name restrictions only used in specs
-    unless: -> { validation_context == :bypass_name_restrictions }
+    # bypass name restrictions only used in specs or for provisional taxa
+    # (including when transitioning from provisional to non-provisional)
+    unless: lambda {
+      validation_context == :bypass_name_restrictions || provisional || will_save_change_to_provisional?
```
pleary: This will allow radical changes to the provisional epithet of provisional names. Might it be better to modify `restrict_name_changes_by_rank` to consider provisional names, and still allow minor variations to the provisional epithet only?

app/webpack/taxa/shared/ducks/taxon.js
pleary: I don't feel strongly about it. If you only want the `[Provisional]` decoration to show on taxa/show then ignore my comment. Just pointing out what we'd need to do to get it to show in other places

pleary (approved):

PR #4728, opened 2025-12-01
Web 859 add to sitestatistics some counts about identification

sylvain-morin:


app/assets/javascripts/stats.js
```diff
@@ -47,11 +47,42 @@ Stats.loadChartsFromJSON = function ( json ) {
   Stats.loadRecentUsers( json );
   Stats.loadDailyUsers( json );
   Stats.loadObservations( json );
+  Stats.loadIdentifications( json );
+  Stats.loadCumulativeIdentifications( json );
+  Stats.loadObservationsUnkown( json );
```
pleary: There's a few typos of `Unkown` for `Unknown`

pleary (approved):
This all looks great, and I think we can merge and deploy anytime.

It does seem like the counting process in `generate_old_stats_unknown_obs.rb` generates a smaller count than the approach in SiteStatistics. That could be due to some inconsistencies in data in the IDs index. Based on record counts, there appears to be some vestigial documents in the IDs index. I am running some `Identification.elastic_sync` processes to make sure at least all IDs are indexed, and there are no docs in the IDs index associated with since deleted IDs.

When counting ID'd observations via the ID index, if we are also counting records that no longer exist, that could inflate the `observations_with_ids`, and lower the value of `created_30_days_not_identified`, contributing to the count discrepancy. Hopefully these sync processes finish overnight

pleary:
One other reason for differences in counts could be due to the approximate nature of the [cardinality aggregation](https://www.elastic.co/docs/reference/aggregations/search-aggregations-metrics-cardinality-aggregation#_counts_are_approximate). When counting queries that match hundreds of millions of records, a 1% error could mean being off by millions. We could add `precision_threshold: 40000` (that's the highest supported value) to any cardinality aggregations to help improve their precision

PR #4745, opened 2025-12-10
feat: add banner on Identify linking to identification tutorial, WEB-845

pleary:


spec/controllers/observation_controller_spec.rb
```diff
@@ -774,5 +774,13 @@
       get :identify
       expect( user.privileged_with?( UserPrivilege::INTERACTION ) ).to be true
     end
+
+    it "sets user hide webinar banner preference if they have 100 or more improving IDs" do
+      expect( Identification ).to receive( :elastic_search ).
+        and_return( WillPaginate::Collection.new( 1, 30, 100 ) )
+      expect( user.prefers_hide_identify_webinar_banner ).to be_nil
+      user.set_webinar_banner_default_preference
```
kfc35: Do you mean for this to be a call to `get :identify`?

pleary: yes I did! Thanks for catching that. I've now updated the spec

PR #4765, opened 2026-01-15
Display warning if user is suspended

zygoballus:
Per request at https://forum.inaturalist.org/t/observations-by-suspended-users-should-indicate-the-user-is-suspended/67360. This code has not been tested and relies heavily on Cunningham’s Law.

### DIFF
--- a/app/webpack/observations/show/components/user_with_icon.jsx
+++ b/app/webpack/observations/show/components/user_with_icon.jsx
@@ -45,6 +45,12 @@ const UserWithIcon = ( {
         { !hideSubtitle && (
           <div className="subtitle">{ subtitleLink }</div>
         ) }
+        { user.suspended && (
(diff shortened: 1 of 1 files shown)


kfc35:
@zygoballus if you’re curious about how i18n works and want to add more to your PR: 

You simply have to add your text to `en.yml` In this case for example, I’d make two strings:
```
user_suspended: This user has been suspended
…
user_suspended_contact_us: If this is your account and you think this was done by mistake, please <a href=\"mailto:help@inaturalist.org\">contact us</a>.
```

And then you would reference those strings like you suggested: `{ I18n.t( “user_suspended” ) }`

You can view #4682 as a minimal example of what to do when you add and reference a new string that should be i18n’d (although in this case you’d use `I18n`, not `inatreact`)

PR #4766, opened 2026-01-16
 Fix community taxon display counting irrelevant ancestor disagreements

loarie:
When displaying "Cumulative IDs: X of Y" for the community taxon, ancestor IDs with disagreement=true were incorrectly counted as votes against, even when the disagreement was with a taxon in a completely different branch.

  For example, on [observation 192292496](https://www.inaturalist.org/observations/192292496), the display showed "Cumulative IDs: 3 of 5" for Cyclocosmiinae, but the algorithm summary correctly shows the denominator as 4 (3+1+0). The extra count came from a Mygalomorphae ID that disagreed with Dysdera crocata (a different branch) being incorrectly counted as a vote against Cyclocosmiinae.

  The fix: Only count an ancestor disagreement as a vote against if the ID's taxon is an ancestor of the previous_observation_taxon (i.e., the disagreement is "up the tree" within the same branch, not a branch switch).

https://linear.app/inaturalist/issue/WEB-898/community-taxon-display-shows-incorrect-denominator-for-cumulative-ids


pleary (approved):
Looks good to me

vizagite:
@pleary it didnt fix the bug fully when original disagreed with ID is lost:

https://www.inaturalist.org/observations/36915377
https://www.inaturalist.org/observations/22443605

PR #4770, opened 2026-01-20
Web 861 segment active users each month by new retained and re

sylvain-morin:


app/controllers/stats_controller.rb
```diff
@@ -441,17 +491,200 @@ def user_segments
...
+        reengaged: data[:reengaged]
+      }
+    end
+    @segment_active_users_max = @segment_active_users_rows.map {| row | row[:active].to_i }.max.to_i
   end
 
   private
 
+  # Fetch a cohort stat record by date (YYYY-MM-DD), defaulting to most recent.
+  def cohort_stat_record_for( stat_type, stat_date: nil )
```
pleary: To keep controllers focused on action response and rendering, and models focused on data fetching, some of these methods might be better placed in their respective models as static methods. This and `cohort_stat_neighbors` could be methods in `CohortStatistic`, `site_stat_record_for` in `SiteStatistic`,   `segmentation_stat_record_for` in `SegmentationStatistic`. `record_for_date` and `stat_record_neighbors` appear to work on multiple model instances, so unless there's a clean way to share them in models, through a concern or other module, leaving them in the controller is OK.

app/controllers/stats_controller.rb
```diff
@@ -366,6 +380,42 @@ def user_segments
...
+        render layout: "bootstrap"
+      end
+    end
+  end
+
+  def daily_active_user_model
+    stats_record, stats_records = site_stat_record_for(
+      stat_date: params[:date]
+    )
+    redirect_to "/" and return unless stats_record
```
pleary: Pretty minor - we're inconsistent with this but this can also be `redirect_to( root_url ) and return`. Seems like there's slightly more use of `root_url` then `"/"` in controller redirects

app/models/cohort_statistic.rb
```diff
@@ -0,0 +1,429 @@
...
+  def self.generate_acquisition_cohorts_stats( at_time = Time.now, options = {} )
+    at_time = at_time.utc.end_of_day
+    scope = CohortStatistic.where( "DATE(created_at) = DATE(?)", at_time ).
+      where( stat_type: "acquisition" )
+    if options[:force]
+      scope.delete_all
+    elsif scope.exists?
+      return
+    end
+    sleep 1
```
pleary: What are the `sleep`s doing in this model? Would be best to avoid them if possible

app/views/stats/acquisition_cohort_statistics.html.haml
```diff
@@ -0,0 +1,126 @@
+- content_for( :title ) do
+  = @title = "Acquisition Cohort Statistics"
+- content_for( :extracss ) do
+  = stylesheet_link_tag :stats
+  :css
```
pleary: This inline CSS appears to be identical to that in `behavior_cohort_statistics.html.haml‎`. I prefer to avoid most inline CSS. Can this be extracted to a stylesheet that is included here and in `behavior_cohort_statistics.html.haml‎`?

app/views/stats/daily_active_user_model.html.haml
```diff
@@ -1,12 +1,29 @@
 - content_for(:title) do
   = @title = "Daily Active User Model"
-.container
+- content_for(:extracss) do
+  = stylesheet_link_tag :stats
+  :css
```
pleary: Empty `:css` block

app/views/stats/segmentation_dau_mau.html.haml
```diff
@@ -0,0 +1,169 @@
...
+    .tooltip {
+        position: absolute;
+        padding: 10px;
+        background-color: #fff;
+        border: 1px solid #ddd;
+        border-radius: 5px;
+        pointer-events: none;
+        opacity: 0;
+    }
+%script{ src: "https://d3js.org/d3.v4.js" }
```
pleary, line 16: We have a local version of `d3.min` JS we reference, for example in the [stats index view](https://github.com/inaturalist/inaturalist/blob/main/app/views/stats/index.html.haml#L16). Looks like it might possibly be version `3.4.11`. If that would work, might be nice to use that instead

app/models/cohort_statistic.rb
sylvain-morin: This sleep has been in SiteStatitics for 6 years... https://github.com/inaturalist/inaturalist/blob/main/app/models/site_statistic.rb#L17

Do you think I should remove both?

pleary: I can't think of a reason any of them are needed. I say let's remove them all

app/views/stats/segmentation_dau_mau.html.haml
sylvain-morin: It would require some changes as we are using some v4 methods in these state pages (this PR just introduced a split of the stats page, no big change or new d3.js introduction).

I would recommend to rather work on a migration, as v3 or v4 are +10 years old:
https://linear.app/inaturalist/issue/WEB-908/migrate-d3js-to-a-most-recent-unique-version

PR #4775, opened 2026-01-27
 Enable provisional taxa for curators with Cortinariaceae restriction

loarie:
- Allow curators (not just admins) to create/edit provisional taxa
- Hide provisional checkbox when parent is not Cortinariaceae descendant
- Allow spaces in provisional name epithets
- Fix URL generation for provisional taxa special characters
- Remove orphaned "Note" label from audit comment field
[Linear Ticket](https://linear.app/inaturalist/issue/WEB-910/enable-provisional-taxa-for-curators)

### DIFF
--- a/spec/controllers/taxa_controller_spec.rb
+++ b/spec/controllers/taxa_controller_spec.rb
@@ -470,10 +470,17 @@ describe TaxaController do
   end
 
   describe "provisional" do
-    before { load_test_taxa( iconic: true ) }
+    before do
+      load_test_taxa( iconic: true )
+      Taxon.instance_variable_set( :@cortinariaceae, nil )
+    end
     let( :fungi ) { Taxon::ICONIC_TAXA_BY_NAME["Fungi"] }
-    let( :cortinariaceae ) { Taxon.make!( name: "Cortinariaceae", rank: Taxon::FAMILY, parent: fungi ) }
-    let( :cortinarius ) { Taxon.make!( name: "Cortinarius", rank: Taxon::GENUS, parent: cortinariaceae ) }
+    let( :cortinariaceae ) do
+      Taxon.make!( name: "Cortinariaceae", rank: Taxon::FAMILY, parent: fungi, iconic_taxon: fungi )
+    end
+    let( :cortinarius ) do
+      Taxon.make!( name: "Cortinarius", rank: Taxon::GENUS, parent: cortinariaceae, iconic_taxon: fungi )
+    end
 
     it "should be updateable by curators" do
       t = Taxon.make!( name: "Cortinarius testus", rank: Taxon::SPECIES, parent: cortinarius, provisional: false )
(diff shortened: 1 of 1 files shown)


app/models/taxon.rb
```diff
@@ -865,7 +869,7 @@ def to_s
   def to_param
     return nil if new_record?
 
-    "#{id}-#{name.gsub( /\W/, '-' )}"
+    "#{id}-#{name.gsub( /[^a-zA-Z0-9]/, '-' )}"
```
alexcameron89, line 872: 👋 I'm just lurking and had a question here: do you want the `gsub` to replace each matching character with a `-` or should sequentially matched characters be replaced by one `-`?

For example:
```ruby
"hello  there".gsub( /[^a-zA-Z0-9]/, '-' ) # w/out + matcher 
=> "hello--there"
"hello  there".gsub( /[^a-zA-Z0-9]+/, '-' ) # with + matcher 
=> "hello-there"
```

```suggestion
    "#{id}-#{name.gsub( /[^a-zA-Z0-9]+/, '-' )}"
```

loarie: Thats how we handle rendering other URLs on the site, e.g. https://www.inaturalist.org/taxa/331124-Citrus---limon for Citrus × limon (vs `331124-Citrus-limon `)

My goal was just to replicate that behavior for special characters in provisional names and not rethink how we're handling constructing URLs. But were that the goal, I'm not sure which would be a better approach.

Specifically, my goal was to replace broken URLs appearing in some places, e.g. the [paperclip icon here](https://www.inaturalist.org/taxa/1663040) which renders as `https://www.inaturalist.org/taxa/1663040-Cortinarius-sp.-'NY02'`
Screenshot 2026-01-27 at 3 40 40 PM
by matching other URL's appearing in other places that are working properly, e.g. the links in [the taxonomic tree here](https://www.inaturalist.org/taxa/48704#taxonomy-tab) which render already renders properly `https://www.inaturalist.org/taxa/1663040-Cortinarius-sp---NY02-`
Screenshot 2026-01-27 at 3 42 11 PM

alexcameron89: Ah perfect, that makes sense. Thank you for the explanation!

pleary: Pretty minor comment related to this. The regex `\W` is the inverse of `\w` which is basically `[a-zA-Z0-9_]` - note the inclusion of `_`. We only have 25 taxa with preferred names that include an underscore, and it doesn't seem like any one of them really need it.

As an example of the minor difference this makes, for taxon [1378334](https://www.inaturalist.org/taxa/1378334-Ctenucha-_affinis_) `Ctenucha _affinis_`, the original to_param would generate `1523161-Ctenucha-_affinis_` and this method generates `1523161-Ctenucha--affinis-`. If anything this version looks a little better.

I don't think we need to change anything, just wanted to point one this small difference

PR #4784, opened 2026-02-05
Ops 282 create production like test users on staging for mobile

sylvain-morin:


app/controllers/admin_controller.rb
```diff
@@ -87,6 +87,90 @@ def deleted_users
...
+      count: count
+    )
+    flash[result.success ? :notice : :error] = result.message
+    redirect_to build_test_users_admin_path
+  end
+
+  private
+
+  def staging_environment?
+    request.host.to_s.include?( "staging" )
```
pleary: I wonder if this might be better handled by a config value we set in the Rails config file rather than checking if this host name contains this substring. That way we can be extra sure that an instance will only allow these features if we've explicitly configured the instance to allow them. They can also then be toggled on in potentially a test or development environment.

app/controllers/admin_controller.rb
```diff
@@ -87,6 +87,90 @@ def deleted_users
...
+      login: params[:login],
+      observations_count: params[:observations_count],
+      identifications_for_others_count: params[:identifications_for_others_count]
+    }
+    render layout: "admin"
+  end
+
+  def build_test_user
+    unless staging_environment?
+      flash[:error] = "Test user creation is only allowed on staging"
```
pleary: If we do update the requirement check to use a config value instead of checking the request host name, maybe this can be made more general, e.g. "Test user creation is not allowed". I think I'd feel more comfortable if the concept of `staging` wasn't involved at all, and we're simply checking if the feature has been enabled via configuration or not

app/views/admin/build_test_users.html.haml
```diff
@@ -0,0 +1,105 @@
+%h2 Build Test Users
+
+%p Create a confirmed user and attach existing content for testing.
+- unless @staging_enabled
+  .alert.alert-warning
+    This tool is disabled because this is not a staging environment.
```
pleary: To be repetitive - same comments as above. I'd personally prefer not mentioning `staging` and instead focus on wether the feature has been enabled via configuration

lib/build_test_user.rb
```diff
@@ -0,0 +1,457 @@
...
+  RANDOM_SAMPLE_THRESHOLD = 1_000_000
+  # Use a floor to avoid very old / sparse ID ranges.
+  MIN_OBSERVATION_ID = 100_000
+  MAX_OBSERVATION_ID = Observation.maximum( :id ).to_i
+  MIN_IDENTIFICATION_ID = 100_000
+  MAX_IDENTIFICATION_ID = Identification.maximum( :id ).to_i
+
+  Result = Struct.new( :success, :message, :user, :details, keyword_init: true )
+
+  def self.build_user( *args, login: nil, password: nil, observations_count: nil,
```
pleary, line 21: If we have an application configuration setting on wether or not test users are allowed to be created, it would be great to check it here and not allow this method to be run on instances that to not have it enabled. That way we can better prevent any way to run the underlying creation logic directly in environments we don't want it to run. Right now the only checks appear to be happening at the web request level, but not at this underlying library level.

lib/build_test_user.rb
```diff
@@ -0,0 +1,457 @@
...
+        identifications_from_own_obs: identifications_from_own_obs,
+        indentifications: indentifications,
+        observations_from_other_ids: observations_from_other_ids,
+        failed_observations: failed_observations,
+        failed_identifications: failed_identifications
+      }
+    )
+  end
+
+  def self.apply_updates( target_user_id:, update_action:, count: )
```
pleary, line 249: Given this library uses static methods instead of working off of an instantiated instance, I'd add the configuration check to any method that might modify data. For example we also don't want to allow anyone to be able to run this method in a command line on an instance that has not been configured to allow test user creation

lib/build_test_user.rb
```diff
@@ -0,0 +1,457 @@
...
+      attempts += 1
+      puts(
+        "BuildTestUser.randomize_identifications attempt=#{attempts} " \
+          "candidates=#{candidates.size} found=#{found.size} total=#{results.size}"
+      )
+    end
+    results.take( count )
+  end
+
+  def self.ensure_notifier_users
```
pleary, line 430: Another method that needs a configuration check so it cannot run on instances where the feature is not enabled

pleary (changes requested):
Aside from the suggestions I made about how we might enable this feature through the application config.yml, and to prevent any part of the creation logic to run if that config value is not set, I think the rest of it seems OK to me

app/controllers/admin_controller.rb
sylvain-morin: implemented the configuration parameter to enable this feature

app/controllers/admin_controller.rb
sylvain-morin: implemented the configuration parameter to enable this feature

lib/build_test_user.rb
sylvain-morin: implemented the configuration parameter to enable this feature

PR #4788, opened 2026-02-10
feat: collection projects can hide leaderboard, WEB-913

pleary:


app/assets/stylesheets/projects/show2.scss
```diff
@@ -698,7 +698,10 @@ a {
       display: flex;
     }
   }
-  margin-bottom: 20px;
+}
+
+.info-grid {
+  margin-top: 20px;
```
caseyhofford, line 704: This adds some potentially unwanted extra margin to the `InsufficientRequirements` and `BeforeEvent` components. We might want to isolate this CSS to just the OverviewTab if we care about this extra padding.

Image
Image

This is a larger topic not for this PR, but in the past I've used CSS modules or `styled-components` as ways to colocate CSS with the components using it.

pleary: Thanks - I made a change to address that

PR #4789, opened 2026-02-10
feat: intraday collection projects display times, WEB-911

pleary:


app/webpack/projects/show/components/app.jsx
```diff
@@ -141,7 +141,17 @@ const App = ( {
...
     const start = project.startDate.format( "MMM D, YYYY" );
     const end = project.endDate.format( "MMM D, YYYY" );
-    eventDates = I18n.t( "date_to_date", { d1: start, d2: end } );
+    if ( start === end ) {
+      if ( project.startDateIncludesTime && project.endDateIncludesTime ) {
+        const startTime = project.startDate.format( I18n.t( "momentjs.time_with_zone" ) );
+        const endTime = project.endDate.format( I18n.t( "momentjs.time_with_zone" ) );
+        eventDates = I18n.t( "date_time_to_time", { date: start, start_time: startTime, end_time: endTime } );
+      } else {
+        eventDates = start;
```
kfc35: This works in the case that both `startDate` and `endDate` do not include time, but can the user specify times for either `startDate` or `endDate` without expressing both? (i.e. a delayed start time that goes to end of day or an earlier end time than end of day)

In which case there would need to be some `if else` blocks for just `project.startDateIncludesTime` and `project.endDateIncludesTime` and handled accordingly

config/locales/en.yml
```diff
@@ -1797,6 +1797,10 @@ en:
   # d2" if "to" might be translated differently for different values of d1 and
   # d2
   date_to_date: "%{d1} - %{d2}"
+  # Express a single date with a time range. Better to use a flexabile form
```
kfc35:
```suggestion
  # Express a single date with a time range. Better to use a flexible form
```

app/webpack/projects/show/components/app.jsx
```diff
@@ -139,9 +139,19 @@ const App = ( {
   if ( project.rule_observed_on && project.startDate ) {
     eventDates = project.startDate.format( "MMM D, YYYY" );
   } else if ( project.rule_d1 && project.rule_d2 && project.startDate && project.endDate ) {
-    const start = project.startDate.format( "MMM D, YYYY" );
-    const end = project.endDate.format( "MMM D, YYYY" );
-    eventDates = I18n.t( "date_to_date", { d1: start, d2: end } );
+    const startDay = project.startDate.format( "MMM D, YYYY" );
+    const endDay = project.endDate.format( "MMM D, YYYY" );
+    if ( startDay === endDay ) {
```
caseyhofford, line 144: Currently the timezone information is being lost in momentDateFromString: https://github.com/inaturalist/inaturalist/blob/315728e674d10be2c6edf268a8d80c03420e3e2d/app/webpack/projects/shared/util.js#L6 which will cause projects to be treated as intraday only if they are in the same day in UTC. In some quick tests removing `.parseZone( ).tz( TIMEZONE )` from https://github.com/inaturalist/inaturalist/blob/0c3d5abdb86d02fb6f63d06dbdc797a8911b67f5/app/webpack/projects/shared/util.js#L11-L12 appears to resolve this, but may have unintended side effects.

Before (set project time with tz: -7:00):

Image

Image

After:

Image

config/locales/en.yml
```diff
@@ -3362,6 +3366,8 @@ en:
     # See https://momentjs.com/docs/#/displaying/format/ for formatting codes
     time_hours: "h:mm a"
     # See https://momentjs.com/docs/#/displaying/format/ for formatting codes
+    time_with_zone: "LT z"
```
caseyhofford, line 3398: With the change suggested above, we need to change to numerical timezone offsets, or explicitly set the timezone name on the moment datetime object.
```suggestion
    time_with_zone: "LT Z"
```

caseyhofford (approved):
Approving to unblock merging if needed pre code-freeze. May still need follow up to ensure timezones are being handled.

app/webpack/projects/show/components/app.jsx
pleary: Was your local browser using UTC as the local time? In your screenshot, the time formatting used in the "Project Requirements" section should be localizing the time for your local timezone. That's showing UTC, which tells me the page is assuming you're in UTC. For the time+offsets you put in the project date editor, those times are in different days UTC time, so I would expect the banner to show multiple days. If your local browser was set to PDT, with those same start and end dates, I'd expect it to show the single (local) day. That's how it appears to be working for me

caseyhofford: Odd, my browser is using local time but seems to be setting the time range to UTC. This shouldn't matter, but what browser are you using? I am using Brave.

PR #4863, opened 2026-04-09
fix: Identify annotating with keyboard; admin user_content page

pleary:


spec/controllers/admin_controller_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+
+require "spec_helper"
+
+describe AdminController do
+  let( :admin_user ) { make_admin }
+
+  describe "user_content" do
+    render_views
+
+    it "should show an error about account age if user signed up in the last 60 days" do
```
kfc35:
```suggestion
    it "should load correctly for the requested user id" do
```

PR #4864, opened 2026-04-10
web 974 map reason to suspended_until value

caseyhofford:
- **wip form**
- **Add suspend_until default for suspend form**
- **form updates**
- **Add suspend_until default for suspend form**
- **Gate timed suspension form to only admins**
- **Add form validations**
- **move suspension timeframe validation to backend**
- **move js out of haml**
- **Set datetime selector minimum on server**
- **Set local-time datetime selector minimum in browser for TZ support**
- **Add moderator action validation specs**
- **add notes to all new translation keys.**
- **feat:web-967: display moderation history details on user admin page**
- **feat:web-971: make suspend ModeratorAction editable**
- **fix: haml bug**
- **allow editing from user page**
- **Display creator and editors on suspensions**
- **Gate timed suspension edit form to admins only**
- **fix: do not prevent suspended_until param for non admins**
- **Add edit suspension translations**
- **share suspension form with edit**
- **Allow admin to suspend curator from Users::show**
- **Add missing translation string comment**
- **feat: web-974: suspension form has reason dropdown**
- **Store translation keys**
- **translate reason spec**
- **Add descriptor text to translations**
- **Update reasons with timeframes**
- **Move suspension reason note**
- **Add suspension_reason_edit_note**
- **Display both creator and editor on moderation_history**
- **Suspension form cleanups**


caseyhofford:
This work was moved into https://github.com/inaturalist/inaturalist/pull/4856

PR #4865, opened 2026-04-10
feat: Add new homepage, WEB-950

sundisc:
Adds a redesigned homepage (/welcome_v2) for preview and testing. Eventually will add an authorization action to the controller to allow admins only, but leaving it out for now in order to test the logged out header design.

  - New WelcomeV2Controller and welcome_v2/index view
  - Two swipeable carousels (testimonials, explore observations) with dot navigation, touch handling, and static fallback   
  when content fits without scrolling
  - Script for generating content from observations, `fetch_observations.rb`
  - FAQ with CSS open/close animation
  - iOS-only sticky app download banner
  - Locale-aware typography using locally-hosted Libre Baskerville, Lato, and Noto CJK subsets
      - Noto subset font generator `subset_noto.rb` to minimize non-latin font file sizes
  - Announcements, Google Webmaster verification, and store URL override wired up from the existing welcome page patterns


config/homepage_observations.json
```diff
@@ -0,0 +1,652 @@
+{
```
pleary, line 1: This file shouldn't live within `app/assets/images`. We generally have things like this exist at a `Site` model instance level, e.g site.stylesheet, or site.custom_footer, etc. It can be good to have different values for different environments, like development or test, so we can control the contents as needed for the given environment and its data. For now, I think moving it to Site might be a bigger move than we want to make at this stage - that's something we can leave for future refactoring. What I might recommend for now is this file goes to `/config/homepage_observations.json` or something like that. I'll comment in the controller a possible improved way to load it

app/assets/fonts/noto/noto-serif-arabic-400.woff2
pleary, line 1: Since all the `noto` fonts should change a lot, let's not have them live in `public/fonts`. They should rather be part of the rails asset pipeline, and live at `app/assets/fonts/noto`. We'll need to modify `app/assets/config/manifest.js` to make sure we read from nested font directories:
```diff
-//= link_directory ../fonts
+//= link_tree ../fonts
```
And we'll need to update the stylesheet to load them properly. I'll add a comment in the stylesheet

app/assets/stylesheets/welcome_v2_fonts.scss
```diff
@@ -0,0 +1,207 @@
...
+//
+// CJK fonts (SC, TC) are NOT in the "Noto Serif" unicode-range stack because
+// their U+4E00-9FFF range overlaps JP/KR, causing browsers to download all of
+// them on any page with kanji. Instead they are separate named families applied
+// via :lang() selectors in welcome_v2.scss.
+
+// Simplified Chinese (zh-CN)
+@font-face {
+  font-family: "Noto Serif SC";
+  src: url("/fonts/noto/noto-serif-sc-400.woff2") format("woff2");
```
pleary: When we move the font files into `public/assets/fonts` we'll need to update how they are referenced here and subsequent lines: 
```diff
-  src: url("/fonts/noto/noto-serif-sc-400.woff2") format("woff2");
+  src: font-url("noto/noto-serif-sc-400.woff2") format("woff2");
```

app/controllers/welcome_v2_controller.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+class WelcomeV2Controller < ApplicationController
+  unless defined?( OBSERVATIONS_DATA )
+    OBSERVATIONS_DATA = JSON.parse(
+      File.read( Rails.root.join( "app/assets/images/welcome_v2/observations/observations.json" ) )
+    ).freeze
+  end
+
+  STEPS_DATA = [
```
pleary: I think this might be better as a plain ruby variable in the view for now, and not a controller constant. e.g.  add this in the view, either at the top, or just before it is used (I was getting an error defining a constant in a `:ruby` block, but defining it as a plain variable worked):
```
:ruby
  steps_data = [
    [:observe,      "step1_body", "camera"],
    [:identify_cta, "step2_body", "identify"],
    [:contribute,   "step3_body", "globe"]
  ].freeze
```

app/controllers/welcome_v2_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    [:observe,      "step1_body", "camera"],
+    [:identify_cta, "step2_body", "identify"],
+    [:contribute,   "step3_body", "globe"]
+  ].freeze
+
+  TESTIMONIALS_DATA = {
```
pleary: Similar to steps, I think this would be better as a ruby variable in the view. I'd also recommend changing `people_path` to `login` with just their login string, and allow the Rails `people_path` helper generate the path dynamically in the view

app/controllers/welcome_v2_controller.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+class WelcomeV2Controller < ApplicationController
+  unless defined?( OBSERVATIONS_DATA )
```
pleary: For loading observations data - I'm not sure we should be making `File.read` calls in a controller constant. How about for now making a new static method on site, e.g. Site.homepage_observation_data, and leveraging Rails.cache to store the data centrally for all apps to fetch as needed. That could look something like:
```ruby
  def self.homepage_observation_data( options = {} )
    if options[:refresh]
      Rails.cache.delete( "sites_homepage_observation_data" )
    end
    Rails.cache.fetch( "sites_homepage_observation_data" ) do
      JSON.parse(
        File.read( Rails.root.join( "config/homepage_observations.json" ) )
      ).freeze
    end
  end
```
Then in this controller replace all `OBSERVATIONS_DATA` with `Site.homepage_observation_data`. As I write in anther comment, we eventually might want to abstract this out further so different sites can have different data, and different environments can have different data. But given the desired release timeframe, that might wait for future refactoring.

app/controllers/welcome_v2_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  TESTIMONIALS_DATA = {
+    gube: { name: "Dzivula Gube", people_path: "/people/dzivulajr_03" },
+    kratzer: { name: "Chris Alice \"Alie\" Kratzer", people_path: "/people/humanbyweight" },
+    sumanapala: { name: "Amila Sumanapala", people_path: "/people/amila_sumanapala" },
+    fachetti: { name: "Nágyla Fachetti Coser", people_path: "/people/nah_fachetti19" },
+    velazco: { name: "Carlos G. Velazco-Macias", people_path: "/people/aztekium" }
+  }.freeze
+
+  STORY_DATA = [
```
pleary: In the same vein as reducing the responsibility of the controller to actions, can we maybe move this data into a Site constant for now, e.g. `HOMEPAGE_STORY_DATA = [ ... ]` in the Site model? Also, is there a reason why are we using `tr.ee` shortened URLs here? If not, I might just use the real URLs and not obfuscate them so the code is more understandable/searchable

app/controllers/welcome_v2_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    @header_donate_url = "#{donate_path}?redirect=true&segment=gmHome&#{HEADER_UTM}&utm_medium=webmobile&utm_content=header_donate_cta"
+    @header_login_url  = "#{login_path}?#{HEADER_UTM}&utm_medium=owned_web&utm_content=header_login_cta"
+    @header_signup_url = "#{signup_path}?#{HEADER_UTM}&utm_medium=owned_web&utm_content=header_signup_cta"
+
+    @locale_key = I18n.locale.to_s
+    @locale_base = @locale_key.split( "-" ).first
+
+    @announcements = Announcement.active_in_placement( Announcement::WELCOME_INDEX, site: @site )
+    @google_webmaster_verification = @site.google_webmaster_verification if @site
+    @steps_data = STEPS_DATA
```
pleary: If we put `STEPS_DATA` and `TESTIMONIALS_DATA` into the view, these local variable definitions can go away

app/controllers/welcome_v2_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+    @testimonials_data = TESTIMONIALS_DATA
+    @story_data = build_story_data
+    @sample_observation_data = build_sample_observation
+    @explore_observations_data = OBSERVATIONS_DATA["explore"] || []
```
pleary: If we move OBSERVATIONS_DATA into a Site static method, this could be updated to `Site.homepage_observation_data["explore"] || []`

app/controllers/welcome_v2_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+  ].freeze
+
+  HEADER_UTM = "utm_source=inaturalist_homepage&utm_campaign=homepage_redesign_2026".freeze
+
+  def index
+    @responsive = true
+    @skip_external_connections = true
+    @skip_react = true
+
+    @store_url = "https://tr.ee/MymHMn"
```
pleary: Any reason to use this over just putting `https://store.inaturalist.org` in the view like we were previously doing?

app/views/welcome_v2/index.html.haml
```diff
@@ -0,0 +1,427 @@
...
+- content_for(:extrahead) do
+  - unless @google_webmaster_verification.blank?
+    - @google_webmaster_verification.split( ";" ).each do |verification|
+      %meta{content: verification.strip, name: "google-site-verification"}
+  %meta{content: "website", property: "og:type"}
+  %meta{content: @site&.name, property: "og:title"}
+  - if @site&.shareable_image?
+    %meta{content: image_url(@site.shareable_image.url), property: "og:image"}
+    %meta{content: image_url(@site.shareable_image.url), property: "twitter:image"}
+  %meta{content: root_url, property: "og:url"}
```
pleary, line 12: Thanks for adding back all the meta tags we currently have in the homepage!

app/views/welcome_v2/index.html.haml
```diff
@@ -0,0 +1,427 @@
...
+-# ── iOS app banner ───────────────────────────────────────────
+= link_to "#{redirect_link_path(31)}?#{@header_utm}&utm_medium=owned_web&utm_content=sticky_footer_cta", class: "wv2-app-banner", "aria-label": "Download iNaturalist on the App Store" do
+  = image_tag "welcome_v2/app-icon.webp", class: "wv2-app-banner__icon", alt: ""
+  .wv2-app-banner__text
+    = image_tag "welcome_v2/inat-wordmark.webp", class: "wv2-app-banner__wordmark", alt: "iNaturalist"
+    %p.wv2-app-banner__tagline Discover nature around you
+  %span.wv2-app-banner__cta Use app
+
+- content_for :extrajs do
+  :javascript
```
pleary: We tend to prefer extracting long custom JS like this into its own file in `app/assets/javascript` then including the file here, e.g. `javascript_include_tag 'homagepage_v2'`. That will allow the JS linter to work better for syntax, and keep presentation and JS separated

app/views/welcome_v2/index.html.haml
```diff
@@ -0,0 +1,427 @@
...
+  .wv2-section.wv2-section--light
+    %h2.text-center= t("#{s}.faq.heading")
+    .wv2-faq
+      - %w[upload_existing who_can_use research_grade no_knowledge need_app].each do |key|
+        %details.wv2-faq__item
+          %summary= t("#{s}.faq.#{key}_question")
+          %p= t("#{s}.faq.#{key}_answer")
+
+-# ── iOS app banner ───────────────────────────────────────────
+= link_to "#{redirect_link_path(31)}?#{@header_utm}&utm_medium=owned_web&utm_content=sticky_footer_cta", class: "wv2-app-banner", "aria-label": "Download iNaturalist on the App Store" do
```
pleary, line 268: This sticky footer is covering the language selector. Ideally we are still able to see the entire height of the footer, or at least the language selector is still usable one way or another

app/controllers/welcome_v2_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+
+  STORY_DATA = [
+    ["moth_heading",      "moth_body",        "pav_johnsson", "New Zealand",  "Frosted Phoenix", "frosted_phoenix", "https://tr.ee/FYNsVb"],
+    ["mantis_heading",    "mantis_body_html", "glendawalter", "Australia",    nil,               "inat",            "https://tr.ee/j29kgK"],
+    ["amphibian_heading", "amphibian_body",   "tcuriel",      "United States", nil,              "california_newt", "https://tr.ee/hQZIFH"]
+  ].freeze
+
+  HEADER_UTM = "utm_source=inaturalist_homepage&utm_campaign=homepage_redesign_2026".freeze
+
+  def index
```
pleary: We will eventually need to make this behave like the [current Welcome controller](https://github.com/inaturalist/inaturalist/blob/web-950-create-bare-bones-page-with-strings-present/app/controllers/welcome_controller.rb#L13) for logged-in users - sending them to `home_path`. We will also need to carve out a new conditional - only rendering the V2 welcome view when on the default site, and using the existing view for all other sites, since we are only changing the homepage for the default site

pleary (changes requested):
Adding some initial thoughts now. I'll continue to review other bits, but wanted to make sure some feedback was flowing.

Overall this is looking really good. I'm just commenting on some polish, recommending approaches consistent with how this codebase does similar things, and just small developer quality of life recommendations. Generally though, great stuff and not too much is needed to get is ready for release!

app/controllers/welcome_v2_controller.rb
sundisc: Good idea! https://github.com/inaturalist/inaturalist/pull/4865/commits/1937fc1158105b9de9cc7f788a5645796a571c11

config/homepage_observations.json
sundisc: Added here: https://github.com/inaturalist/inaturalist/pull/4865/commits/1937fc1158105b9de9cc7f788a5645796a571c11

app/assets/fonts/noto/noto-serif-arabic-400.woff2
sundisc: Updated here, but haven't tested yet because I need to set up a local prod environment: https://github.com/inaturalist/inaturalist/pull/4865/commits/553c39c00c527713c0e5cdc7d3c2307da66e8abb

app/assets/stylesheets/welcome_v2_fonts.scss
sundisc: https://github.com/inaturalist/inaturalist/pull/4865/commits/553c39c00c527713c0e5cdc7d3c2307da66e8abb

app/controllers/welcome_v2_controller.rb
sundisc: https://github.com/inaturalist/inaturalist/pull/4865/changes/685486716c7f132cc413be08f5376920d824bb5e

app/controllers/welcome_v2_controller.rb
sundisc: https://github.com/inaturalist/inaturalist/pull/4865/changes/685486716c7f132cc413be08f5376920d824bb5e

app/controllers/welcome_v2_controller.rb
sundisc: https://github.com/inaturalist/inaturalist/pull/4865/changes/685486716c7f132cc413be08f5376920d824bb5e but will switch to login seperately

sundisc: https://github.com/inaturalist/inaturalist/pull/4865/commits/347337c1cb812464a6e90aa7a2a4a7ab7ece147c

app/controllers/welcome_v2_controller.rb
sundisc: It enables tracking on external links. Seems like `https://store.inaturalist.org` just points to the threadless store, so we wouldn't be able to track. Looking forward we will eventually have Google Analytics event tracking.

app/views/welcome_v2/index.html.haml
sundisc: https://github.com/inaturalist/inaturalist/pull/4865/commits/78e5b6d09423932bf03feaf508c561a378a070c8

app/views/welcome_v2/index.html.haml
sundisc: That's really strange.. there's padding there to prevent that and it works for all of the sizes I tested and on my phone. What's your device + OS?

pleary: Ah sorry, this was on my end. I made some changes to test locally and on staging1 in a non-iOS web browser, but didn't make all the changes necessary (failed to add the bottom padding to #inatnotice). You can ignore this

app/assets/stylesheets/welcome_v2.scss
```diff
@@ -0,0 +1,1467 @@
...
+:lang(zh-TW) .wv2,
+:lang(zh-HK) .wv2      { --font-serif: "Noto Serif TC", Georgia, serif; }
+:lang(ja) .wv2,
+:lang(ja-phonetic) .wv2 { --font-serif: "Noto Serif JP", Georgia, serif; }
+:lang(ko)         .wv2  { --font-serif: "Noto Serif KR", Georgia, serif; }
+$font-sans:    "Lato", sans-serif;
+$font-mono:    "Courier Prime", "Courier New", monospace;
+
+$letter-spacing-tight: -0.03em;
+$breakpoint-mobile:    767px;
```
pleary: This is pretty minor pixel peeping, but there's a 2px browser width range from 786px to 787px that's a little funky - some things have transitioned to web styles like padding in the hero section, but fonts remain small. Also the links in the header pop in 1px later. The difference seems to be due to combination of max-width and min-width @media queries with the same px value.

I think it would be good to have min-width use a value 1px higher. I also may have messed this up in the header - that one may be using a media query 1 px too high. I'd say ideally the hero padding, hero font size, and header links pop-in all change at exactly the same browser width.

sundisc: Ah nice call.. hadn't spent enough time testing in that range. Here I move the breakpoint variables to their own file, rename them, and use them in the header as well as welcome_v2 styles https://github.com/inaturalist/inaturalist/pull/4865/commits/0f40d309f51360fa8373ff3ba3e3a3c61e2719d5

app/controllers/welcome_v2_controller.rb
sundisc: Ok this feels all set.
Bulk of the changes here: https://github.com/inaturalist/inaturalist/pull/4865/commits/9d3a9253929ae55a86c054e94b2c8967bb05d83f
Some meaningless refactors here: https://github.com/inaturalist/inaturalist/pull/4865/commits/0f40d309f51360fa8373ff3ba3e3a3c61e2719d5

pleary (approved):
Looks OK to me to merge!

PR #4866, opened 2026-04-10
fix(quick): automated unsuspend conditional fix

caseyhofford:
Quick fix for bug in ModerationAction index. Corrects conditional display for `Moderator/flagger` column to use elsif to fix displaying `deleted user` on `Flags`. Adds an additional check for `ModeratorAction::UNSUSPEND` to display the automated unsuspend text.

Testing Instruction:
1. Flag a user
2. Update flag `<flag>.update(user_id: 0)`
3. Load moderation history for user
4. Confirm that the Moderator/flagger column does not display `deleted user`

Before:
image

After:
image

### DIFF
--- a/app/views/users/moderation.html.haml
+++ b/app/views/users/moderation.html.haml
@@ -66,7 +66,7 @@
         %td
           - if record.user_id == 0 && record.is_a?( Flag )
             = @site.site_name_short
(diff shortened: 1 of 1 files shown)


pleary (approved):
Thanks!

PR #4869, opened 2026-04-13
web 991 replace local email header image

caseyhofford:
Updates email header again this time with full identity image (including logo) as requested in: (https://linear.app/inaturalist/issue/WEB-965/generate-emailsnotifications#comment-4800f9e7)


caseyhofford:
Discussed w/ Abhas and confirmed we want wordmark, not full identity image on email headers.

PR #4871, opened 2026-04-14
Store reasons for suspension edits with audit history

caseyhofford:
- feat: store suspension edits as audit
- unsuspend page includes a note warning that users will see unsuspension reasons
-   require `audit_comment` when updating a ModerationHistory::SUSPEND `suspended_until`

### DIFF
--- a/app/models/moderator_action.rb
+++ b/app/models/moderator_action.rb
@@ -65,7 +65,7 @@ class ModeratorAction < ApplicationRecord
   after_save :delete_resource_update_actions, if: ->( moderator_action ) { moderator_action.action == HIDE }
   after_destroy :notify_resource_on_destroy
 
-  audited only: [:suspended_until, :last_edited_by_user_id],
+  audited only: [:reason, :suspended_until, :last_edited_by_user_id],
     if: ->( moderator_action ) { moderator_action.action == SUSPEND }
 
   def self.current_private_actions
--- a/app/models/moderator_action.rb
+++ b/app/models/moderator_action.rb
@@ -42,6 +42,10 @@ class ModeratorAction < ApplicationRecord
   validates :action, inclusion: ACTIONS
   validates :reason, length: { minimum: MINIMUM_REASON_LENGTH, maximum: MAXIMUM_REASON_LENGTH }
   validates :reason, length: { maximum: MAXIMUM_SUSPEND_REASON_LENGTH }, if: -> { action == SUSPEND }
+  validates :audit_comment, presence: true, on: :update, if: -> { action == SUSPEND }
+  validates :audit_comment,
+    length: { minimum: MINIMUM_REASON_LENGTH, maximum: MAXIMUM_REASON_LENGTH },
+    allow_blank: true
   validate :only_curators_and_staff_can_hide, on: :create
   validate :only_staff_can_make_private
   validate :only_staff_can_rename, on: :create
@@ -59,6 +63,9 @@ class ModeratorAction < ApplicationRecord
   after_save :notify_resource
   after_save :delete_resource_update_actions, if: ->( moderator_action ) { moderator_action.action == HIDE }
   after_destroy :notify_resource_on_destroy
+
+  audited only: [:suspended_until, :last_edited_by_user_id],
+    if: ->( moderator_action ) { moderator_action.action == SUSPEND }
 
   def self.current_private_actions
     moderated_private_resource_ids = ModeratorAction.where( private: true ).pluck( :resource_id )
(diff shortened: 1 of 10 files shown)


app/views/moderator_actions/_suspension_form.html.haml
```diff
@@ -7,14 +9,14 @@
   - if current_user.is_admin?
     - reason_options = ModeratorAction::SUSPENSION_REASONS.keys.map { |key| [t( "suspension_reasons.#{key}" ), key] }
     - reason_options << [t( :suspension_reason_custom ), "custom"]
-    = f.form_field :suspension_reason, label: t( :suspension_reason_label ) do
+    = f.form_field :suspension_reason, label: t( editing ? :edit_reason_label : :suspension_reason_label ) do
```
pleary: This form field is for the dropdown select for suspension reason, not the reason for editing. We should keep the label `suspension_reason_label` even if the form is rendered for an edit.

app/views/moderator_actions/_suspension_form.html.haml
```diff
@@ -7,14 +9,14 @@
...
       = select_tag :suspension_reason, options_for_select( reason_options, selected_reason || reason_options.first[1] ), class: "form-control", id: "suspension_reason"
     .custom-reason-field{ style: selected_reason == "custom" ? "" : "display: none;" }
-      = f.text_area :reason, reason_opts
```
pleary, line 13: I don't think we want to replace the `:reason` text_area with an `:audit_comment`, rather I think we want to add a new textarea for `:audit_comment`. For example - what if the edit wants to make is to edit the custom text of the suspension reason? They should be able to do so, and also leave an audit comment stating why they edited the reason. There is also some JS logic that hides or populates the contents of the form field with ID `moderator_action_reason` based on the suspension_reason dropdown. We want that populating the ModeratorAction `reason` attribute, not the `audit_comment`

app/views/moderator_actions/_suspension_form.html.haml
```diff
@@ -7,14 +9,14 @@
...
+      = f.text_area reason_field, reason_opts.merge( id: "moderator_action_reason" )
       %p.help-block= t( :suspension_reason_custom_note )
-    = f.hidden_field :reason, id: "hidden_reason"
+    = f.hidden_field reason_field, id: "hidden_reason"
   - else
-    = f.text_area :reason, reason_opts
+    = f.text_area reason_field, reason_opts
```
pleary: One thing I noticed - we currently allow users to edit suspensions they create. Until the above change is made to have one form field for the suspension reason, and one for the audit_comment, the edit suspension form for non-admins doesn't allow the user to edit anything about the suspension. It just gives the user the ability to provide an `Edit reason` without letting them edit anything. Restoring the suspension reason alongside the edit reason should fix this issue

app/models/moderator_action.rb
```diff
@@ -60,6 +65,9 @@ class ModeratorAction < ApplicationRecord
   after_save :delete_resource_update_actions, if: ->( moderator_action ) { moderator_action.action == HIDE }
   after_destroy :notify_resource_on_destroy
 
+  audited only: [:suspended_until, :last_edited_by_user_id],
```
pleary: Should we add `:reason` here, since that is one of the suspension attributes we want users to be able to modify? (e.g. there was a typo, they want to change the wording, tone, etc)

pleary (changes requested):
Mostly looks pretty good. I left a few comments for suggested changes

app/views/moderator_actions/_suspension_form.html.haml
caseyhofford: Good catch.

app/views/moderator_actions/_suspension_form.html.haml
```diff
@@ -10,12 +11,14 @@
     = f.form_field :suspension_reason, label: t( :suspension_reason_label ) do
       = select_tag :suspension_reason, options_for_select( reason_options, selected_reason || reason_options.first[1] ), class: "form-control", id: "suspension_reason"
     .custom-reason-field{ style: selected_reason == "custom" ? "" : "display: none;" }
-      = f.text_area :reason, reason_opts
+      = f.text_area :reason, reason_opts.merge( id: "moderator_action_reason" )
```
pleary: This no longer needs an explicit ID option as `moderator_action_reason` will automatically be the textarea ID

config/locales/en.yml
```diff
@@ -5725,6 +5725,12 @@ en:
   suspension_reason_custom_note: Please note this reason will be displayed to the suspended user.
   # note displayed when a curator is editing a suspension
   suspension_reason_edit_note: Please update the reason and explain why you are changing the user's suspension.
+  # label for the original suspension reason displayed when editing
+  original_suspension_reason: Original suspension reason
```
pleary: This key doesn't appear to be used at all. If not, it can be removed

pleary (approved):
Thanks for the changes. I left two comments about pretty minor non-blocking changes we could make. Otherwise this looks OK to merge to me

caseyhofford:
Thanks for the review @pleary, I've implemented those changes!

PR #4873, opened 2026-04-14
feat: enable task logging in auto_unsuspend rake task, related to OPS-348

pleary:


lib/tasks/inaturalist.rake
caseyhofford, line 52: Is it better to use `TaskLogger.error` here as well so we get the same `TaskLogger` details?

caseyhofford (approved):
Looks good, approved! Thanks for adding this, added a comment about whether all `Rails.logger.error` calls in this task should be replaced with `TaskLogger` for consistency.

lib/tasks/inaturalist.rake
pleary: yeah, it might be - or at least do both since there's some defensive programming around the case if `task_logger` isn't defined because a `log_task_name` wasn't given. I'll merge as-is after specs finish, and we can make that improvement eventually just to not delay release much longer

PR #4874, opened 2026-04-14
feat: Final homepage changes

sundisc:
Final homepage changes for full production deployment.

- Fixes observation stat
- Removes admin gating and displays homepage v2 in logged out state for default site only
- Switches to correct meta attribute for "description"


pleary (changes requested):
The welcome controller specs will need to be updated to reflect the change of permissions there. Other than that this looks OK to me

pleary (approved):

PR #4875, opened 2026-04-14
web 969 suspension api changes

caseyhofford:
Updates JSON API response for suspended users to include message and suspended_until time. Achieves this by updating the SuspendedError class to include a default #to_h method.


pleary (approved):
Looks good to me!

PR #4878, opened 2026-04-17
feat: Show log in in favor of sign up, WEB-1003

sundisc:
Hide sign up instead of log in and reduce spacing on the or to make it look more natural.

https://github.com/user-attachments/assets/b91ea1db-1aa3-4e04-828d-5602afbe3f76


pleary (approved):
Thanks!

PR #4882, opened 2026-04-20
fix: prevent excessively long delayed job unique hash

pleary:


caseyhofford (approved):
Looks good, ran the test locally, indexing is idempotent so changing the hash shouldn't cause any issues if `Project#sync_delegated_projects` were called before and after the hash changes.

PR #4884, opened 2026-04-21
feat: Add search link when search bar is not displayed, WEB-1004

sundisc:
- Add `Search` to `More` menu when the search bar isn't displayed in the header (i.e. user is logged out).
- Remove a lot of unnecessary overrides that were either useless or caused header dropdown menus to render too high.

**Search link demo:**
https://github.com/user-attachments/assets/b7317984-4451-449a-8383-94669d86f9ff

**Menu height before:**
Screenshot 2026-04-21 at 4 26 58 PM
**Menu height after:**
Screenshot 2026-04-21 at 4 27 09 PM

### DIFF
--- a/app/views/shared/_header.html.haml
+++ b/app/views/shared/_header.html.haml
@@ -111,7 +111,8 @@
           = t(:more)
           %i.fa.fa-angle-down
         %ul.dropdown-menu{ aria: { labelledby: "header-more-dropdown-toggle" } }
-          %li.header-md.header-lg= link_to t( :search ), search_path
+          - unless header_search_open
+            %li.header-md.header-lg= link_to t( :search ), search_path
           - if logged_in?
             %li.header-xs.header-sm= link_to t( :identify ), identify_observations_path
           %li= link_to t( :taxa_info ), taxa_path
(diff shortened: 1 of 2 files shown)


app/views/shared/_header.html.haml
```diff
@@ -111,6 +111,8 @@
           = t(:more)
           %i.fa.fa-angle-down
         %ul.dropdown-menu{ aria: { labelledby: "header-more-dropdown-toggle" } }
+          - unless header_search_open
```
pleary: On all pages except when `@skip_external_connections` is true, the search is always displayed in the full-width version of the page. For example load an observation page and the search bar or button exists. header_search_open is derived from essentially `session[:header_search_open].yesish?`. When a user minimizes the search bar, that becomes false. But that doesn't remove search from the header, it just collapses it down to an icon. So - if you minimize the search bar, then reload the page, you have a minimized search icon in the header, plus this search option in the More menu.

I think what we want to do here is change `unless header_search_open` to `if @skip_external_connections`. That way we only include search in the "More" menu if it isn't displayed otherwise (we're checking @skip_external_connections to see if it should be shown in the first place).

pleary (approved):
I left a comment suggesting one minor change, but generally this is looking good to me

PR #4885, opened 2026-04-21
feat: On mobile account creation prompt for app download, WEB-942

sundisc:
https://linear.app/inaturalist/issue/WEB-942/on-mobile-account-creation-prompt-for-app-download

~~This ticket had two requests:~~
- ~~Show sticky banner with link to download app on iOS only~~
    - ~~Break out sticky footer from `welcome_v2` to prevent duplication~~
    - ~~Include new UTM campaign codes for dashboard link~~
- ~~Always show dashboard as responsive~~
    - ~~Add some spacing between subscribe buttons to make them look better by adding an imitation of a Bootstrap 5 utility class `mt-sm-1`~~

The ticket was descoped to just create a shared instance of the app download prompt for if we eventually need it for the Android release. It also includes translations for the text in the prompt.


app/views/shared/_ios_app_banner.html.haml
```diff
@@ -0,0 +1,8 @@
+- utm_campaign_params = local_assigns[:utm_campaign_params]
+- url = "#{redirect_link_path(31)}?#{[utm_campaign_params, 'utm_medium=owned_web&utm_content=sticky_footer_cta'].compact.join('&')}"
+= link_to url, class: "wv2-app-banner", "aria-label": "Download iNaturalist on the App Store" do
```
pleary: I think we should have this string go through i18n

app/views/shared/_ios_app_banner.html.haml
```diff
@@ -0,0 +1,8 @@
...
+  = image_tag "welcome_v2/app-icon.webp", class: "wv2-app-banner__icon", alt: ""
+  .wv2-app-banner__text
+    = image_tag "welcome_v2/inat-wordmark.webp", class: "wv2-app-banner__wordmark", alt: "iNaturalist"
+    %p.wv2-app-banner__tagline Discover nature around you
```
pleary: Similarly, I think we'll want to internationalize this string and `Use app` on the next line

pleary (changes requested):
Thanks for updating the name of `inat-wordmark.webp`. Some background on that - the image wasn't loading for me locally in development. I tracked down the issue to essentially the same as https://github.com/rails/sprockets/issues/749. For us `-wordmark` was being considered the file fingerprint. That combined with the fact that `inat.webp` also exists, and it's etag is not `-wordmark`, was resulting in sprockets rendering a `:not_found` error for that file.

I've requested some additional internationalization, but otherwise this seems OK to me. I've also left a comment on the Linear issue clarifying if this should be dismissible. I wouldn't mind waiting for an answer to that question as my personal opinion is it should be dismissible on release, but currently is not.

sundisc:
Adding the localization, but converting to draft for now since we want to work on some responsive changes and maybe change the banner behavior: https://linear.app/inaturalist/issue/WEB-942/on-mobile-account-creation-prompt-for-app-download#comment-1dc40d35

sundisc:
The ticket was descoped to just create a shared instance of the app download prompt for if we eventually need it for the Android release. It also includes translations for the text in the prompt.

pleary (approved):
Looks good to me!

caseyhofford (approved):
I tested it out on my iphone with safari and it looks good. The redirect_link did not work for me locally (sent me to `/redirect_link` which gave a 404), but maybe this is specific to my local environment or site setup.

sundisc:
> I tested it out on my iphone with safari and it looks good. The redirect_link did not work for me locally (sent me to `/redirect_link` which gave a 404), but maybe this is specific to my local environment or site setup.

It's your local environment - RedirectLink is a model, and ID 31 probably doesn't exist locally for you.

caseyhofford:
> > I tested it out on my iphone with safari and it looks good. The redirect_link did not work for me locally (sent me to `/redirect_link` which gave a 404), but maybe this is specific to my local environment or site setup.
> 
> It's your local environment - RedirectLink is a model, and ID 31 probably doesn't exist locally for you.

Thanks @sundisc, that makes sense, add a local RedirectLink and worked as expected.

PR #4891, opened 2026-04-30
feat: update s3 upload to use transfer_manager; update kt-paperclip

pleary:
Some changes to address the deprecation of `upload_file` on `Aws::S3::Object` and using `Aws::S3::TransferManager#upload_file` instead as recommended. The `kt-paperclip` update merges in changes in that package to address the same

### DIFF
--- a/lib/darwin_core/archive.rb
+++ b/lib/darwin_core/archive.rb
@@ -820,20 +820,21 @@ module DarwinCore
 
     def upload_to_aws_s3
       return unless @opts[:aws_s3_path]
+
       s3_config = YAML.load_file( File.join( Rails.root, "config", "s3.yml" ) )
       client = ::Aws::S3::Client.new(
         access_key_id: s3_config["access_key_id"],
         secret_access_key: s3_config["secret_access_key"],
         region: CONFIG.s3_region
       )
-      resource = Aws::S3::Resource.new( client: client )
-      bucket = resource.bucket( CONFIG.s3_bucket )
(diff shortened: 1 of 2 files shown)


lib/darwin_core/archive.rb
```diff
@@ -820,20 +820,21 @@ def self.generate_ranked_ancestors( taxon_id, cached_taxa = { } )
...
         access_key_id: s3_config["access_key_id"],
         secret_access_key: s3_config["secret_access_key"],
         region: CONFIG.s3_region
       )
-      resource = Aws::S3::Resource.new( client: client )
-      bucket = resource.bucket( CONFIG.s3_bucket )
-      object = bucket.object( @opts[:aws_s3_path] )
-      object.upload_file( @opts[:path], {
+      transfer_manager = ::Aws::S3::TransferManager.new( client: client )
+      transfer_manager.upload_file(
```
caseyhofford, line 831: This all looks right, but I haven't been able to test file upload locally, and wasn't sure if I should try running something like `tools/dwca.rb` locally to test the DarwinCore upload path. I'm sure this will work and is low risk (I think) but let me know if there is a way to test this behavior locally.

caseyhofford (approved):
Looks good, couldn't test uploads locally but verified against the AWS docs https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/TransferManager.html#upload_file-instance_method

PR #4893, opened 2026-05-01
Use fundraise up form on announcements

caseyhofford:
- **Removed shadow DOM rendering** from announcements. Announcement HTML is now rendered inline with scoped `<style>` tags (via `@scope`) instead of inside a `<template shadowrootmode="open">`.
- **Added `prefers_embed_fru` preference** to announcements. When enabled, the FRU widget script and supporting JS are loaded into the page `<head>` via `content_for(:extrahead)`.
- **`body_for_mobile`** method on `Announcement` wraps the announcement body in a full HTML document with the FRU script for mobile app WebView rendering.

- **hideFruButtons** function is defined in `fundraise_up/fundraise_up.js` and is loaded into the page when an announcement has `prefers_embed_fru` enabled. It can be called from the announcement body.

```html
<script>hideFruButtons();</script>
```


app/assets/javascripts/announcements/dismiss.js
```diff
@@ -0,0 +1,12 @@
+function initAnnouncementDismiss( announcementId ) {
```
caseyhofford: Moves existing announcement dismiss logic from app/views/announcements/_announcement.html.haml:14 into a js file.

app/assets/javascripts/fundraise_up/fundraise_up.js
```diff
@@ -0,0 +1,37 @@
+function initFruFallback( containerSelector ) {
```
caseyhofford: Adds a fallback to handle cases where fundraise up is blocked on the client (e.g. adblocker, or strict cross-origin policies). The fallback button is always rendered at app/views/announcements/_announcement.html.haml:23 and is hidden if the FRU iframe successfully renders.

app/assets/javascripts/fundraise_up/fundraise_up.js
```diff
@@ -0,0 +1,37 @@
...
+  var observer = new MutationObserver( function ( ) {
+    if ( container.querySelector( "iframe" ) ) {
+      fallback.style.display = "none";
+      observer.disconnect( );
+    }
+  } );
+  observer.observe( container, { childList: true, subtree: true } );
+}
+
+function hideFruButtons() {
```
caseyhofford, line 1: This is kind of hacky, but it seems like the only way to get a currency value on a single button. I am going to reach out to FundraiseUp support to ask if they can just allow this natively, but for now this is what I came up with. The helper hides all but the first button of a button group. The idea is this function can be called from within a `<script>` tag in the announcement HTML if we are trying to display only one button.

app/views/announcements/_announcement.html.haml
```diff
@@ -1,30 +1,36 @@
...
-            %i.fa.fa-close
+
+- if announcement.prefers_embed_fru? && !content_for?( :fundraise_up_script )
+  - content_for( :extrahead ) do
+    = javascript_include_tag "fundraise_up/fundraise_up_widget_init"
+    = javascript_include_tag "fundraise_up/fundraise_up"
+  - content_for( :fundraise_up_script, "loaded" )
+
+- scoped_body = Nokogiri::HTML.fragment( announcement.body )
+- scoped_body.css( "style" ).each do |style_tag|
```
caseyhofford, line 11: This replaces the shadow DOM for CSS isolation.

app/views/announcements/_announcement.html.haml
```diff
@@ -1,30 +1,43 @@
 - preview = defined?( preview ) && preview
 - create_announcement_impression( announcement ) unless preview
-%div{ id: "announcement-#{announcement.id}" }
```
caseyhofford, line 3: I recommend viewing this diff with "hide whitespace" enabled (example below), to make it easier to interpret. There are a lot of indentation changes being identified as line changes without that option enabled.

Image

app/assets/javascripts/fundraise_up/fundraise_up.js
```diff
@@ -0,0 +1,37 @@
+function initFruFallback( containerSelector ) {
+  var container = document.querySelector( containerSelector );
+  if ( !container ) return;
+  var fallback = container.querySelector( ".fru-fallback-link" );
+  if ( !fallback ) return;
+  var fruAnchor = container.querySelector( "a:empty[href^=\"#\"][style*=\"display: none\"]" );
```
caseyhofford: This is the format of the tag provided for embeds from FRU. Although it would be better to select by ID, the codes from FRU do not set and ID, and I think this is a fairly reliable selector, especially within the announcement container.

app/assets/javascripts/fundraise_up/fundraise_up.js
```diff
@@ -0,0 +1,37 @@
...
+      fallback.style.display = "none";
+      observer.disconnect( );
+    }
+  } );
+  observer.observe( container, { childList: true, subtree: true } );
+}
+
+function hideFruButtons() {
+  var container = document.querySelector( ".announcement" );
+  var observer = new MutationObserver( function ( ) {
```
caseyhofford: This is kind of messy and I don't love it, but it seems like the only way to get a currency value on a single button. I am going to reach out to FundraiseUp support to ask if they can just allow this natively, but for now this is what I came up with.

app/assets/javascripts/fundraise_up/fundraise_up.js
```diff
@@ -0,0 +1,37 @@
...
+    var iframe = container.getElementsByTagName( "iframe" )[0];
+    if ( !iframe || !iframe.contentDocument ) return;
+    observer.disconnect( );
+    var innerObserver = new MutationObserver( function ( ) {
```
caseyhofford, line 7: We need to have two MutationObservers because after the iframe is added to the DOM, there is still a delay before the actual button elements render.

app/assets/templates/example_announcement.html
```diff
@@ -0,0 +1,44 @@
+<style type="text/css">
```
caseyhofford: This was taken from one of Thea's actual announcements, this was added mostly just to give me something to copy paste from. I think there might be value in having an example of an announcement body version controlled, but I'm not sure it belongs here, so I'm very willing to delete this, but at the very least it is useful for testing.

caseyhofford (commented):
Oops, some of my comments ended up as a review 🤷

app/helpers/application_helper.rb
```diff
@@ -1828,4 +1828,8 @@ def fundraise_up_js
       </script>
     HTML
   end
+
+  def fru_donate_url
```
caseyhofford: This could also be an environment variable, but we may want to update this function to take optional params and change the URL at some point.

app/views/announcements/_announcement.html.haml
```diff
@@ -1,30 +1,43 @@
...
-          = button_to dismiss_announcement_path( announcement ), method: :put, remote: true, form_class: "dismiss-announcement", class: "dismiss-announcement-link", title: t(:dismiss) do
-            %i.fa.fa-close
+    } );
+  .bootstrap
+    .announcement
+      = scoped_body.to_html.html_safe
+      - if announcement.dismissible? && current_user && !preview
+        = button_to dismiss_announcement_path( announcement ), method: :put, remote: true, form_class: "dismiss-announcement", class: "dismiss-announcement-link", title: t(:dismiss) do
+          %i.fa.fa-close
+      - if announcement.prefers_embed_fru?
```
caseyhofford: Adds a fallback button if we are using the FRU embed, this is hidden by initFruFallback once the iframe from FRU has loaded.

app/assets/javascripts/fundraise_up/fundraise_up.js
pleary: Can you explain when we'd want to use this? I tested with the example announcement and that is not calling this method, nor is anything else, yet I see one button. Is this still used for something?

app/helpers/application_helper.rb
pleary: I think this will need to be heavily customized, potentially for each different announcement. Could this be an attribute of announcements? Perhaps announcement creators could enter a URL like this instead of, or in addition to, using the `prefers_embed_fru` checkbox?

app/views/announcements/_announcement.html.haml
```diff
@@ -1,30 +1,43 @@
...
+  - content_for( :extrahead ) do
+    = fundraise_up_js.html_safe
+    = javascript_include_tag "fundraise_up/fundraise_up"
+  - content_for( :fundraise_up_script, "loaded" )
+
+- scoped_body = Nokogiri::HTML.fragment( announcement.body )
+- scoped_body.css( "style" ).each do |style_tag|
+  - style_tag.content = "@scope {\n#{style_tag.content}\n}"
+%div{ id: "announcement-#{announcement.id}", class: "announcement" }
+  = render "shared/common_stylesheets"
```
pleary: All these stylesheet and JS includes should be removed if we're moving away from shadowdom. Their inclusion here, without a shadowdom, is causing CSS problems due to changes order of style definitions

app/views/announcements/_announcement.html.haml
```diff
@@ -1,30 +1,43 @@
...
+  = stylesheet_link_tag "bootstrap_bundle"
+  = stylesheet_link_tag "bootstrap-rtl" if @rtl
+  = javascript_include_tag "jquery/jquery-1.11.3.min.js"
+  :javascript
```
pleary: We also don't need this JS block if not using shadowdom

app/views/announcements/_announcement.html.haml
```diff
@@ -1,30 +1,43 @@
...
-            %i.fa.fa-close
+    } );
+  .bootstrap
+    .announcement
+      = scoped_body.to_html.html_safe
+      - if announcement.dismissible? && current_user && !preview
+        = button_to dismiss_announcement_path( announcement ), method: :put, remote: true, form_class: "dismiss-announcement", class: "dismiss-announcement-link", title: t(:dismiss) do
+          %i.fa.fa-close
+      - if announcement.prefers_embed_fru?
+        %a{ class: "fru-fallback-link" , href: fru_donate_url }
```
pleary: Do we need a fallback? I'll send a screen recording of what I'm seeing, but this fallback is creating some element jitter as it is initially displayed then later removed but only after JS loads. My inclination would be to remove it unless there are a lot of people affected by this.

app/views/donate/index.html.haml
```diff
@@ -33,6 +33,13 @@
     .col-md-5.col-md-offset-1.col-xs-12
       .fundraiseup-form
         %a{ href: "#XWLLKFCZ", style:"display: none" }
+        .fru-fallback
```
pleary: This fallback is not getting removed because nothing matches the `fru-fallback-link` selector triggering removal of fallbacks. As I said in another comment, the fallback behavior is pretty jarring, especially on deskton, and I'd be in favor of removing it unless there's a compelling reason to keep it.

pleary (changes requested):
I left some initial thoughts and comments. Some changes are needed, and others I mentioned for consideration. Overall it generally works OK. I did test with a VPN and was able to see different currencies. The delay for the fundraise donation button to load I suspect is unavoidable, but also a little distracting

app/views/announcements/_announcement.html.haml
pleary: Just noting that scoped styles behave differently than styles in shadowdoms. With shadodoms you can use selectors from the root of the document, but it appears scope selectors start at the scoped element. Just noting that in case some CSS selectors that previously worked stop working when switching to scopes

app/views/announcements/_announcement.html.haml
caseyhofford: It's hard to say how often this will actually happen in practice, likely not too often. My main rationale for having a fallback for this is to be extra defensive against the possibility of breaking the donation flow. Do we have any monitoring tools in place to send reports or errors from the frontend so we could track the frequency of these issues?

My only other concern here is that a layout shift will still happen without the fallback, because the embed will load after the rest of the announcement. Because the embeds can be different heights, this may be difficult to address reliably. Adding a more reliable fallback that actually renders in the same location as these buttons might be the best solution. For now I have removed the fallback, but am open to adding it back in with better default placement if preferred.

app/views/announcements/_announcement.html.haml
caseyhofford: Good call out, from spot checking announcements on production, it seems like announcements generally use styles inline on elements, so we shouldn't run into issues with this, and the scope will ensure that stylesheets that are included do not cause changes in the broader webpage. I also updated the scope selector to use `@scope (#announcement-#{announcement.id}) { ... ` so that css styles won't mix between announcements.

app/views/donate/index.html.haml
caseyhofford: Removed in e56396db892a3191e754ab26d382e12612f144ae

app/views/announcements/_announcement.html.haml
caseyhofford: Good call, fixed in 791a92a95653bd6049c1cc5e3b897b9991dcdab6

app/views/announcements/_announcement.html.haml
caseyhofford: Makes sense, good catch, fixed in 791a92a95653bd6049c1cc5e3b897b9991dcdab6

app/assets/javascripts/fundraise_up/fundraise_up.js
caseyhofford: This was included because Thea requested to have the option to include a single button with a currency value on it. Unfortunately the FRU embed only supports currency values on ["Button Groups"](https://fundraiseup.com/docs/elements-catalog/#button-group). This JS is intended to hide all but the first button in a button group as a workaround. Try embedding this component with the hideFruEmbed() call to test this out.

```
<a href="#XYAJZHRA" style="display: none"></a>
<script>
  hideFruButtons()
</script>
```

app/helpers/application_helper.rb
caseyhofford: This was initially going to be just the URL for the fallback button, so I will remove this as well for now, unless we go back to using a more reliable fallback. Done in a00504b0c227b69d30761c4be91f7434b41d0cc0

caseyhofford (commented):
Removed the fallback for now, and implemented other feedback. Take a look at the behavior now, the first announcement here is with an explicit min-height which would need to be included in the body when creating the announcement.

https://github.com/user-attachments/assets/d02fadbf-2938-4f52-aa90-c262b15672ea

caseyhofford:
@pleary that force push shouldn't mess with the "changes since last review" since it just amended added after the last review, but let me know if that still caused issues with reviewing new changes. The changes since last review should all be in this diff https://github.com/inaturalist/inaturalist/pull/4893/changes/1985afca652b4d355d95bad2b3faae5cd604dce1..a00504b0c227b69d30761c4be91f7434b41d0cc0

app/assets/javascripts/fundraise_up/fundraise_up.js
pleary: Great - thanks. I tested the buttons with that method and I can confirm it hides all but the first one

app/views/donate/index.html.haml
```diff
@@ -33,6 +33,11 @@
     .col-md-5.col-md-offset-1.col-xs-12
       .fundraiseup-form
         %a{ href: "#XWLLKFCZ", style:"display: none" }
+      %noscript
+        %p= t( :donate_enable_javascript )
+      = javascript_include_tag "fundraise_up/fundraise_up"
+      :javascript
+        initFruFallback( ".fundraiseup-form" );
```
pleary: This method has since been removed. Since we're not using a fallback right now, maybe all the changes to this file can be reverted?

config/locales/en.yml
```diff
@@ -2018,6 +2018,8 @@ en:
   donate: DONATE
   # Imperative verb; link to a page where a user can donate money to iNaturalist
   donate_: Donate
+  # Message shown on the donate page when JavaScript is disabled
+  donate_enable_javascript: Please enable JavaScript to use the donation form.
```
pleary: If we remove this string from the donate page, it won't have any other uses right now, so perhaps could be removed. Unless we think we'll want to use it later and want to get it translated

app/views/announcements/_announcement.html.haml
pleary: Agreed that the embed loading is unavoidable. But for me the fallback loading above the announcement then disappearing was much more jarring. If we can eventually line up a fallback with where the donate button or form would normally be, that would be much better. We can test options over time - I'm all for merging in as-is with no fallbacks for now

pleary (approved):
Thanks Casey - this is all looking good. In my opinion, removing the fallback for now is better in the short term. I like the idea of a fallback that would be more visually inline. Let's discuss options and ways to measure in a future meeting, but not let that hold up release.

I suggested some small changes which aren't blockers. OK with me to merge

config/locales/en.yml
caseyhofford: Good call! Done in f6a1d65e8ec4b605741a92692e9d8fe9b805555c

app/views/announcements/_announcement.html.haml
caseyhofford: Sounds good! Added a linear ticket to track this: https://linear.app/inaturalist/issue/WEB-1023/add-fru-fallback-wo-layout-shift

app/views/donate/index.html.haml
caseyhofford: Good call, reverted all changes to this file.

PR #4901, opened 2026-05-10
make mappable decide on circle "diameter" check with square diagonal

vizagite:
related bug on mobile app that is fixed recently by @jtklein  - https://github.com/inaturalist/iNaturalistReactNative/pull/3406

the positional accuracy is radius and comparing to diagonal of square is not the intended check here I guess, as such diagonal is a diameter of circle then

### DIFF
--- a/app/models/observation.rb
+++ b/app/models/observation.rb
@@ -3153,7 +3153,7 @@ class Observation < ApplicationRecord
 
   def calculate_mappable
     return false if latitude.blank? && longitude.blank?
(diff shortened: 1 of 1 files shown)


jwidness:
This check was intentional -- it was designed to keep obscured observations mappable. If this merges as-is with no other changes, all obscured observations will become unmappable because their `public_positional_accuracy` is the `uncertainty_cell_diagonal_meters`.

PR #4902, opened 2026-05-12
feat: add taxon range geom validations; clean out of bounds place and taxon range geoms

pleary:
In order to upgrade the API to Node.js 24.x, we'll want to upgrade mapnik to >4.6.0. In order to do that, we need to fix some of our geometries that have coordinates outside the expected bounds so they render properly with the updated mapping library. This PR updates PlaceGeometry and TaxonRange geoms that have coordinates outside the expected bounds. It does so by making a temporary column, adding cleaned versions of the small subset of invalid geoms into those columns, then copies the cleaned versions back into their original columns while removing the temporary columns.

This also adds geometry validation to TaxonRange similar to what we have in PlaceGeometry to prevent the creation of new geometries that have out of bounds coordinates, to ensure future geometries don't have similar problems.

Cleaning these geometries will allow us to update mapnik in the API, setting us up to upgrade to Node.js 24.x


spec/models/taxon_range_spec.rb
```diff
@@ -1,50 +1,99 @@
...
+              -122.239551544189 37.8719807055375,
+              -122.247619628906 37.8547693305679
+            )
+          )
+        )
+      WKT
+      expect( taxon_range ).to be_valid
+    end
+
+    it "should be invalid with a two-point polygon" do
```
caseyhofford: NIT: I'd have a slight preference to test a 3 point polygon here since we require 4 points. I think testing 2 points here might suggest that three points would be valid.

spec/models/taxon_range_spec.rb
```diff
@@ -1,50 +1,99 @@
...
+      expect( taxon_range ).not_to be_valid
+    end
+
+    it "should be invalid with a latitude greater than 90" do
+      taxon_range = TaxonRange.make!
+      impossible_polygon = "MULTIPOLYGON(((0 89,0 91,1 91,0 91,0 89)))"
+      taxon_range.geom = impossible_polygon
+      expect( taxon_range ).not_to be_valid
+      expect( taxon_range.errors.size ).to eq 1
+    end
```
caseyhofford, line 89: We could add a spec for longitude as well:
```suggestion
    end
    
    it "should be invalid with a longitude greater than 180" do
      taxon_range = TaxonRange.make!
      impossible_polygon = "MULTIPOLYGON(((180 89,181 89,180 89,180 89,180 89)))"
      taxon_range.geom = impossible_polygon
      expect( taxon_range ).not_to be_valid
      expect( taxon_range.errors.size ).to eq 1
    end

```

spec/models/taxon_range_spec.rb
```diff
@@ -1,50 +1,99 @@
...
-describe TaxonRange, "create_kml_attachment" do
-  it "should create an kml attachment from geometry" do
-    tr = without_delay { make_taxon_range_with_geom }
-    expect( tr.range ).not_to be_blank
-    kml = File.open( tr.range.path ).read
-    expect( kml ).to be =~ /<kml /
+describe TaxonRange do
+  describe "create_kml_attachment" do
+    it "should create an kml attachment from geometry" do
+      tr = without_delay { make_taxon_range_with_geom }
```
caseyhofford, line 16: Just a note, locally I had to install `gdal` in order for this command to succeed. The `ogr2ogr` command failed at https://github.com/inaturalist/inaturalist/blob/clean-geoms/app/models/taxon_range.rb#L99 . We could [add gdal installation](https://gdal.org/en/stable/download.html#binaries) to [the setup guide.](https://github.com/inaturalist/inaturalist/wiki/Development-Setup-Guide).

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      # clean geoms where some lngs are out of bounds
+      execute <<-SQL
+        UPDATE #{table_name}
+        SET cleaned_geom = ST_Multi(
+          ST_CollectionExtract(
+            ST_MakeValid(
+              ST_Intersection(
+                geom,
+                ST_MakeEnvelope(-179.999999, -90, 179.999999, 90)
```
caseyhofford, line 34: Do we need to specify an SRID? Are all stored geometries using the same projection? I verified that the staging lite database PlaceGeometries only has geometries stored with SRID=0 however I don't have any TaxonRange records locally to validate this against. Alternatively we might want to just pass the last param as `ST_SRID(geom)`, so that the envelope will use the same spatial reference as the geometry.
`ST_MakeEnvelope(-179.999999, -90, 179.999999, 90, ST_SRID(geom))`

We may want SRID 4326, since [we assume that in inaturalistAPI](https://github.com/inaturalist/iNaturalistAPI/blob/main/lib/map_generator.js#L83).

Local output of testing SRID on geometries in the staging lite database:
```
3.3.7 :012 > PlaceGeometry.all.map { |p| p.geom.srid }.tally
PlaceGeometry Load (842.0ms)  SELECT "place_geometries".* FROM "place_geometries"
 => {0=>3831}
3.3.7 :013 > TaxonRange.all.map { |p| p.geom.srid }.tally
  TaxonRange Load (4.8ms)  SELECT "taxon_ranges".* FROM "taxon_ranges"
   (4.1ms)  SELECT f_geometry_column,coord_dimension,srid,type FROM geometry_columns WHERE f_table_name='taxon_ranges'
 => {}
```

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
```diff
@@ -0,0 +1,66 @@
...
+    add_column :place_geometries, :cleaned_geom, :geometry
+    add_column :taxon_ranges, :cleaned_geom, :geometry
+
+    [:place_geometries, :taxon_ranges].each do | table_name |
+      # update lngs of geoms where all lngs are out of bounds
+      execute <<-SQL
+        UPDATE #{table_name}
+        SET cleaned_geom = ST_Translate(
+          geom,
+          -360 * FLOOR((ST_XMin(geom) + 180) / 360),
```
caseyhofford, line 14: This looks right to me, and from some manual testing with this formula does handle all longitude values over 180 and under 180:
```
3.3.7 :021 > min = -190
3.3.7 :022 > min + (-360 * ((min + 180) / 360).floor)
 => 170
3.3.7 :023 > min = 190
3.3.7 :024 > min + (-360 * ((min + 180) / 360).floor)
 => -170
3.3.7 :025 > min = 400
3.3.7 :026 > min + (-360 * ((min + 180) / 360).floor)
 => 40
3.3.7 :027 > min = -400
3.3.7 :028 > min + (-360 * ((min + 180) / 360).floor)
 => -40
3.3.7 :029 >
```

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
```diff
@@ -0,0 +1,66 @@
...
+          )
+      SQL
+
+      # clean geoms where some lngs are out of bounds
+      execute <<-SQL
+        UPDATE #{table_name}
+        SET cleaned_geom = ST_Multi(
+          ST_CollectionExtract(
+            ST_MakeValid(
+              ST_Intersection(
```
caseyhofford, line 32: I believe this will trim the portion of the geometry extending past the antimeridian, which will lose the portions of the bounds that exceed the antimeridian. I think we need to create two polygons, one on each side of the antimeridian.

I did some digging to try and find alternative approaches to splitting the polygons in the database, but didn't find anything that seemed more reliable. There is a PostGIS `::geography` type, that uses spheroid calculations rather than projections, however it is less efficient and has limited calculations that can be run. https://postgis.net/docs/using_postgis_dbmanagement.html#PostGIS_Geography

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
```diff
@@ -0,0 +1,66 @@
...
+      execute <<-SQL
+        UPDATE #{table_name}
+        SET geom = cleaned_geom
+        WHERE cleaned_geom IS NOT NULL
+          AND NOT ST_IsEmpty(cleaned_geom)
+
+      SQL
+    end
+
+    remove_column :place_geometries, :cleaned_geom
```
caseyhofford, line 59: I feel a little hesitant to remove all of the source geometries in the same migration. I think we might benefit from first moving the `geom` column to a backup (e.g. `original_geom`) which we can remove after it has been live for some amount of time (e.g. a few weeks) to see if we get reports of any new issues with geometries. This provides an easy rollback strategy if we discover new issues with any polygons.

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
```diff
@@ -0,0 +1,66 @@
...
+
+class CleanPlaceGeometriesAndTaxonRanges < ActiveRecord::Migration[6.1]
+  def up
+    add_column :place_geometries, :cleaned_geom, :geometry
+    add_column :taxon_ranges, :cleaned_geom, :geometry
+
+    [:place_geometries, :taxon_ranges].each do | table_name |
+      # update lngs of geoms where all lngs are out of bounds
+      execute <<-SQL
+        UPDATE #{table_name}
```
caseyhofford, line 11: We might want to batch these changes since they could run quite slow in production.

caseyhofford (changes requested):
Looking good! Thanks for making this happen. I have a few comments, the only blockers are:
- ST_Intersection will trim polygons to only one side of the antimeridian.
- I think we should maintain the old geometry in a backup column temporarily in case we encounter issues with the translated geometries.

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
pleary: All place and taxon range geometries _should_ be compatible with the 4326 SRID projection, but in the DB right now all their SRID's are 0. We could update them at some point, but I'd like to separate that action from updating the coordinates themselves. This PR is just for updating the coordinates and I'd like to leave all SRIDs the same. If we want to make a future PR to update SRIDs, I'm OK with that. We'll also need to ensure all new and updated geometries use the same

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
pleary: Good looking out for that. I did test this on staging and it only took a few seconds, so I think we can avoid overly complicating it. There are only a couple hundred affected geometries

spec/models/taxon_range_spec.rb
pleary: The setup guide already does say to you may want to run `brew install gdal`. It says it's "optional" which might be worth changing that wording

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
pleary: There are only a couple hundred affected geometries, and I checked a lot of them. The general problem seems to be one point is a couple meters over the antimeridian, not that there's a huge portion of valuable range information over the antimeridian. So I'm OK with the trimming as it stands. If we had an easy way to split reliable, I'm all ears. In my explorations, this seems like the best approach to actually get this done

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
pleary: I was planning on exporting these two tables first so we have a backup on have if we need it. Would that address this concern? We could also keep the backup in a columnin the live database for a little while, just making the tables larger for a bit, then we'll need a future migration to remove them. Either way I'll plan on exporting the data in case we find issues further down the line after we delete the "original" column

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
pleary: I checked the numbers and there are only 30 places, and 9 taxon ranges, where only some and not all coordinates cross the antimeridian. I just manually checked all of them, and with 1 exception, I don't see any problem with this approach of trimming the out of bounds coordinates. The 1 exception is a rarely used user-created place - I can volunteer to recreate that polygon after the migration is done. Given this small subset and that they mostly are unaffected, my inclination is to proceed with the migration as written since it appears to work well and reliably. But if you have a proposal for a drop-in replacement that includes splitting that will work as reliably, I'm up for evaluating it

spec/models/taxon_range_spec.rb
pleary: I updated this spec, and the version in place_geometry I was replicating

spec/models/taxon_range_spec.rb
pleary: I added this spec for taxon ranges and place geometries

pleary:
Ready for another review @caseyhofford. I didn't yet update the migration to preserve the original geom in separate column as I wanted to get your feedback on wether or not exporting the data to archival purposes first addresses that concern. If not, I'll update the migration to preserve the original in a new column, and make an issue for removing that column sometime in the future

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
caseyhofford: OK, I think this should be fine since we are using `SRID=0` for both `ST_MakeEnvelope` and `geom`. I don't know if updating the SRIDs is high priority but just wanted to make sure we were taking this into consideration here, thanks for the info!

spec/models/taxon_range_spec.rb
caseyhofford: Oops, I missed that part my bad, no need to change this.

spec/models/taxon_range_spec.rb
caseyhofford: Thanks!

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
caseyhofford: Thanks for verifying all these geometries. Yes, splitting them is probably not worthwhile if they are all just extending a few meters over the antimeridian, I was more concerned about large polygons spanning both sides of the antimeridian. Thanks for finding the exception, I can also help with recreating that one.

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
caseyhofford: Yes, this works as well, and if the increase in table sizes is very significant than I think that is a good reason to just stick with the backup. The main advantage I could see for creating an `original` column is that we would be able to use `db:rollback` to revert the change quickly in case we needed to rollback the deploy for some reason. This isn't blocking but just thought I'd bring up that option.

spec/models/taxon_range_spec.rb
caseyhofford: Thank you!

db/migrate/20260511210120_clean_place_geometries_and_taxon_ranges.rb
caseyhofford: Great! I wasn't sure how many geometries this was affecting, but those numbers don't sound concerning.

caseyhofford (approved):
Looks good! Thanks for addressing the feedback. Exporting the data does address the concern, and my only reason for leaning towards the extra column is we could support reverting with a `db:rollback` rather than having to do a manual recovery of the table, but both seem like fine options to me.

PR #4903, opened 2026-05-13
feat: change project edit date and time form fields, WEB-834

pleary:
This PR updates how project date-related fields are edited. Some of the complication in code is a result of splitting up date from time into two separate form fields, so there is no longer a 1-1 mapping between form field and project rule.

There's an additional complication that the "Exact" date type can either result in an `observed_on` attribute, or a `d1` and `d2` attribute, depending on if a time range is included or not. Similarly, when loading a project, a project with `d1` and `d2` attributes may need to use the "Exact" or "Range" date type, depending on if they include times, and, depending on the viewers time zone, if those date/times are the same calendar day or not.

Various new validation is added as well, particularly checking to make sure end date/times are later than start/date times which is something we hadn't been doing before

### DIFF
--- a/app/webpack/projects/form/components/regular_form.jsx
+++ b/app/webpack/projects/form/components/regular_form.jsx
@@ -13,6 +13,9 @@ import TaxonSelector from "./taxon_selector";
 import PlaceSelector from "./place_selector";
 import UserSelector from "./user_selector";
 
+const MonthNames = ( "january february march april may june july august "
+  + "september october november december" ).split( " " );
+
 class RegularForm extends React.Component {
   constructor( props ) {
     super( props );
@@ -136,27 +139,25 @@ class RegularForm extends React.Component {
     const { project, updateProject } = this.props;
     const key = `time-checkbox-${_.kebabCase( fieldName )}`;
     return (
-      <div>
-        <div className="checkbox">
-          <label>
-            <input
-              type="checkbox"
-              key={key}
-              id={key}
-              defaultChecked={project[fieldName]}
-              onChange={e => {
-                const updates = {
-                  [fieldName]: e.target.checked || null
-                };
-                if ( !e.target.checked ) {
-                  Object.assign( updates, resetUpdates );
-                }
-                updateProject( updates );
-              }}
-            />
-            { I18n.t( "views.projects.new.date_pickers.include_times" ) }
-          </label>
-        </div>
+      <div className="checkbox">
+        <label>
+          <input
+            type="checkbox"
+            key={key}
+            id={key}
+            defaultChecked={project[fieldName]}
+            onChange={e => {
+              const updates = {
+                [fieldName]: e.target.checked || null
+              };
+              if ( !e.target.checked ) {
+                Object.assign( updates, resetUpdates );
+              }
+              updateProject( updates );
+            }}
+          />
+          { I18n.t( "views.projects.new.date_pickers.include_times" ) }
+        </label>
       </div>
     );
   }
@@ -264,8 +265,6 @@ class RegularForm extends React.Component {
       project,
       setRulePreference
     } = this.props;
-    const monthNames = ( "january february march april may june july august "
-      + "september october november december" ).split( " " );
     return (
       <div
         style={{ position: "relative" }}
@@ -277,7 +276,7 @@ class RegularForm extends React.Component {
           defaultValue={project.rule_month ? project.rule_month.split( "," ) : []}
           noneSelectedText={I18n.t( "views.projects.new.date_pickers.select_months" )}
           data={
-            _.map( monthNames, ( month, i ) => (
+            _.map( MonthNames, ( month, i ) => (
               {
                 value: i + 1,
                 label: I18n.t( `date_format.month.${month}` )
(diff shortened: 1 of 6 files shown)


app/webpack/projects/form/components/regular_form.jsx
```diff
@@ -28,6 +33,279 @@ class RegularForm extends React.Component {
...
+    );
+  }
+
+  monthDatePickers( ) {
+    const {
+      project,
+      setRulePreference
+    } = this.props;
+    const monthNames = ( "january february march april may june july august "
+      + "september october november december" ).split( " " );
```
sundisc: Super minor, but would be better at the top of the file as a constant so it's not recalculated every render.

app/webpack/projects/form/components/regular_form.jsx
```diff
@@ -28,6 +33,279 @@ class RegularForm extends React.Component {
...
+        { this.errorOverlay( fieldName, ref.current, options.errorKeyAlt ) }
+      </>
+    );
+  }
+
+  includeTimesCheckbox( fieldName, resetUpdates = { } ) {
+    const { project, updateProject } = this.props;
+    const key = `time-checkbox-${_.kebabCase( fieldName )}`;
+    return (
+      <div>
```
sundisc: Looks like this `<div>` could be removed.

app/webpack/projects/shared/models/project.js
```diff
@@ -44,13 +44,14 @@ const Project = class Project extends inatjs.Project {
       this.search_params.collection_preview = true;
     }
     Object.assign( this.search_params, additionalSearchParams );
-    this.setPreviewSearchParams( );
     const start = this.rule_observed_on || this.rule_d1;
     const end = this.rule_observed_on || this.rule_d2;
+    this.ruleD1Moment = util.momentDateFromString( this.rule_d1 );
+    this.ruleD2Moment = util.momentDateFromString( this.rule_d2 );
```
sundisc: Are these ruleD*Moment vars ever used?

app/webpack/projects/shared/models/project.js
```diff
@@ -73,49 +74,237 @@ const Project = class Project extends inatjs.Project {
...
+    if ( this.exactDateD1 && this.exactDateD2
+      && !this.errors.exactDateD1 && !this.errors.exactDateD2
+    ) {
+      if ( this.exactDateD2Moment.isSameOrBefore( this.exactDateD1Moment ) ) {
+        this.errors.exactDateD2 = I18n.t( "views.projects.new.end_time_must_be_after_start_time" );
+        return;
+      }
+      // if both times are valid, set tentative date range rules
+      this.rule_d1 = this.exactDateD1Moment.format( "YYYY-MM-DD HH:mm UTCZ" );
+      this.rule_d2 = this.exactDateD2Moment.format( "YYYY-MM-DD HH:mm UTCZ" );
```
sundisc: Probably not a problem, but `validateRangeDates` sets these to `YYYY-MM-DD HH:mm Z` instead of `YYYY-MM-DD HH:mm UTCZ` (which I think are the same?)

app/webpack/projects/shared/models/project.js
```diff
@@ -227,7 +416,15 @@ const Project = class Project extends inatjs.Project {
...
-        _.map( _.filter( this.rule_preferences, p => p.value !== null ), p => [p.field, p.value] )
+        _.map(
+          _.filter( this.rule_preferences, p => {
+            if ( _.includes( ["d1", "d2", "observed_on", "month"], p.field ) ) {
+              return false;
+            }
+            return p.value !== null;
+          } ),
+          p => [p.field, p.value]
+        )
```
sundisc, line 425: I sort of prefer switching to the native JS versions of `map` and `filter` if possible, or just `reduce` to do it all in one pass. Could see an argument in keeping the lodash functions to keep the file consistent though...

sundisc (commented):
Looks good! Much cleaner in the `regular_form` after breaking the parts into methods. Have some minor comments that aren't blocking.

app/webpack/projects/form/components/regular_form.jsx
pleary: Thanks - I moved this constant to the top

app/webpack/projects/form/components/regular_form.jsx
pleary: I've removed it

app/webpack/projects/shared/models/project.js
pleary: Ah indeed I added this early on and they eventually become unused. I've removed them

app/webpack/projects/shared/models/project.js
pleary: Thanks this is a good catch. Things do work with this format, but it's not ideal. The `UTC` component of the format is for form display purposes only as requested in the issue, but ideally removed when submitting to project update, which is the purpose of the `rule_*` attributes. I've removed the UTC bit

app/webpack/projects/shared/models/project.js
pleary: Noted. In this case I've decided to stick with lodash since we have several other instances of _.map and _.filter in the project edit code, and none of native map and filter. I hear the desire to stick with native, and will try and keep that in mind for any new code. Maybe at some point we do a pass and update existing files, attempting to be consistent within files as much as is reasonable

pleary:
Thanks for the review comments. I pushed another commit with some updates. Aside from map and filter, hopefully I've addressed everything

sundisc (approved):
Looks good, thanks!

PR #4904, opened 2026-05-13
Update issue forms and guidance, LEAD-26

danrademacher:
Working on improving guidance for OS contributors, esp on filing issues. NOTE: Rendering of these files is impossible to test locally. We'll see how it looks in the PR.

### DIFF
--- a/.github/ISSUE_TEMPLATE/config.yml
+++ b/.github/ISSUE_TEMPLATE/config.yml
@@ -1,0 +1,27 @@
+# Place this file at .github/ISSUE_TEMPLATE/config.yml so that Github 
+# will present these links to users and direct traffic to the forum.
+
+contact_links:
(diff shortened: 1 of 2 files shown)


danrademacher:
It's not really possible to test these changes locally. So I committed them to a personal repo on GH pages to confirm they work. Here's the evidence:
image

So any GH user can still make an issue, but we put a lot more guardrails to encourage them to use the forum.

Not shown in the image -- I changed the "Technical Bug Report" title to "Report a code-level technical issue" -- to prevent confusion between those items.

As far as I can tell in online searches ([for example](https://github.com/orgs/community/discussions/21601)), there is no method to force the `contact_links` to show up before the issue template (which would be the ideal case here)

pleary (approved):
Thanks!

PR #4905, opened 2026-05-14
web 1001 improve announcement selection logic

caseyhofford:
Adds a parent_announcement_id association to announcements.

Announcement.active used a global cascading locale fallback that caused locale-specific announcements in one family to suppress completely unrelated global announcements. Dismissing locale specific announcements also led to users seeing the global announcement.                                 
                                                                                      
Solution: Adds a parent_announcement_id column to group translation/targeting variants into single-level families. Each family is filtered independently using a tiered "prefer specific, fall back to global" pattern across three filters:

1. IP country — prefer geo-targeted variants, fall back to global                   
2. Site — prefer site-specific variants, fall back to global                          
3. Locale — prefer exact locale → language-only → global          
                                                                                        
A general pick_best_matches(group, *tiers) helper drives all three, trying each tier's predicate in order and returning the first non-empty result. Dismissing any announcement in the family hides the whole family.

Big diff, but half of it is specs to ensure I got the logic right.


spec/models/announcement_spec.rb
```diff
@@ -527,6 +528,226 @@
...
+    end
+
+    it "shows all locales when no locales set" do
+      a = create :announcement, body: "Simple body"
+      expect( a.dropdown_label ).to include( "(all locales)" )
+    end
+  end
+
+  describe "duplicate_as_user" do
+    it "links duplicate as child of original" do
```
caseyhofford: We could change this to only auto-link records when added through the "Create Targeted Variant" button while preserving the original behavior for the "Duplicate" button. I am not sure how often "Duplicate" is being used for cases other than targeting.

app/controllers/announcements_controller.rb
```diff
@@ -138,6 +146,16 @@ def load_oauth_applications
     @oauth_applications.sort_by!( &:name )
   end
 
+  def load_parent_announcement_options
```
caseyhofford, line 153: Used in the create/edit announcements form to select a parent. This does not filter to site ID for site admins, because I believe we want site admins to be able to choose a global announcement to inherit from.

app/models/announcement.rb
```diff
@@ -338,9 +373,8 @@ def self.active( options = {} )
     site = options[:site]
     user = options[:user]
     scope = Announcement.
+      includes( :sites ).
       where( '? BETWEEN "start" AND "end"', Time.now.utc ).
-      joins( "LEFT OUTER JOIN announcements_sites ON announcements_sites.announcement_id = announcements.id" ).
```
caseyhofford, line 343: We can include sites in the query rather than joining to use ruby for the site filtering at line 441.

app/models/announcement.rb
```diff
@@ -357,76 +391,77 @@ def self.active( options = {} )
       scope = scope.where( "? = ANY( clients ) OR clients = '{}'", options[:user_agent_client] )
     end
 
-    # Site filtering
-    scope = if site
```
caseyhofford, line 362: Site filtering is now done within the [`filter_family` method at line 441](https://github.com/inaturalist/inaturalist/pull/4905/changes#diff-dafad8f7f24127720f042c933df7c00fa4455f7ab1264df8da009069de486be7R441) which allows families of announcements to use a site targeted announcement if there is a match, and fallback to the global announcement if not.

app/models/announcement.rb
```diff
@@ -357,76 +391,77 @@ def self.active( options = {} )
...
+    all_announcements = scope.to_a
+    locale_str = I18n.locale.to_s
+    lang_only = locale_str.split( "-" ).first
 
-    # Locale- and site- specific targeting above may have excluded
-    # announcements without those filters, so we need to add them back in and
-    # check their targeting. This is kind of a pointless extra db query. We
-    # might just want to do locale and site filtering in Ruby instead of the
-    # DB for simplicity.
-    if announcements.blank?
```
caseyhofford, line 395: This is no longer needed with the announcement families approach.

app/models/announcement.rb
```diff
@@ -357,76 +391,77 @@ def self.active( options = {} )
...
-      announcements = base_scope.where( "(locales IS NULL OR locales = '{}') AND sites.id IS NULL" ).select do | a |
-        a.targeted_to_user?( user ) && !a.dismissed_by?( user )
-      end
+    geoip_country = if options[:ip]
```
caseyhofford, line 430: This ensures `geoip_country` is defined, and either a country or nil when checked on line 432.

app/models/announcement.rb
```diff
@@ -357,76 +391,77 @@ def self.active( options = {} )
...
     announcements.sort_by do | a |
       [
-        a.locales.include?( I18n.locale ) ? 0 : 1,
+        a.locales.include?( I18n.locale.to_s ) ? 0 : 1,
         a.id * -1
       ]
     end
   end
 
+  def self.filter_family( group, user:, site:, ip:, geoip_country:, locale_str:, lang_only: )
```
caseyhofford, line 455: I pulled these filters out into a separate method so that it is clear what filters are applied on each family. These filters are dismissal (which dismisses the whole family), user targeting, ip, site and locale.

app/models/announcement.rb
```diff
@@ -357,76 +391,77 @@ def self.active( options = {} )
...
-    announcement_target_site = announcements.detect do | annc |
-      annc.site_ids.present? && annc.excludes_non_site
-    end
-    if announcement_target_site
-      announcements = announcements.select do | annc |
-        annc.site_ids.present?
-      end
+    # TODO: remove excludes_non_site logic once the column is fully deprecated.
+    # This was a workaround now replaced by parent_announcement_id family grouping.
+    if announcements.detect {| a | a.site_ids.present? && a.excludes_non_site }
```
caseyhofford, line 443: The logic here is the same as the previous block. I honestly think we should probably just remove this check since it was a workaround for the same issue, which is handled better with the announcement families. I am only leaving it here in case there are old announcements that expect this to still work.

app/models/announcement.rb
```diff
@@ -357,76 +391,77 @@ def self.active( options = {} )
...
+    end
+
+    pick_best_matches( filtered,
+      ->( a ) { a.locales&.include?( locale_str ) },
+      ->( a ) { locale_str != lang_only && a.locales&.include?( lang_only ) },
+      ->( a ) { a.locales.blank? } )
+  end
+  private_class_method :filter_family
+
+  def self.pick_best_matches( group, *tiers )
```
caseyhofford, line 489: This supports fallback logic for any filter, and would support any number of matching rules, where we try the next filter if the previous filter returned no results.

app/views/announcements/_form.html.haml
```diff
@@ -45,14 +45,15 @@
       = f.check_box :dismissible, label_after: true
       = f.check_box :prefers_target_staff, label_after: true
       = f.check_box :target_creator, label_after: true
+      - has_children = @announcement.persisted? && @announcement.child_announcements.exists?
+      = f.select :parent_announcement_id, @parent_announcement_options, { include_blank: t( :none ) }, { disabled: has_children, description: has_children ? t( "views.announcements.parent_announcement_disabled_desc" ) : t( "views.announcements.parent_announcement_desc" ), description_tip: true }
```
caseyhofford, line 50: The select dropdown is disabled with a new tooltip added explaining why if the announcement already has children.

app/views/announcements/_form.html.haml
```diff
@@ -45,14 +45,15 @@
...
     .col-xs-4
       = f.select :ip_countries, country_options, { include_blank: t( :all ) }, multiple: true, description: t( "views.announcements.ip_countries_desc" ), description_tip: true
       - if current_user.is_admin?
         .admin
           = f.select :site_ids, @sites.map{ |s| [s.name, s.id] }, { include_blank: t(:all) }, multiple: true, description: t( "views.announcements.site_ids_desc" ), description_tip: true
       - else
         = f.hidden_field :site_ids
-      = f.check_box :excludes_non_site, description: t( "views.announcements.excludes_non_site_desc" ), description_tip: true
```
caseyhofford, line 56: I think we should prevent people from using this setting now since it was a workaround for the same feature. This previously caused all non site specific announcements to be removed if a site specific announcement with this setting was present.

app/views/announcements/show.html.haml
```diff
@@ -85,10 +89,17 @@
                   %th{ title: attr_name }=attr_name
                   %td
                     - val = @announcement.send( a )
-                    - if a == "ip_countries" || a == "exclude_ip_countries"
-                      = val.map { | code | Place::COUNTRIES_BY_NAME.values.detect { | country | country[:code] == code }.try(:[], :name) }.sort.join( ", " )
+                    - if a == "ip_countries"
```
caseyhofford, line 93: This is a kind of confusing way that we render the attributes of announcements on show. I am matching this pattern to display the added `parent_announcement_id` attribute, as well as `exclude_ip_countries` which was not being converted to a readable string previously.

app/controllers/announcements_controller.rb
```diff
@@ -138,6 +146,16 @@ def load_oauth_applications
...
+    candidates = Announcement.
+      where( parent_announcement_id: nil ).
+      where( '"end" > ?', 30.days.ago ).
+      order( id: :desc ).
+      limit( 100 )
+    candidates = candidates.where.not( id: @announcement.id ) if @announcement&.persisted?
```
pleary, line 159: Low chance of being truly relevant, but we probably want to inject the target announcement's parent announcement, if it's for whatever reason not in the last 30 days. For example if there's an older announcement set to have a parent that is another older announcement, and you visit the child announcement edit page, the parent dropdown will not include the parent's child, since it won't be returned from this query. Rather, the dropdown will default to "None". If you then save the child, it's parent will be reset to nil.

There's another case where - right now you're offering a button to "Create a targeted variant" on an announcement's details page. When clicking that button on a parent announcement, the parent is not pre-populated on the new announcement edit page. That's mostly because of the "not" condition added here, and that announcement will refer to the original, which is persisted. Given the text of the button, it might make sense to pre-populate the parent for "variants". Perhaps that warrants another parameter sent to the "duplicate" endpoint to differentiate creating a duplicate announcement (where parent is nil) from creating a "targeted variant" (where parent is the original)

app/controllers/announcements_controller.rb
```diff
@@ -4,8 +4,9 @@ class AnnouncementsController < ApplicationController
   before_action :authenticate_user!, except: [:active]
   before_action :site_admin_required, except: [:active, :dismiss]
   before_action :load_announcement, only: [:show, :edit, :update, :destroy, :dismiss, :duplicate]
-  before_action :load_sites, only: [:new, :edit, :create, :duplicate]
-  before_action :load_oauth_applications, only: [:new, :edit, :create, :duplicate]
+  before_action :load_sites, only: [:new, :edit, :create, :update, :duplicate]
+  before_action :load_oauth_applications, only: [:new, :edit, :create, :update, :duplicate]
+  before_action :load_parent_announcement_options, only: [:new, :edit, :create, :update, :duplicate]
```
pleary, line 9: Might be good to have a rails-level validation that a parent exists, e.g. `validates_presence_of :parent_announcement, if: :parent_announcement_id?`. The foreign key constraint added with the migration will enforce the parent exists, but relying on that means the code could would hit an uncaught exception, not an invalid ActiveRecord instance. I'd say for any foreign key constraint in the DB, there should be a corresponding model validation for the same

db/migrate/20260507201412_add_parent_announcement_id_to_announcements.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddParentAnnouncementIdToAnnouncements < ActiveRecord::Migration[6.1]
+  def change
+    add_column :announcements, :parent_announcement_id, :integer
+    add_index :announcements, :parent_announcement_id
+    add_foreign_key :announcements, :announcements, column: :parent_announcement_id, on_delete: :nullify
```
pleary, line 7: No problem to add this - we probably should be using more foreign key constraints. I'll just mention that we currently don't use any foreign key constraints at all in the DB. We are currently relying entirely on model validations and association `dependent` actions for cleanup. For smaller tables and properly indexed columns like here with Announcements, this shouldn't raise any performance problems. There could possibly be performance problems with huge tables and improperly indexed columns, but we'd probably also hit those same performance issues with equivalent model level validations - they'd just be doubled up when the DB also checks constraints

spec/models/announcement_spec.rb
pleary: I guess this is similar to what I mentioned in another comment - we may need to differentiate the "duplicate" from "create targeted variant" action and how that affects populating the parent announcement ID

pleary (changes requested):
Thanks - this is looking great. I like showing the grouped announcement on the details page, and the ability to easily create another variant. I have a few comments here worth looking into. Generally looking very good though. I'd recommend we wait until next week to merge/deploy this in case anything here might affect announcements we plan to display over the weekend

app/controllers/announcements_controller.rb
caseyhofford: Good catch, fixed both of these in 183e4d3f3b4d9f9120230718dba45a254dafcd5e by injecting the parent in the parent_announcement_options and adding a variant flag to the duplicate method.

app/controllers/announcements_controller.rb
caseyhofford: Good call, I've added this in 1ccbaad92a004cdb821bc2b6c35ea8caee7c45ac

db/migrate/20260507201412_add_parent_announcement_id_to_announcements.rb
caseyhofford: Makes sense, if you think the announcements table will not hit these performance issues I would vote to leave it, especially because the `on_delete: :nullify` will help ensure we don't get invalid records if a parent is deleted. We could also handle this in the `destroy` route if that's the preferred style.

app/controllers/announcements_controller.rb
```diff
@@ -111,9 +119,13 @@ def dismiss
...
   def duplicate
-    @announcement = @announcement.duplicate_as_user( current_user )
-
-    # Show the form with all fields prefilled (unsaved record)
+    original = @announcement
+    @announcement = original.duplicate_as_user( current_user )
+    if params[:variant]
+      @parent_announcement_options.unshift( [original.dropdown_label, original.id] ) unless
+        @parent_announcement_options.any? {| _label, id | id == @announcement.parent_announcement_id }
+      @announcement.parent_announcement_id = original.id
```
pleary: When creating a variant, if the original is itself a child announcement, the parent should be set to the original's parent, not the original itself. Users can get there by click on the "create a targeted variant" button on the details page of a child announcement

pleary (approved):
Looks good to me!

app/controllers/announcements_controller.rb
caseyhofford: Good catch, thanks, fixed in f8b4038c32b5d4bf6c36ae5f1842e026ca1ce96e

PR #4906, opened 2026-05-14
feat: update users email index to be unique, WEB-948

pleary:


caseyhofford (approved):
Looks good, runs locally! We need to update structure.sql for another migration.

PR #4907, opened 2026-05-14
feat: update delayed jobs unique_hash index to be unique, OPS-304

pleary:


caseyhofford (approved):
Looks good! Ran this locally and confirmed that it deduplicated the duplicate jobs that I had locally.

PR #4910, opened 2026-05-15
feat: use node 24, OPS-332

pleary:
* use Node.js 24.15.0
* update some Github actions packages, and force use of Node 24 for workflows


sylvain-morin (approved):

caseyhofford (approved):
Looks good, got this up and running locally after switching to node 24.15.0 with `nvm use 24.15.0`

PR #4913, opened 2026-05-18
Fix place match when searching in non english locale

caseyhofford:
Adds ` language: "en" ` to Google geocoder call to return result in english to match iNaturalist places API.


pleary (approved):
Looks good to me. Thanks!

PR #4915, opened 2026-05-22
fix: Add search to header menu at any size, WEB-1004

sundisc:
Search was previously not displaying in the More menu at window sizes < 992px because of the `header-md` class. This class was added by mistake, there's no benefit to using it. Removing it shows the search item whenever the More dropdown is displayed.


pleary (approved):
Confirmed this works for me

PR #4916, opened 2026-05-28
feat: add pagination headers to posts/for_user endpoint, WEB-1015

pleary:


caseyhofford (approved):
Looks great, thanks for adding those specs!

PR #4917, opened 2026-05-28
chore:web-1024: add two virtuous tags to UserVirtuousTag#POSSIBLE_TAGS

caseyhofford:
- `Do Not Contact`
- `Mid-Range Donors`

Confirmed these displays in the announcement logged in user targeting UI:
image

Did not test tags w/ actual virtuous API Keys.


pleary (approved):
Thanks!

PR #4918, opened 2026-05-28
feat: index exemplar ID user logins, reindex after update, WEB-1020

pleary:


app/models/user.rb
```diff
@@ -1551,6 +1553,27 @@ def reindex_observations_with_voted_annotations_after_destroy_later
     true
   end
 
+  def self.reindex_exemplar_identifications_after_login_change( user_id )
+    return unless ( user = User.find_by_id( user_id ) )
```
caseyhofford, line 1557: NIT: this saves a database query, and will have the same result unless `User.find_by_id( user_id )` does not exist.

```suggestion
    return unless ( user.id = user_id )
```

app/models/user.rb
```diff
@@ -1551,6 +1553,27 @@ def reindex_observations_with_voted_annotations_after_destroy_later
...
+
+    exemplar_identification_ids = user.exemplar_identifications.pluck( :id )
+    return if exemplar_identification_ids.empty?
+
+    # batch the reindexing in potentially multiple delayed jobs
```
caseyhofford, line 1562: Is this comment a TODO to update the indexing below?

caseyhofford (approved):
Looks good, one small suggestion for using id comparison rather than object comparison and a clarification about batching, but neither should block merging.

app/models/user.rb
pleary: This is actually an assignment and checking the record exists, not a comparison. Our rubocop rules allow this when the assignment is in parenthesis like it is here. But I can see how this is easily overlooked, hence our roles that disallow doing this w/o parens. Maybe I should stop doing this altogether and force the assignment to be on a separate line from the conditional. This static method accepts an integer parameter as it will most often be invoked within a delayed job, where it can be preferred to fetch the User record when the job runs

app/models/user.rb
pleary: This is mean to be a comment describe what the next block does. When delaying the indexing of many records by IDs, if there are enough IDs, multiple delayed jobs will be created so the indexing can happen in multiple threads. I guess I was just noting that I acknowledge `exemplar_identification_ids` may have many thousands of values, but that's OK because the delayed indexing will be batched

pleary:
Thanks @caseyhofford. I made one more commit to add a similar mapping type for ID bodies so we can optionally search them with partial string matches like this PR was already doing for users. I'll re-request a review based on this latest commit

caseyhofford (approved):
Looks good, matches the changes on https://github.com/inaturalist/iNaturalistAPI/pull/628

app/models/user.rb
caseyhofford: Oh sorry that makes sense, I just misread this.

PR #4919, opened 2026-05-28
feat: update app owner application text, WEB-865

pleary:


config/locales/en.yml
```diff
@@ -730,12 +730,23 @@ en:
...
+    an "app owner." This application should describe the functionality of your proposed app
+    and how you plan to use it. In order to submit an application your account must be at
+    least 2 months old and must have made at least 10 <a href="https://help.inaturalist.org/support/solutions/articles/151000170241"
+    target="_blank" rel="noopener noreferrer">improving identifications</a>
+    <i>for other users</i> in the last month. We may ask follow-up questions about your plan
+    and we reserve the right to turn down applications for any reason.
   app_owner_application_inelligible: |
     In order to submit an application your account must be at least 2 months old and must
     have made at least 10 improving identifications in the last month.
-  app_owner_application_inelligible2: |
```
caseyhofford, line 736: Just want to confirm that we actually want to replace the old key, rather than adding a new key.

caseyhofford (approved):
Looks good! Just one comment to confirm we want to remove the `app_owner_application_inelligible2` key rather than adding a new one, but that shouldn't block merging.

config/locales/en.yml
pleary: In this case, yes I think we do want to remove this. This was already a replacement for the original `app_owner_application_inelligible`. This PR removes any use of the older replacement, and leaves us with just the new replacement `app_owner_application_inelligible2_html` and the original fallback of `app_owner_application_inelligible`. Since `app_owner_application_inelligible2` isn't used at all, we should probably just remove it

PR #4921, opened 2026-05-29
Web 751 generate sitemaps for pages we want webcrawlers to visit

sylvain-morin:


app/controllers/robots_txt_controller.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+class RobotsTxtController < ActionController::Base
```
pleary: Ideally this would extend from ApplicationController like other controllers. Among other things, that will call the `set_site` method which will be helpful elsewhere in this controller

app/controllers/robots_txt_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  MAIN_SITE_ROBOTS_TEMPLATE_PATH = Rails.root.join( "public", "robots-www.txt" )
+  PARTNER_ROBOTS_TEMPLATE_PATH = Rails.root.join( "public", "robots-partners.txt" )
+
+  def robots
```
pleary, line 11: By leveraging the already looked up @site instance, this can be distilled down to e.g. 
```
    if @site == Site.default
      render plain: main_robots_content, content_type: "text/plain; charset=utf-8"
      return
    end

    render plain: partner_robots_content, content_type: "text/plain; charset=utf-8"
```

app/controllers/robots_txt_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+      return
+    end
+
+    render plain: default_robots_content, content_type: "text/plain; charset=utf-8"
+  end
+
+  private
+
+  def main_robots_content
+    site = Site.default
```
pleary: If using the above recommendation will already know @site is the default site here, so this check can be removed

app/controllers/robots_txt_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    return default_robots_content unless site
+
+    template = read_robots_template( MAIN_SITE_ROBOTS_TEMPLATE_PATH )
+    sitemap_url = "#{base_url_for_site( site )}/sitemap-www/sitemap.xml"
```
pleary: The `base_url_for_site` method could be simplified to e.g. `URI.join( @site.url, "sitemap-www/sitemap.xml" )`

app/controllers/robots_txt_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    [template, "Sitemap: #{sitemap_url}", ""].join( "\n" )
+  end
+
+  def partner_robots_content( site )
+    template = read_robots_template( PARTNER_ROBOTS_TEMPLATE_PATH )
+    sitemap_url = "#{base_url_for_site( site )}/sitemap-partners/sitemap-#{site.id}.xml"
```
pleary: If changed to `URI.join( @site.url, "sitemap-partners/sitemap-#{@site.id}.xml" )`, and above, the `base_url_for_site` can be removed

app/controllers/robots_txt_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    return "" unless File.file?( path )
+
+    File.read( path ).rstrip
+  end
+
+  def base_url_for_site( site )
+    site.url.to_s.sub( %r{/$}, "" )
+  end
+
+  def default_site_host?( host )
```
pleary: If using the recommended change to use @site, this method won't be needed

app/controllers/robots_txt_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    Site.default&.url.to_s.downcase.include?( host )
+  end
+
+  def default_robots_content
```
pleary: If using the recommended change to use @site, there will never be a fallback case where this is needed and can be removed

PR #4923, opened 2026-05-29
feat: more gracefully handle delayed job unique_hash race condition, WEB-1032

pleary:
Treat DB-level `PG::UniqueViolation` for the Delayed::Job `unique_hash` attribute as Rails `validates_uniqueness_of` validation errors, not uncaught exceptions

### DIFF
--- a/config/initializers/delayed_jobs_unique_hash.rb
+++ b/config/initializers/delayed_jobs_unique_hash.rb
@@ -23,7 +23,13 @@ module Delayed
         # before the job can be saved. Catch these unique_hash uniqueness errors
         # and add an error to the instance instead of raising an exception
         def save( *args, **kwargs, & )
-          super
+          # create a transaction, or savepoint within an open transaction, so
+          # that if a unique index violation occurs at the DB level an open
+          # transaction is not left in a fail state, causing subsequent queries
+          # to raise errors
+          self.class.transaction( requires_new: true ) do
+            super
+          end
         rescue ::ActiveRecord::RecordNotUnique, ::PG::UniqueViolation => e
           raise unless e.message.include?( "index_delayed_jobs_on_unique_hash" )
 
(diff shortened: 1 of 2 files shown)


config/initializers/delayed_jobs_unique_hash.rb
```diff
@@ -13,6 +14,27 @@ class Job < ::ActiveRecord::Base
...
+        # and add an error to the instance instead of raising an exception
+        def save( *args, **kwargs, & )
+          super
+        rescue ::ActiveRecord::RecordNotUnique, ::PG::UniqueViolation => e
+          raise unless e.message.include?( "index_delayed_jobs_on_unique_hash" )
+
+          self.unique_hash_taken = true
+          errors.add( :unique_hash, :taken )
+          false
+        end
```
sundisc, line 39: Tricky thing that AI picked up on that I wouldn't have: this might being an issue when a duplicate delayed job is started in a transaction.. Postgres will have already abandoned the transaction so the next item in the transaction would raise `PG::InFailedSqlTransaction`. So we'd still have an exception.

Maybe worth looking into! FWIW it wrote this test that fails:
```
  it "leaves the surrounding transaction usable after a rescued uniqueness failure" do
    User.delay( unique_hash: "first!" ).find( 1 )
    second_job = Delayed::Job.new( unique_hash: "first!" )
    # many jobs are enqueued from model callbacks, i.e. inside an open
    # transaction. When the INSERT violates the unique index, Postgres aborts
    # the entire transaction, so rescuing the error is not enough - we must
    # also roll back to a savepoint or every subsequent statement in the
    # transaction will raise PG::InFailedSqlTransaction
    ActiveRecord::Base.transaction do
      expect( second_job.save( validate: false ) ).to be false
      expect( second_job.unique_hash_taken ).to be true
      expect do
        Delayed::Job.count
      end.not_to raise_error
    end
  end
```

pleary: Thanks for catching that. I added in a savepoint to catch any errors during the DelayedJob save so the entire transaction doesn't fail - we can roll back only the failed job save and keep the wrapping transaction viable

sundisc (approved):
Looks good, thanks!

PR #4924, opened 2026-06-02
improve: Add TypeScript + CSS-modules build infrastructure, WEB-984

sundisc:
Foundational tooling for the [WEB-984 taxon-detail responsive work](https://linear.app/inaturalist/issue/WEB-984/initial-responsive-fixes-taxon-detail-page). Adds Typescript, a CSS module loader, some global types and type packages, and a job separate from the matrix step to run linting only once (as opposed to 14x before*).

\* There are some further optimizations possible in the `build-and-test` job. Might be worth considering building a container image for this job and reusing that instead of building the image for each matrix step.

### DIFF
--- a/.github/workflows/CI-build-test.yml
+++ b/.github/workflows/CI-build-test.yml
@@ -23,25 +23,10 @@ jobs:
     - name: Run TypeScript type check
       run: npm run typecheck
 
-  build-webpack:
-    name: Build webpack
-    runs-on: ubuntu-22.04
-    steps:
-    - uses: actions/checkout@v5
-    - run: npm install
-    - run: npm run webpack
-    - name: Upload webpack output
-      uses: actions/upload-artifact@v4
-      with:
-        name: webpack-output
-        path: app/assets/javascripts/webpack
-        retention-days: 1
-
   # Might be worth considering buildling a container image for this job and reusing that instead of
   # building the image for each matrix step.
   build-and-test:
     name: Build/Test
-    needs: build-webpack
     runs-on: ubuntu-22.04
     services:
       postgres:
@@ -106,11 +91,8 @@ jobs:
       with:
         bundler-cache: true
 
-    - name: Download webpack output
-      uses: actions/download-artifact@v4
-      with:
-        name: webpack-output
-        path: app/assets/javascripts/webpack
+    - run: npm install
+    - run: npm run webpack
 
     - name: Before script
       run: ci/scripts/before_script.sh
--- a/.github/workflows/CI-build-test.yml
+++ b/.github/workflows/CI-build-test.yml
@@ -23,8 +23,8 @@ jobs:
     - name: Run TypeScript type check
       run: npm run typecheck
 
-  # Might be worth considering building a container image for this job and reusing that instead of
-  # building the image for each matrix step.
+  # WEB-1046: Might be worth considering building a container image for this job and reusing that
+  # instead of building the image for each matrix step.
   build-and-test:
     name: Build/Test
     runs-on: ubuntu-22.04
@@ -115,6 +115,6 @@ jobs:
         TRAVIS_CI: true
         CI: true
 
-    # This is running for each matrix step (14x) when it should only be running once.
+    # WEB-1046: This is running for each matrix step (14x) when it should only be running once.
     - name: Run restricted rubocop
       run: bundle exec rubocop ./app/controllers/users ./app/controllers/admin ./app/es_indices
--- a/.github/workflows/CI-build-test.yml
+++ b/.github/workflows/CI-build-test.yml
@@ -23,8 +23,25 @@ jobs:
     - name: Run TypeScript type check
       run: npm run typecheck
 
+  build-webpack:
+    name: Build webpack
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v5
+    - run: npm install
+    - run: npm run webpack
+    - name: Upload webpack output
+      uses: actions/upload-artifact@v4
+      with:
+        name: webpack-output
+        path: app/assets/javascripts/webpack
+        retention-days: 1
+
+  # Might be worth considering buildling a container image for this job and reusing that instead of
+  # building the image for each matrix step.
   build-and-test:
     name: Build/Test
+    needs: build-webpack
     runs-on: ubuntu-22.04
     services:
       postgres:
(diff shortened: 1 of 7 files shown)


.github/workflows/CI-build-test.yml
```diff
@@ -7,6 +7,24 @@ env:
...
+    steps:
+    - uses: actions/checkout@v4
+    - run: npm install
+    - name: Run restricted eslint
+      run: |
+        npm run eslint -- \
+          --no-error-on-unmatched-pattern \
+          app/webpack/computer_vision/language \
+          "app/webpack/**/*.tsx" \
+          "app/webpack/**/*.ts"
```
caseyhofford, line 22: We will probably want to define a process for adding existing `*.js` files as they get linted so they stay linted. This is great for now, but just calling that out now for the future.

app/webpack/globals.d.ts
```diff
@@ -0,0 +1,46 @@
...
+  tab: ( command: string ) => void;
+  data: ( key: string ) => unknown;
+  get: ( index: number ) => Element | undefined;
+  taxonMap: ( options?: unknown ) => void;
+}
+interface JQueryDeparam {
+  ( str: string ): Record<string, unknown>;
+  querystring( ): Record<string, unknown>;
+}
+interface JQueryStub {
```
caseyhofford, line 29: We use a few other jQuery functions in react which we might want to include here, unless this is intentional to prevent us from using these jQuery functions in the future.

```
$.ajax
$.each
$.scrollTo
```

.github/workflows/CI-build-test.yml
```diff
@@ -7,6 +7,24 @@ env:
...
+    - name: Run TypeScript type check
+      run: npm run typecheck
+
+  # Might be worth considering building a container image for this job and reusing that instead of
```
caseyhofford: That sounds like a good improvement and would probably decrease the frequency of package installations causing CI to fail.

.github/workflows/CI-build-test.yml
```diff
@@ -97,8 +115,6 @@ jobs:
         TRAVIS_CI: true
         CI: true
 
+    # This is running for each matrix step (14x) when it should only be running once.
```
caseyhofford: Let's create a ticket for this and the above comment. I like to prefix comments like this with `TODO: ` to make them easy to find.

app/webpack/.eslintrc
```diff
@@ -67,9 +74,38 @@
...
+        "no-unused-vars": "off",
+        "@typescript-eslint/no-unused-vars": ["error", { "argsIgnorePattern": "^_" }],
+        "@typescript-eslint/no-explicit-any": "error",
+        "react/prop-types": 0
+      }
+    },
+    {
+      "files": ["*.d.ts"],
+      "rules": {
+        "@typescript-eslint/no-explicit-any": "off"
```
caseyhofford: Do we want to allow `:any` in type definitions? I'm OK with this especially since we will be interacting with a lot of JS code, but just want to make sure this is intentional. Currently `globals.d.ts` does not use `any` so it would pass without this rule.

tsconfig.json
```diff
@@ -0,0 +1,18 @@
...
+    "lib": ["ES6", "DOM"],
+    "jsx": "react",
+    "strict": true,
+    "noEmit": true,
+    "allowJs": true,
+    "checkJs": false,
+    "esModuleInterop": true,
+    "allowSyntheticDefaultImports": true,
+    "moduleResolution": "node",
+    "ignoreDeprecations": "6.0",
```
caseyhofford: Do we need to enable this `ignoreDeprecations` flag since this will break with typescript 7? I  was able to run tsc with the change below but I'm not sure if this causes issues with importing legacy node modules to typescript or something. 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html

```suggestion
    "module": "ESNext",
    "moduleResolution": "bundler",
```

caseyhofford (commented):
Looking good! Thanks for doing this. Just a few non blocking comments, the `ignoreDeprecations` flag is my biggest question.

.github/workflows/CI-build-test.yml
sundisc: Created: https://linear.app/inaturalist/issue/WEB-1046/optimize-build-and-test-step-in-ci

.github/workflows/CI-build-test.yml
sundisc: Totally.. I figure we can just add them to this list or create a text file listing them all. Ideally we'd convert the file to Typescript in addition linting!

app/webpack/.eslintrc
sundisc: Oh good call, let's remove it for now.. can always add this back if we need to.

app/webpack/globals.d.ts
sundisc: I just didn't add them because those aren't used in the files I converted to TS. Can't hurt to add them now though.

tsconfig.json
sundisc: No we don't, good call! Realizing that my boilerplate here was a little out of date so I'm revising some things.

tsconfig.json
```diff
@@ -0,0 +1,17 @@
+{
+  "compilerOptions": {
+    "target": "ESNext",
+    "lib": ["ESNext", "DOM"],
```
caseyhofford, line 4: I think we can set the `target` and `lib` to target ES6 for higher cross browser compatibility (https://compat-table.github.io/compat-table/es6/). I don't think `target` does not need to match `module` because target will determine the output of the transpiler, and `module` allows for modern modules to be imported.

[Module option docs](https://www.typescriptlang.org/tsconfig/#module)
[Target option docs](https://www.typescriptlang.org/tsconfig/#target)


```suggestion
    "target": "ES6",
    "lib": ["ES6", "DOM"],
```

app/webpack/globals.d.ts
caseyhofford: Makes sense, thanks for adding these anyways.

app/webpack/.eslintrc
caseyhofford: Thanks!

caseyhofford (commented):
Looking great! I just think we need to switch `target` back to `ES6` in `tsconfig.json`, but looks great.

tsconfig.json
sundisc: I should have mentioned this earlier, but Typescript is only used for type checking in this case, not compilation. That's handled by babel in webpack.

caseyhofford (approved):
@sundisc thanks for clarifying that babel handles the actual compilation of our TS, so we can disregard my comment about setting `target` to `ES6` in`tsconfig.json`.

PR #4925, opened 2026-06-02
Web 881 unknown taxa button sets identified=true

caseyhofford:
This fixes a UX bug where selecting an "unknown" taxa on the discover or identify pages returned viruses and bacteria because they are not "iconic" tax. Now this button will return only observations with no identifications.

### DIFF
--- a/app/assets/javascripts/ang/controllers/observation_search.js.erb
+++ b/app/assets/javascripts/ang/controllers/observation_search.js.erb
@@ -162,7 +162,8 @@ var SearchController = function (
   $scope.possibleSubviews = { observations: ["map", "grid", "table"] };
   $scope.possibleFields = ["iconic_taxa", "month", "swlat", "swlng",
     "nelat", "nelng", "place_id", "taxon_id", "page", "view", "subview",
-    "locale", "preferred_place_id", "ident_user_id", "spam", "test"];
+    "locale", "preferred_place_id", "ident_user_id", "spam", "test",
+    "identified"];
   $scope.defaultView = "observations";
   $scope.defaultSubview = "grid";
   $rootScope.mapType = "map";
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/ang/controllers/observation_search.js.erb
```diff
@@ -202,6 +203,19 @@ var SearchController = function (
   $scope.$watchGroup( ["mapType", "mapLabels", "mapTerrain"], function ( ) {
     $rootScope.$emit( "setMapType", $scope.mapType, $scope.mapLabels, $scope.mapTerrain );
   } );
+  $scope.toggleUnknownFilter = function ( ) {
+    if ( $scope.params.identified === "false" ) {
+      delete $scope.params.identified;
+    } else {
+      $scope.params.iconic_taxa_hash = {};
```
sundisc, line 210: Since this sets an empty hash, I'm no longer able to select unknown _and_ some other set of iconic taxa. Not sure if that's intended.

https://github.com/user-attachments/assets/d168531b-b644-4412-8a64-6c608eb8a629

app/webpack/observations/identify/components/filters_button.jsx
```diff
@@ -152,16 +156,27 @@ class FiltersButton extends React.Component {
...
           key={`btn-${t.name}`}
           onClick={( ) => {
-            let newIconicTaxa;
-            if ( _.includes( params.iconic_taxa, t.name ) ) {
-              newIconicTaxa = _.without( params.iconic_taxa, t.name );
+            if ( isUnknown ) {
+              if ( params.identified === false ) {
+                updateSearchParams( { identified: null } );
+              } else {
+                updateSearchParams( { iconic_taxa: [], identified: false } );
```
sundisc, line 163: Same thing here - setting `iconic_taxa` to `[]` prevents being able to select other iconic taxa along with unknown.

sundisc (commented):
Code looks good to me, just want to be sure we have the right UX around selecting iconic taxa + unknown.

app/assets/javascripts/ang/controllers/observation_search.js.erb
caseyhofford: This is because the `identified=false` combined with an iconic_taxa filter will always return no results because they are separate params so we get the result of requiring both no identification AND an iconic_taxa identification on the same observation.

app/webpack/observations/identify/components/filters_button.jsx
caseyhofford: Same as the comment above, if we select an `iconic_taxa` and `identified: false` we will get no results.

sundisc (approved):
Sounds like that behavior is what we expect - looks good to me!

pleary (approved):
Looks good to me!

PR #4926, opened 2026-06-02
Playwright e2e minimal

caseyhofford:


spec/e2e/app_commands/eval.rb
```diff
@@ -0,0 +1 @@
+Kernel.eval(command_options) unless command_options.nil?
```
caseyhofford: This is generated by the cypress_on_rails gem's generator for playwright. It allows us to execute arbitrary ruby from playwright. This is safe since it only runs in the test environment, however I am open to removing it since the only place it is currently being used is to confirm that the e2e tests are running in the test environment https://github.com/inaturalist/inaturalist/pull/4926/changes#diff-8db6429527114106effa636aed0a98dc8e9189213917c296682666bd30a02d9cR13

spec/e2e/playwright/tests/users/login.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+  } );
+} );
+
+test.describe( "Login happy path", () => {
+  let testEmail: string;
+
+  test.beforeEach( async () => {
+    await appClean();
+    const user = await appMake( "create", "user", {
+      email: `e2e_login_${Date.now()}@gmail.com`,
```
sundisc: Just curious, why do we interpolate the current date into the email?

spec/e2e/playwright/playwright.config.ts
```diff
@@ -0,0 +1,38 @@
+import path from "path";
+import { defineConfig, devices } from "@playwright/test";
+import { envConfig } from "../shared/env.config";
+
+export default defineConfig( {
+  globalSetup: "./global-setup",
+  webServer: {
+    command: "bundle exec rake inaturalist:generate_translations_js && bundle exec rails server -e test -p 3001",
```
sundisc: I got stuck for a while because I didn't have precompiled assets which caused tests to hang (specifically when creating a using, which sends a welcome email, which has assets that trigger compilation). Could be avoided by adding a precompililation step here: `rake assets:precompile`

spec/e2e/playwright/tests/observations/observations-show.spec.ts
```diff
@@ -0,0 +1,31 @@
+import { test, expect } from "@playwright/test";
+import { ObservationDetailPage } from "../../page-objects/observation-detail.page";
+import { mockObservationFetch } from "../../fixtures/observation-response";
+import { app, appClean, appMake } from "../../support/on-rails";
+
+test.describe( "Observation detail page", () => {
+  let detailPage: ObservationDetailPage;
+
+  test.beforeEach( async ( { page } ) => {
+    await appClean();
```
sundisc: Not a big deal right now, but I don't think we need this because Faker creates unique entries by default. This is currently preventing us from increasing workers > 1 because one test could wipe the DB while another test is mid-flight, causing a failure.

I guess the downside of removing would be the case where two tests are explicitly defining data the same way, which would cause duplicate errors. For example two tests creating a user with email "blah@example.com" will cause an error.

spec/e2e/app_commands/eval.rb
sundisc: Could we make an `app_command` specifically for reading the app env instead of allowing arbitrary code eval?

sundisc: Or we could add something to `config/initializers/cypress_on_rails.rb` with `c.before_request` like:
```
c.before_request = lambda do | _request |
  # Defense in depth: even if use_middleware is ever loosened (e.g. to enable
  # the bridge on a non-test host), refuse to execute any command outside test.
  next if Rails.env.test?
    
  [403, { "Content-Type" => "text/plain" }, ["E2E command endpoint is test-only"]]
end
```

spec/e2e/playwright/tests/users/login.spec.ts
sundisc: Oh now I see, you're ensuring uniqueness. You could leave email out here and Faker would generate a unique one itself.

sundisc (commented):
A few comments worth considering, but otherwise looking good! Would be nice to add a little bit of documentation somewhere.

config/application.rb
```diff
@@ -148,6 +148,11 @@ class Application < Rails::Application
         ]
       }
     end
+
+    if defined?(CypressOnRails) && Rails.env.test?
+      require 'cypress_on_rails/middleware'
+      config.middleware.use CypressOnRails::Middleware
+    end
```
sundisc: Actually not needed because `c.use_middleware = Rails.env.test?` does `app.middleware.use Middleware` automatically. Tested and seems to work without this block.

caseyhofford: Good call! Fixed in 470e5ecc1ff993fc35de6417a8f72aca6ab3d538

spec/e2e/app_commands/eval.rb
caseyhofford: Good call, I switched to using a new app command called env in 572b2a56bbad563c9fa68b8a93af2f9488a66ec1 and also added the before request ae78969660a1357e28b645aba6dab06c6bddf7fd,
Maybe this is overkill but there could still be data pollution issues without using cypress on rails if data is created through interactions with the page,

spec/e2e/playwright/playwright.config.ts
caseyhofford: Good call, I removed this from `.github/workflows/CI-e2e-tests.yml` and run it here now so that assets will be precompiled on each test run in any environment. ba2abd596b8c5717d978b905c0af6bd652837ae5

spec/e2e/playwright/tests/users/login.spec.ts
caseyhofford: Good call! Remvoed this and fixed in 8fc692fdcb91b6f05a4aaf864c74be1100d560b6

spec/e2e/playwright/tests/observations/observations-show.spec.ts
caseyhofford: Thanks this is definitely the move, we will definitely want to run e2e tests in parallel once there are more. We also might as well take advantage of the features of Faker! I removed both uses of `appClean()` in 4acf45d956c5398140f1f82fa8103d3f787cfcfa then decided to fully remove `app_commands/clean.rb` in 396673f5690721991cf63946f38e361f31f206fd

caseyhofford (commented):
Thanks for the review @sundisc, made those changes and also replaced the `tsconfig.json` with yours from #4924

spec/e2e/playwright/tests/observations/observations-show.spec.ts
```diff
@@ -0,0 +1,30 @@
+import { test, expect } from "@playwright/test";
+import { ObservationDetailPage } from "../../page-objects/observation-detail.page";
+import { mockObservationFetch } from "../../fixtures/observation-response";
+import { app, appClean, appMake } from "../../support/on-rails";
```
sundisc: Looks like `appClean` is still hanging around in a couple of imports. Should get picked up when you merge in main with the new TS branch linting.

sundisc (approved):
Just a couple of dead imports but otherwise looks good.. thanks!

spec/e2e/playwright/tests/observations/observations-show.spec.ts
caseyhofford: Thanks! Fixed in 053b6b983a87fa98694fb25e84080a950df93db7

PR #4928, opened 2026-06-08
fix: Fix header overflow on smaller screens, WEB-1026

sundisc:
https://linear.app/inaturalist/issue/WEB-1026/fix-header-overflow-on-smaller-screens

Hides header labels to prevent overflow. Extracts CSS that handled this for mobile size previously and made it a reusable mixin, which is used in both the tablet and mobile cases.

Added header-specific tests as well as a general horizontal overflow test that can be reused on any page.

**ALSO** allow E2E tests to run on as many works as it decides.

https://github.com/user-attachments/assets/37970b7a-a977-4d6e-8b44-d901faead84d

### DIFF
--- a/app/assets/stylesheets/index.scss
+++ b/app/assets/stylesheets/index.scss
@@ -2527,6 +2527,9 @@ body:not(.responsive) {
 }
 
 @mixin header-narrow-menu {
+  .search {
+    flex: 1;
+  }
   #narrow-menu {
     display: flex;
     align-items: center;
--- a/app/assets/stylesheets/index.scss
+++ b/app/assets/stylesheets/index.scss
@@ -2216,6 +2216,7 @@ $header-line-height: 34px;
   .search {
     margin-bottom: 0;
     margin-right: 10px;
+    min-width: 0;
     display: -webkit-box;
     display: -webkit-flex;
     display: -moz-box;
@@ -2569,7 +2570,7 @@ body:not(.responsive) {
   }
 }
 
-@media (min-width: $md) and (max-width: $xl-max) {
+@media (max-width: $xl-max) {
   body.responsive #header {
     @include header-narrow-menu;
     #mainnav {
@@ -2586,8 +2587,7 @@ body:not(.responsive) {
       margin-right: 15px;
     }
     #header {
-      @include header-narrow-menu;
-      #headersearch, #mainnav {
+      #headersearch {
         display: none;
       }
       #messagesnav, #updatesnav {
--- a/app/assets/stylesheets/index.scss
+++ b/app/assets/stylesheets/index.scss
@@ -2144,21 +2144,10 @@ $header-line-height: 34px;
       }
     }
   }
-  @media (min-width: 992px) and (max-width: 1199.99px) {
+  @media (min-width: $lg) and (max-width: $xl-max) {
     .header-md { display: inline-block; }
     .dropdown-menu {
       .header-md { display: block; }
-    }
-    &.search-open {
-      .header-md { display: none; }
-      .header-xs { display: inline-block; }
-      .dropdown-menu {
-        .header-xs { display: block; }
-      }
-      .btn-focus {
-        font-size: 15px !important;
-        padding: 6px 7px !important;
-      }
     }
   }
   @media (min-width: 1200px) {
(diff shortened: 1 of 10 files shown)


app/assets/stylesheets/index.scss
```diff
@@ -2537,6 +2526,58 @@ body:not(.responsive) {
...
+        }
+      }
+    }
+  }
+}
+
+@media (min-width: $md) and (max-width: $xl-max) {
+  body.responsive #header {
+    @include header-narrow-menu;
+    #mainnav {
```
caseyhofford, line 2576: Since we are using the same mixin for both this and the media query below both include
```
  body.responsive #header {
    @include header-narrow-menu;
    #mainnav {
      display: none;
    }
  }
```
would it make sense to change this media query to `@media (max-width: $xl-max) {...` and then remove the matching css from the `@media (max-width: $mobile-max) {...` below?

spec/e2e/playwright/tests/shared/header.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+      expect(
+          mainnav,
+          `"${label}" from #narrow-menu hamburger menu should also appear in the #mainnav`
+        ).toContain( label );
+      }
+    }
+  } );
+} );
+
+expectNoHorizontalOverflow( "/" );
```
caseyhofford: Maybe move this to the top or bottom of the test suite so it doesn't get too hidden between these other test blocks.

spec/e2e/playwright/tests/shared/header.spec.ts
```diff
@@ -0,0 +1,144 @@
...
+    } ).filter( Boolean );
+
+    return { mainnav, hamburger };
+  } );
+}
+
+test.beforeAll( async () => {
+  const user = await appMake( "create", "user", { password: TEST_PASSWORD } );
+  testEmail = user.email as string;
+  await app( "add_test_group", { user_id: user.id, test_groups: "responsive-header" } );
```
caseyhofford, line 36: Awesome, thanks for adding this helper!

config/environments/test.rb
```diff
@@ -7,7 +7,16 @@
   # and recreated between test runs.  Don't rely on the data there!
   config.cache_classes = true
 
-  config.eager_load = false
+  # Normally the test env autoloads lazily (false) for fast RSpec boots. But the
```
caseyhofford, line 10: Good catch, thanks for fixing this.

caseyhofford (approved):
Looks great! Just a couple small comments but nothing blocking. Thanks for adding specs 😄

app/assets/stylesheets/index.scss
sundisc: Oh good call.. added it to https://github.com/inaturalist/inaturalist/pull/4928/commits/742d373385c2fe02a7f681ee0fc9cbecd07a17e7

spec/e2e/playwright/tests/shared/header.spec.ts
sundisc: For sure: https://github.com/inaturalist/inaturalist/pull/4928/commits/e6c4b5391ce4536f4d863867aab75c4ebcf8727c

app/assets/stylesheets/index.scss
caseyhofford: Thanks!

caseyhofford (approved):
Looks great, thank you!

pleary (approved):
This looks good to me. Just want to make sure we're aware there's a couple things in the logged-in full-size header that don't appear in the narrow header menu - favorite projects under Community, and a link to Donate under More. We also lose the message and notification counts on even narrower displays. The latter isn't new, but the former two would have shown until a narrower breakpoint. Not a blocker, just mentioning it so we're aware as we plan to roll out responsive modes to more common pages soon.

sundisc:
@pleary Ah yes, good call. Added the donate link. The projects part has a bit more complexity around it so created a separate ticket for that it. https://linear.app/inaturalist/issue/WEB-1053/ensure-parity-between-headers-mobile-and-desktop-links

I don't think I understand your second point about message and notification counts - I still see them at very narrow screens.

Screenshot 2026-06-11 at 4 07 40 PM

Going to go ahead and merge now though since you said that's a preexisting issue.

PR #4929, opened 2026-06-08
web 1014 timed suspension feedback

caseyhofford:
- **fix(WEB-1050): update max length for suspension reason**
- **fix(WEB-1049): update suspension reason strings**
- **fix(WEB-1048): update user reinstatement email copy**
- **fix: send suspension reason in reinstatement emails for timed suspensions only**
- **feat(WEB-1047): add suspended email**
- **move localization into email templates**
- **fixup! fix: send suspension reason in reinstatement emails for timed suspensions only**

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -5,11 +5,11 @@
 module ApplicationHelper
   include Ambidextrous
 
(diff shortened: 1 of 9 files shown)


pleary (approved):
Looks good to me

PR #4931, opened 2026-06-09
fix(WEB-1021): set tabIndex=-1 for nominate checkbox on ID modal

caseyhofford:
Prevents tab navigating through the add identification modal from selecting the `Nominate as an ID Tip (diagnostic characteristics that teach others to identify this organism)` checkbox.

Test this out at `/observations/identify`

image

### DIFF
--- a/app/webpack/observations/identify/components/identification_form.jsx
+++ b/app/webpack/observations/identify/components/identification_form.jsx
@@ -98,6 +98,7 @@ class IdentificationForm extends React.Component {
               <input
                 type="checkbox"
                 id="nominate-id"
+                tabIndex={-1}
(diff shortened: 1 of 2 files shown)


pleary (changes requested):
Looking good. There's one other place we use this checkbox and that's on the observation details page in `activity_create_panel.jsx`. It would be good to add this there for consistency. With that, OK with me to merge

caseyhofford:
Good catch patrick, handled the observations show page checkbox as well in 075fc35f715c942bf83e4b91dee46bedd4ce241b

pleary (approved):
Thanks!

PR #4934, opened 2026-06-10
feat: Implement React unit testing, WEB-946

sundisc:
- Adds Jest along with other packages and setup which will allow us to test rendered components' behavior.
- Adds tests for taxa's `util.js`, which uncovered an issue around the usage of `.has( )`, which is now fixed.
  - This was actually causing a bug where visiting a taxon page with an unsupported tab anchor, such as https://www.inaturalist.org/taxa/55556-Oncopeltus-fasciatus#highlights-tab, would present the user with an empty tab area.
- Moves babel config to `babel.config.js` to prevent repeating ourselves in both Webpack and Jest setups
- Adds unit testing step to CI


pleary (approved):
Thanks - this will be very useful

PR #4935, opened 2026-06-10
feat: multilingual URLs and hreflang for taxon pages WEB-1052

sylvain-morin:


lib/sitemap/sitemap_www_sitemap_generator.rb
```diff
@@ -101,6 +105,26 @@ def generate_taxa( dir )
       end
     end
 
+    def generate_taxa_for_locale( dir, locale )
```
pleary, line 108: The [Google sitemap docs](https://developers.google.com/search/docs/specialty/international/localized-versions#example_2) show something a little more elaborate for including localized variant URLs in a sitemap. They recommend the sitemap be setup basically like the `<link` tags - that each variant has all other variants listed in a `<xhtml:link`. From what I can see if we do that, the `x-default` version should probably also be included. I don't know what downsides are to not doing it - like search engines not understanding the localized URL variants are all connected until they perform a crawl when they hopefully make that connection from the `<link` tags

app/controllers/taxa_controller.rb
```diff
@@ -181,6 +181,14 @@ def show
 
     return render_404 unless @taxon
 
+    # Locale-prefixed URL (e.g. /fr/taxa/...) — return 404 if the taxon has no valid
+    # common name in the requested locale. Keeps locale URLs meaningful for search engines.
+    if ( url_locale = request.path_parameters[:locale].presence )
+      lexicon_key = TaxonName::LEXICONS_BY_LOCALE[url_locale]
+      lexicon = lexicon_key && TaxonName::LEXICONS[lexicon_key.upcase.to_sym]
+      return render_404 unless lexicon && @taxon.taxon_names.any? {| tn | tn.lexicon == lexicon && tn.is_valid? }
```
pleary, line 192: Do we want all versions of the site, default and partner, to respond to these URLs? They seem to be mainly for search engine indexing, but we are disallowing indexing of these URLs for partner sites. For that reason, maybe we only want the default site to be able to respond to these

pleary (approved):
I left some comments, but nothing that need to block merging this

PR #4936, opened 2026-06-11
My sites filter on announcements index

caseyhofford:
Updates the announcements index for site admins to list all announcements, allowing site admins to duplicate global announcements to create variants for their site. In addition this explicitly enforces preventing site admins from editing announcements from other sites, rather than relying on these being inaccessible to partner site admins.


pleary (approved):
Looks good to me

PR #4941, opened 2026-06-16
feat: index what observation quality grade would be if not captive, WEB-912

pleary:
When observations are captive, calculate what their quality_grade would be if they were not captive and add that to the ES index. That will allow us to search for observations that are only casual because they are captive - i.e. the observations where quality_grade is casual and quality_grade_if_not_captive is defined but not casual

### DIFF
--- a/app/es_indices/observation_index.rb
+++ b/app/es_indices/observation_index.rb
@@ -11,11 +11,11 @@ class Observation < ApplicationRecord
   scope :load_for_index, lambda {
     includes(
       { user: [:flags, :stored_preferences] }, :confirmed_reviews, :flags,
-      :observation_links, :quality_metrics, :observation_geo_score,
+      :observation_links, :quality_metrics, :observation_geo_score, :community_taxon,
       :votes_for, :stored_preferences, :tags,
       { annotations: :votes_for },
-      { photos: :flags },
-      { sounds: :user },
+      { photos: [:flags, :moderator_actions] },
+      { sounds: [:user, :flags, :moderator_actions] },
       { identifications: [:stored_preferences, :taxon, :moderator_actions] }, :project_observations,
       { taxon: [:conservation_statuses] },
       { observation_field_values: :observation_field },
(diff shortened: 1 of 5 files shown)


caseyhofford (approved):
Looks good!

jwidness:
@pleary publicly exposing `quality_grade_if_not_captive` would effectively solve THE [top-voted forum feature request](https://forum.inaturalist.org/t/make-captive-cultivated-not-automatically-no-id-needed/112) (separating captive-that-would-be-needs-id from captive-that-would-be-rg), is there any plan to make that available?

pleary:
There's no current plan to make additional features around that attribute. The feature request that this was implemented for was for a specific project feature we recently [announced in a blog post](https://www.inaturalist.org/blog/133423-new-collection-project-filter-exclude-all-casual-except-captive).

That said, this was implemented in a way that should make it fairly straight-forward to add additional API parameters to explore these data. The trickiest part is just describing in understandable terms what those parameters would do. We'd also need to carefully construct API parameter for specific features due to how our API works where each additional parameter is basically an AND. Since we are not redefining what `quality_grade` is, we're just exposing a parameter that allows you to search what the quality_grade _would be_ if observations were not marked captive, we need to be careful with certain potential use cases. For example: show me all observations that are research grade, or would be research grade if not marked captive. Something like that necessitates a new parameter, since as soon as you add a `quality_grade=research` parameter, you are by definition excluding all casual observations, thus all observations marked as captive that would other be research grade

jwidness:
Just to be clear, I think there are two separate asks here: exposing the existing `quality_grade_if_not_captive` field in the API, and building out a full-featured non-clashing UI. My main ask is only the first, though I would definitely support the second as well.

I do see your point about non-intuitive parameter clashing, but I don't think it's really of a different scale than what's already present in the API. For example, you already can't compose `verifiable=true&quality_grade=casual` to get all verifiable observations plus casual observations.

PR #4947, opened 2026-06-17
feat: more cases to set exemplar IDs as inactive; remove access restrictions, WEB-1054

pleary:


sundisc (approved):
Looks great! One small thing though. It looks like any user could unnominate by posting to the identification endpoint... `require_owner` doesn't cover `:unnominate` and `unnominate_as_exemplar` doesn't check permissions. The button is hidden in the UI so it's unlikely this will happen, but some malicious user could potentially do this. I'm ok with a fix for this being a follow-up as it's a sort of evil corner case.

pleary:
Thanks for the feedback. I'll add some additional permissions checks there to go out shortly after release

PR #4951, opened 2026-06-18
fix(web-1016): cache apt-get results in pre-build

caseyhofford:
This PR runs `apt-get update` and `install` in `CI-pre-build.yml` persisting the results to a cache, shared by the current branch, for up to one day. This cache is then loaded by each matrix instance.

There are also a couple other optimizations added:

- The `-no-install-recommends` flag doesn't install unnecessary package dependencies.

- Set `timeout-minutes: 10` so that dependency installation will fail quickly if connections are slow.

- Set apt options to retry on request failures and timeout a request if it does not connect in 15 seconds.

- We disable man page installation which speeds up installation.

### DIFF
--- a/.github/workflows/CI-build-test.yml
+++ b/.github/workflows/CI-build-test.yml
@@ -90,12 +90,16 @@ jobs:
       run: |
         curl --verbose --show-error http://localhost:9200
 
+    - name: Compute cache date
+      id: cache-date
+      run: echo "date=$(date +%Y-%m-%d)" >> "$GITHUB_OUTPUT"
+
     - name: Restore apt package cache
       id: apt-cache
       uses: actions/cache@v4
       with:
         path: ~/apt-packages
-        key: apt-packages-${{ runner.os }}-${{ hashFiles('.github/workflows/CI-pre-build.yml', '.github/workflows/CI-build-test.yml') }}
+        key: apt-packages-${{ runner.os }}-${{ steps.cache-date.outputs.date }}-${{ hashFiles('.github/workflows/CI-pre-build.yml', '.github/workflows/CI-build-test.yml') }}
 
     - name: Install dependencies
       timeout-minutes: 10
--- a/.github/workflows/CI-build-test.yml
+++ b/.github/workflows/CI-build-test.yml
@@ -96,7 +96,7 @@ jobs:
 
     - name: Restore apt package cache
       id: apt-cache
-      uses: actions/cache@v4
+      uses: actions/cache/restore@v4
       with:
         path: ~/apt-packages
         key: apt-packages-${{ runner.os }}-${{ steps.cache-date.outputs.date }}-${{ hashFiles('.github/apt-packages.txt') }}
--- a/.github/workflows/CI-build-test.yml
+++ b/.github/workflows/CI-build-test.yml
@@ -99,7 +99,7 @@ jobs:
       uses: actions/cache@v4
       with:
         path: ~/apt-packages
-        key: apt-packages-${{ runner.os }}-${{ steps.cache-date.outputs.date }}-${{ hashFiles('.github/workflows/CI-pre-build.yml', '.github/workflows/CI-build-test.yml') }}
+        key: apt-packages-${{ runner.os }}-${{ steps.cache-date.outputs.date }}-${{ hashFiles('.github/apt-packages.txt') }}
 
     - name: Install dependencies
       timeout-minutes: 10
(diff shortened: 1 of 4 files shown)


.github/workflows/CI-build-test.yml
```diff
@@ -90,10 +90,34 @@ jobs:
...
+        path: ~/apt-packages
+        key: apt-packages-${{ runner.os }}-${{ steps.cache-date.outputs.date }}-${{ hashFiles('.github/apt-packages.txt') }}
+
     - name: Install dependencies
+      timeout-minutes: 10
       run: |
-        sudo apt-get -qq update --fix-missing
-        sudo apt-get -yqq install libpq-dev build-essential libcurl4-openssl-dev gdal-bin proj-bin proj-data libgeos-dev libgeos++-dev libproj-dev ffmpeg
+        sudo rm -f /var/lib/man-db/auto-update
+        sudo ln -sf /bin/true /usr/bin/mandb
```
caseyhofford, line 108: Disable man page installations to increase apt-get install speed. Might be overkill, but we don't need man pages in CI.

.github/workflows/CI-build-test.yml
```diff
@@ -90,10 +90,34 @@ jobs:
...
+    - name: Compute cache date
+      id: cache-date
+      run: echo "date=$(date +%Y-%m-%d)" >> "$GITHUB_OUTPUT"
+
+    - name: Restore apt package cache
+      id: apt-cache
+      uses: actions/cache@v4
+      with:
+        path: ~/apt-packages
+        key: apt-packages-${{ runner.os }}-${{ steps.cache-date.outputs.date }}-${{ hashFiles('.github/apt-packages.txt') }}
```
caseyhofford, line 102: We read from the cache created in `CI-pre-build` allowing us to skip any network requests for `apt-get`, [scoped to the branch](https://github.com/marketplace/actions/cache#cache-scopes), with a key that will change each day, or if there are any changes to packages.

.github/workflows/CI-pre-build.yml
```diff
@@ -15,3 +15,29 @@ jobs:
...
+    - name: Compute cache date
+      id: cache-date
+      run: echo "date=$(date +%Y-%m-%d)" >> "$GITHUB_OUTPUT"
+
+    - name: Cache apt packages
+      id: apt-cache
+      uses: actions/cache@v4
+      with:
+        path: ~/apt-packages
+        key: apt-packages-${{ runner.os }}-${{ steps.cache-date.outputs.date }}-${{ hashFiles('.github/apt-packages.txt') }}
```
caseyhofford, line 29: Same as above, we create a cache for the output of apt, [scoped to the branch](https://github.com/marketplace/actions/cache#cache-scopes), with a key that will change each day, or if there are any changes to packages.

.github/workflows/CI-pre-build.yml
```diff
@@ -15,3 +15,29 @@ jobs:
...
+      if: steps.apt-cache.outputs.cache-hit != 'true'
+      run: |
+        sudo apt-get update --fix-missing
+        xargs sudo apt-get install -y --no-install-recommends -d \
+          -o Acquire::Retries=5 \
+          -o Acquire::http::Timeout="15" \
+          < .github/apt-packages.txt
+        mkdir -p ~/apt-packages/archives ~/apt-packages/lists
+        cp /var/cache/apt/archives/*.deb ~/apt-packages/archives/
+        sudo rsync -a --exclude='partial' --exclude='lock' /var/lib/apt/lists/ ~/apt-packages/lists/
```
caseyhofford, line 42: We have to use rsync here because we want to skip files we don't have permission for, which is not supported by `cp`.

.github/workflows/CI-build-test.yml
sundisc: Two nits:
- Why do we need date here? Shouldn't we only care if there are changes to the packages?
- Looks like you can use `actions/cache/restore@v4` since we're only restoring here, which will [skip the post steps](https://github.com/osy/actions-cache/blob/main/restore/README.md#only-restore-cache)

sundisc (approved):
Awesome! I have a couple of optional tiny questions / comments, but feel free to ignore them... looks good.

pleary (approved):
Looks like this will be a great improvement!

.github/workflows/CI-build-test.yml
caseyhofford: - I include date because the package versions available from the repos will change over time, and we don't pin theses packages to a specific version. I believe we DO pin versions in chef for actual deploys, but I figured having updated dependencies daily would be close enough, but weekly might be closer to reality and give us more benefit from the cache.

caseyhofford: - Thanks! That's a good call, updated in 879682da2d408d6cc6f0d67b6e5a93f116486de0

caseyhofford:
Thanks!

caseyhofford:
Thanks!

PR #4952, opened 2026-06-23
Handle throttled wikipedia responses

caseyhofford:
This PR improves our caching of throttled responses from external APIs.

- We store the HTTP response status code in the ApiEndpointCache table.

- We implement a 30 minute timeout for all throttled responses.

Adding the following file before starting the rails server allows you to mimic 429 responses from Wikipedia when testing in the development environment.

```
# config/initializers/_tmp_wikipedia_throttle.rb
if Rails.env.development?
  require "webmock"
  WebMock.enable!
  # Allow all real connections (EOL, GBIF, etc.) so only the Wikipedia stub below
  # is intercepted. Stubs always take precedence over allow_net_connect!, so the
  # Wikipedia request returns a simulated throttle response while every other API
  # call hits the real network.
  WebMock.allow_net_connect!
  WebMock.stub_request( :get, /wikipedia\.org/ ).
    to_return( status: 429, body: "You are making too many requests." )
end
```


sundisc (commented):
Looks good to me as long as we're ok with the ApiEndpointCache changes occurring in the Flickr service as well. Seems like `status_code` isn't set in that service but not sure if it matters.

app/models/api_endpoint_cache.rb
```diff
@@ -1,18 +1,42 @@
 class ApiEndpointCache < ApplicationRecord
+  THROTTLE_RETRY_MINUTES = 30
+  THROTTLED_STATUS_CODE = 429
+  # Shared by .throttled_response? (Ruby) and the .throttled scope (SQL) — keep in sync.
+  THROTTLED_BODY_PHRASE = "too many requests".freeze
 
   belongs_to :api_endpoint
 
+  scope :throttled, lambda {
```
pleary: I'm thinking there's a good chance this will be a pretty slow query with the wildcard search on `response`. It can be a large table with particularly large `response` columns. We'll already know when we write the row whether or not we consider it a throttled request. I'm wondering if it would be better to do one of two things - force the status_code of all requests that return true for `throttled?` to be `THROTTLED_STATUS_CODE`, or add another column that stores a throttled boolean. I'd be OK with the former to save a column. But if we feel like we want to preserve which responses match `THROTTLED_BODY_PHRASE`, but have a status code other than 429, we could add an additional column. Then this query can just be `status_code = THROTTLED_STATUS_CODE`, or if we add a column `throttled = true`

app/models/api_endpoint.rb
```diff
@@ -1,9 +1,14 @@
 class ApiEndpoint < ApplicationRecord
-
   has_many :api_endpoint_caches
 
+  def recently_throttled?
```
pleary, line 4: Even with an index on `request_completed_at`, depending on how quick the throttled scope is, we may want to cache this for at least say 5 minutes, or even up to ApiEndpointCache::THROTTLE_RETRY_MINUTES. I'm thinking this could be a slow query, and its unclear how often it may get called. `refresh_wikipedia_summary` (which isn't called a lot, but as an example) will end up calling it twice - every time with a different value of `request_completed_at`, preventing any kind of DB-level query caching

db/migrate/20260630203548_add_api_endpoint_completed_at_index_to_api_endpoint_caches.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddApiEndpointCompletedAtIndexToApiEndpointCaches < ActiveRecord::Migration[6.1]
+  def change
+    add_index :api_endpoint_caches, [:api_endpoint_id, :request_completed_at],
```
pleary: One thing I might recommend is that we try and deploy this sometime in off-peak hours. We don't currently have a fully populated ApiEndpointCache table in an environment where we can test how long this index will take to create. It may take more than a few seconds, during which time the table will be locked to writes, potentially affecting end-user response time. The fewer active users and fewer writes on this table during creation, the better

pleary (changes requested):
Thanks - this is looking great. I left a few comments to consider. I do think we should make some change to reduce the amount of computation the `throttled` scope will need. My other comments are similarly related to possible performance issues querying and operating on this very large table. Other than those things, this is looking OK to me. As I say in one comment - when this is ready to merge let's coordinate on that deployment. I'd prefer to merge and deploy this at a time when site usage is lower

app/models/api_endpoint_cache.rb
caseyhofford: Good call, I set status code on the ApiEndpointCache records in 031b9baa77461c4e7769676e7263d29290aac72e and then also moved the logic around persisting the response and code into ApiEndpointCache in 7bb08217c97fed0735e23dbe3dd5a0da09af0e42 to avoid pulling in the constant `THROTTLED_STATUS_CODE` from `ApiEndpointCache` to `MetaService`  and let the Model have more responsibility for how records are persisted.

app/models/api_endpoint.rb
caseyhofford: This is a very good point and made me realize it is probably better to store last_throttled_at on ApiEndpoint rather than derive that from ApiEndpointCache since throttling applies to the API, not individual requests. This addresses the performance concerns here, as well as concerns around creating the index, because the index is no longer needed. I've made this change in 5a355d1

db/migrate/20260630203548_add_api_endpoint_completed_at_index_to_api_endpoint_caches.rb
caseyhofford: I think I will drop this index in favor of adding the `last_throttled_at` timestamp to the ApiEndpoint 5a355d1.(see https://github.com/inaturalist/inaturalist/pull/4952/changes#r3531777118)

caseyhofford:
> Looks good to me as long as we're ok with the ApiEndpointCache changes occurring in the Flickr service as well. Seems like `status_code` isn't set in that service but not sure if it matters.

Good point @sundisc , I think there is a linear issue for flickr 429 responses specifically. https://linear.app/inaturalist/issue/WEB-455/flickr-429-errors-preventing-photo-import

Because the flickr gem handles response codes differently, this is a slightly larger change, but I will do this as a follow up.

caseyhofford (commented):
Thanks @sundisc and @pleary , based on the review I've implemented some significant changes to how `recently_throttled` is determined (storing on ApiEndpoint, rather than ApiEndpointCache). This should address the performance concerns you raise @pleary and just feels like a better design. These changes are all contained in the last 4 commits of this branch (https://github.com/inaturalist/inaturalist/pull/4952/changes/add575ce65a20bd6f69d294a5926ef86ce57403b..5a355d124c57acd83be0545b55fcfe5bf1687d07).

pleary (approved):
I think these changes will help a lot the potential slow query performance of an earlier version. I think this will all be a good improvement. I say let's go ahead and merge this and see how it goes.

I could potentially see a future improvement if we need it - right now we're rendering the throttling message if `recently_throttled?` is true, and the description is blank. It is possible there was _some_ recently throttled response, and the target description is legitimately blank (i.e. the API response was fetched successfully, but empty). In those cases we may prefer to show the `no_matching_article_from_x` message as opposed to the throttling one. But doing that seems like it would be a pain given how all this was originally implemented, since the underlying API information is obfuscated through TaxonDescribers or `taxon.set_wikipedia_summary`. Given it's not an easy distinction to make, I say it's not worth making the changes necessary to accommodate that unless we often see the throttling message when the target response was not actually throttled.

PR #4956, opened 2026-06-24
fix(WEB-1062): fix announcement observation date filters

caseyhofford:
- invalidates `last_observation_created_at_cache_key` when observations are created or destroyed
- converts `last_observation_created_at` to a date before comparison with announcement filters


pleary (approved):
Looks good

PR #4958, opened 2026-06-25
feat: Observation Details Responsiveness, WEB-983

sundisc:
- Responsiveness updates to the observation details page. Lots of small changes to make sure everything fits, but broadly:
    - Modify ActivityItem content flow to fit narrow screens
    - Convert MoreFromUser to use shared Carousel and TaxonThumbnail
    - Ensure Annotations table content fits by slightly modifying some columns
- Converted MoreFromUser to .tsx and refactored some of the logic. Added tests.
- Converted ActivityItem to .tsx without very much refactoring (this is a huge file that should be split at some point). Added tests.
- Very slight modification to Annotations that didn't feel worthy of .tsx conversion
- Carousel buttons disappear when all items fit
- Added ability to define photo to be used in TaxonThumbanil as optional prop
- TaxonThumbnail text grows / shrinks based on size of thumbnail


app/assets/stylesheets/index.scss
```diff
@@ -2180,7 +2180,7 @@ $header-line-height: 34px;
     opacity: 1;
   }
   .navtabs {
-    margin-right: 20px;
+    margin-right: clamp(5px, 3.5vw, 20px);
```
sundisc, line 2183: Let margin on hamburger menu and logo compress a bit when screen gets narrow so that the header doesn't overflow.

app/webpack/observations/show/components/app.tsx
```diff
@@ -206,28 +249,24 @@ class App extends React.Component {
...
-              {
-                I18n.t( "notes", {
-                  defaultValue: I18n.t( "activerecord.attributes.observation.description" )
-                } )
-              }
-            </h3>
-            <UserText text={observation.description} />
-          </Col>
-        </Row>
+        <div className="description">
```
pleary, line 260: This class is applying italics to this div, which we probably don't want

app/webpack/observations/show/components/app.tsx
```diff
@@ -257,215 +295,167 @@ class App extends React.Component {
...
+                    </MenuItem>
+                  </SplitButton>
+                </div>
+              ) : config?.currentUserCanInteractWithResource( observation ) && (
+                <FollowButtonContainer />
+              ) }
+          </div>
+          <div className="top_container">
+            <div className="top_row">
+              <div className={`photos_column ${photosColClass}`}>
```
pleary, line 376: There's a 1px right border on the `photos_column` class that can probably be removed

app/webpack/observations/show/components/app.tsx
```diff
@@ -257,215 +295,167 @@ class App extends React.Component {
...
+                        uniqueKey="DateObserved"
+                      />
+                    ) }
+                  </div>
+                  <div>
+                    <span className="bold_label">{ I18n.t( "label_colon", { label: I18n.t( "submitted" ) } ) }</span>
+                    {this.displayDateAdded( )}
+                  </div>
+                </div>
+                <MapContainer />
```
pleary, line 409: The map section contains a "Details" link that opens an overlay which is rendering beyond the viewport on narrow displays. It would be nice to be able to make that overlay more responsive

Image

app/webpack/observations/show/components/activity_item.tsx
```diff
@@ -0,0 +1,900 @@
...
+    }
+    if ( userCanInteract && item.firstDisplay && !hideCompare ) {
+      buttons.push( (
+        <button
+          key={`id-compare-${item.uuid}`}
+          type="button"
+          className="btn btn-default btn-sm"
+          onClick={e => {
+            if ( onClickCompare ) {
+              return onClickCompare(
```
pleary, line 463: The compare modal is very complicated and has quite a few responsiveness issues. I wonder if its even worth trying to address them, or maybe we should hide the compare modal on narrow displays. There is also a link to open the compare modal in `CommunityIdentification`

app/webpack/observations/show/components/app.tsx
```diff
@@ -257,215 +295,167 @@ class App extends React.Component {
...
+                <TagsContainer />
+              ) }
+              <ObservationFieldsContainer />
+              <IdentifiersContainer />
+              <CopyrightContainer />
+            </div>
+          </div>
         </div>
         <div className="data_quality_assessment">
           <AssessmentContainer innerWrapper={AssessmentLazyLoad} />
```
pleary, line 451: There's a few elements in the DataQualityAssessment that get pretty cramped and start to overlap on narrow displays. Maybe that upper-right div just needs to wrap under the `Quality Grade:` label, so that QG label doesn't wrap, that small graphic has room to grow, and the voting div below can grow to the width of the div

Image

app/webpack/observations/show/components/app.tsx
```diff
@@ -257,215 +295,167 @@ class App extends React.Component {
...
+                <div className="faves_row">
+                  <FavesContainer />
+                </div>
+              </div>
+            </div>
+          </div>
+          <div className="middle">
+            <div className="middle_left">
+              { description }
+              <ActivityContainer />
```
pleary: Some modals are rendering a bit offscreen. For example on a narrow phone display and you try to delete a comment or ID

Image

app/webpack/observations/show/components/app.tsx
```diff
@@ -257,215 +295,167 @@ class App extends React.Component {
...
+            </div>
+            <div className="opposite_activity">
+              <CommunityIdentificationContainer />
```
pleary: Not sure how tricky this would be, but I wonder if this component would be best rendered above the activity second on narrow displays. Subjective and maybe not what we want. I think it would read better though to show the community taxon above the list of IDs users have added

app/webpack/observations/show/components/app.tsx
```diff
@@ -257,215 +295,167 @@ class App extends React.Component {
...
-                <Row>
-                  <Col xs={responsive ? null : 6} sm={responsive ? 6 : null}>
-                    <NearbyContainer />
-                  </Col>
-                  <Col xs={responsive ? null : 6} sm={responsive ? 6 : null}>
-                    <SimilarContainer />
-                  </Col>
-                </Row>
-              </Grid>
+              <div className="more_from_user">
```
pleary, line 456: Before this PR, the Data Quality and More From sections span the entire width of the page, so their background colors are consistent and there's no border that switches back to the page background color. In this PR these sections have a margin so there's a grey border on either side, on slightly wider displays like desktop browsers. Not a big deal, but it is a change. I personally prefer having the section backgrounds span the width of the screen w/o a grey border 

Image

app/webpack/observations/show/components/app.tsx
pleary: I'm seeing some slight overflow on the narrowest of displays for the Identification search overlay

Image

app/webpack/observations/show/components/app.tsx
pleary, line 470: The Flagging modal can overflow on narrow displays. That modal is opened via the "Flag as inappropriate" link in the DQA section. It can also be opened when flagging photos, IDs, and comments

Image

pleary, line 473: This is another modal that overflows on narrow displays. It is opened by clicking the "About" button in the Community Taxon section

pleary (changes requested):
Generally looking really good. I noted a few rendering issues or changes from the existing styles we'll probably want to address. Some of the more complicated bits we may just want to suppress on narrow displays if they will be too difficult to make responsive, or have less usability on  narrow displays (e.g. the compare modal, the community ID modal).

As I mentioned on the Taxon Photos PR, we'll want to make sure all users see the responsive version of obs show. In my testing I did that by setting `@responsive = true` and `@skip_min_width = true`, and ensuring my own user had opted out of all test groups

app/webpack/observations/show/components/activity_item.tsx
sundisc: Ah.. missed that modal. That is complicated, and something that I think we don't need in mobile right now. That button is now hidden when lower than the $md (767px) breakpoint

app/webpack/observations/show/components/app.tsx
sundisc: Good point! Spanned them to the edges

app/webpack/observations/show/components/app.tsx
sundisc: Hid similar to the compare button, but waited until the screen size exceeds the $lg breakpoint because this modal's table element is quite wide.

app/webpack/observations/show/components/app.tsx
sundisc: Strange, it should be wrapping at $md but it doesn't seem to be for you. Can you check again? I also made a change to wrap the progress bar and "This observation is Research Grade!" section below Quality Grade (before it wrapped all the way under the table).

app/webpack/observations/show/components/app.tsx
sundisc: I was originally organizing based on the way divs were originally structured, but I think your idea makes more sense here. I put it between notes and activity.
Screenshot 2026-07-08 at 12 02 33 PM

app/webpack/observations/show/components/app.tsx
pleary: I'm seeing them stretch a little wider, but not beyond the width of `#ObservationShow`, which has a max-width of `1320px`. I'm also now seeing a slight gap in between the QDA and "More from". Seems like `. QualityMetrics` has bottom padding, but that padding extends below the wrapping `.data_quality_assessment`. Pretty minor in the context of these larger improvements

app/webpack/observations/show/components/activity_item.tsx
pleary: A couple other minor modal width issues in the `CommunityIdentification` section. Tapping on the taxon photo in that section opens a modal with a taxon description and a link to the taxon page, which can be wide. And clicking on the green or red dashes representing identifications opens a modal showing that taxon's taxonomic ancestry. That one can also be wider than very narrow displays

app/webpack/observations/show/components/app.tsx
pleary: Thanks this is looking better now. I agree that's the best order of display for this section

app/webpack/observations/show/components/app.tsx
pleary, line 479: Another minor edge case, but when we display the `RtlTestGroupToggle` alert div for some users, it's in a `.container` which can extend beyond the page contents towards the shorter end of the `.container` breakpoints. For example on displays just over 1200px, or just over 1000px, I see a horizontal scrollbar because the fixed width of this div overflows the page width

pleary (approved):
I left a few more minor comments. We'll likely continue to find small edge cases on less commonly used elements, or for content edge cases we didn't test on. Overall though, this looks like a great improvement over what's in production for the responsive test, and as far as I can tell this is ready to put in front of users. I'm not seeing any blockers for release. Thanks!

PR #4959, opened 2026-06-29
feat: add project filter to exclude observations casual for reasons other than captive, WEB-912

pleary:
Allow collection projects to set a filter to use a new API parameter `not_casual_excluding_captive`. If configured to allow casual observations, that filter will exclude all casual observations that would remain casual if they weren't marked captive. Essentially allow projects to include casual observations when the only reason they are casual is because they are marked captive


caseyhofford (approved):
Looks good, tested locally!

PR #4964, opened 2026-06-30
fix: Ensure mobile settings header and footer span screen, WEB-1067

sundisc:
While there, Abhas asked me to implement a few design changes.

Before:
Screenshot 2026-06-30 at 3 47 19 PM

After:
Screenshot 2026-06-30 at 5 10 02 PM


pleary (approved):
Looks good to me

PR #4968, opened 2026-07-08
web 1011 refactor gbif observation links

caseyhofford:
- Finishes the `lib/data_partner_linkers/gbif.rb` task to match other partners (e.g. `lib/data_partner_linkers/ala.rb`).
- Adds `lib/tasks/gbif_observation_links.rake` to cleanup data drift of `ObservationLinks` between ES and PG.
- Adds a migration `db/migrate/20260708030248_add_gbif_data_partner.rb` to seed the data partner row.
- Leaves `tools/gbif_observation_links.rb` to allow easily swapping back in case of issues.
- Adds support for GBIF Occurrence SQL download API: https://techdocs.gbif.org/en/data-use/api-sql-downloads decreasing the download from ~300gbs to ~4gbs.
- Increases efficiency by running two SQL SELECT statements instead of 2n SELECT statements per batch.
- Increases batch size to 10k rows. This was tested locally and works, and shouldn't risk overwhelming memory for the process running this. PG should also not have any issues at this scale with selects on indexed columns. We may be able to go higher but this seems like a good tradeoff. 100k rows took 13s, 10k took 15s, 1k took 20s.

Tested locally on a 1M record subset of the GBIF API SQL output, the `#process_result` method ran in 15.4seconds. This would translate to around 40 minutes, though timing might change in production.

**This DOES NOT include other possible improvements by using DelayedJobs:**
- Using DelayedJobs to schedule the job after requesting the download. Requesting a download from GBIF may take hours, so we don't need a process looping this whole time.
- Running the batches of 10k in parallel with DelayedJobs.

[**Ops Follow Up Ticket**](https://linear.app/inaturalist/issue/OPS-399/update-gbif-sync-scheduled-task-and-run-drift-cleanup)


db/migrate/20260708030248_add_gbif_data_partner.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
```
pleary: My recommendation for now would be to not create a DataPartner record for this. I'm not sure the `DataPartner.sync_observation_links` is best suited to very long-running linking jobs. I also suspect, but haven't thoroughly checked, that the way it uses `last_sync_observation_links_at` to schedule which tasks to run will cause every other scheduled run to fail since the jobs take a non-zero amount of time to finish and we can't be certain how consistent the run times are of any given task. For now I recommend we schedule `DataPartnerLinkers::GBIF` in its own cron task, which is what we've been doing for ALA, and that does not require a DataPartner linker DB record to exist.

Alternatively we could add some column to DataPartners to indicate they should not be auto scheduled, and update the `DataPartner.sync_observation_links` method so it doesn't try to run the ones we want to trigger separately. The really long ones I feel more comfortable scheduling separately. We'll also want to use the `sql_download` option for this task, which the centralized `DataPartner.sync_observation_links` won't accommodate right now

lib/data_partner_linkers/gbif.rb
```diff
@@ -41,6 +44,24 @@ def request
       logger.debug "Received key: #{@key}"
     end
 
+    # Requests an occurrence download via GBIF's SQL Downloads API instead of
+    # the predicate API. The SQL selects only the columns process_result reads
+    # (gbifID, catalogNumber), so GBIF generates a much smaller archive.
+    # See https://techdocs.gbif.org/en/data-use/api-sql-downloads
+    def request_filtered
+      url = "https://#{@username}:#{@password}@api.gbif.org/v1/occurrence/download/request"
```
pleary: Usernames and passwords may need to be sent through `CGI.escape` if they contain non URI-safe characters, for example `@`. We also probably shouldn't log this full interpolated url (`logger.debug "Requesting #{url}"`) as it contains authentication information in it. These both apply to the untouched `request` method as well

lib/tasks/gbif_observation_links.rake
```diff
@@ -0,0 +1,54 @@
...
+  desc "Reindex observations whose GBIF ObservationLinks are missing from their Elasticsearch docs"
+  task backfill_es_drift: :environment do
+    drifted_ids = gbif_es_drifted_observation_ids
+    puts "[#{Time.now}] #{drifted_ids.size} observations have GBIF ObservationLinks not represented in ES"
+    if drifted_ids.empty?
+      puts "[#{Time.now}] Nothing to reindex"
+      next
+    end
+    puts "[#{Time.now}] Queueing delayed jobs to reindex #{drifted_ids.size} observations..."
+    Observation.elastic_index!( ids: drifted_ids, delay: true, run_at: 1.minute.from_now )
```
pleary: My only comment here is if there's some major change in IDs, or something goes wrong, this list of drifted_ids could be really large since there's over 150M records in the remote source. I think building in some kind of batching to create multiple indexing jobs with a fixed limit (e.g. 1k) here, or building queueing and indexing in the main method itself, accommodating somehow the other task that just prints the IDs, could help prevent some issue where we queue up an indexing job with millions of IDs in it

pleary (changes requested):
Overall this is looking really good. I left a couple comments. This should be a huge improvement to how well this process works, and how long it takes. Thanks for the work on it!

db/migrate/20260708030248_add_gbif_data_partner.rb
caseyhofford: Ok! That makes sense @pleary, I'll remove this migration and we can schedule it manually, then consider updating the `DataPartner` system once we have an idea of how long the updated task takes to run. Done in 660cb2c47fcc29c4512b2baa3fa2b7cdda556d63 (the diff includes `structure.sql` noise but that is fixed after the merge commit f4efc7ad48844c4cef249d58fa41f3c99177fea0).

lib/data_partner_linkers/gbif.rb
caseyhofford: Good call, fixed this in 692ecd1c135f1886331282895314a238ef0623ae

lib/tasks/gbif_observation_links.rake
caseyhofford: Good call, I've update the rake task to batch these requests, and incrementally pull IDs out of the queue for reindexing. This took a few commits with some other cleanup in between. Probably easiest to see the net changes by looking at https://github.com/inaturalist/inaturalist/pull/4968/changes/692ecd1c135f1886331282895314a238ef0623ae..c3cc21cb2a21d2cf91fb563579987f5dcf36a7dd and looking at changes for just this file.

caseyhofford:
Thanks @pleary! I've addressed the comments.

pleary (approved):
Looks great - thanks for making these changes

PR #4969, opened 2026-07-09
feat: change project requirements casual label when excluding non-captives, related to WEB-912

pleary:
Projects have a new setting to only include casual observations when the only reason they are casual is because they are marked non-wild. This PR updated the project requirements quality grade label from `Casual` to `Casual (non-wild only)` to indicate that project setting is active

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -1031,6 +1031,9 @@ en:
   card: Card
   casual: casual
   casual_: Casual
+  # Label for the Casual quality grade, but indicating it only includes observations that are marked
+  # casual because they have a quality assessment vote that they are non-wild
+  casual_non_wild_only_: Casual (non-wild only)
   casual_tooltip_html: '"Casual" observations are missing media, location, or a date, or do not meet the requirements of the "Research" or "Needs ID" quality grades for other reasons. <a href="https://www.inaturalist.org/pages/help#quality">Learn more about quality grades</a>'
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -1031,6 +1031,9 @@ en:
   card: Card
   casual: casual
   casual_: Casual
+  # Label for the Casual quality grade, but indicating it only includes observations that are marked
+  # casual because they have a quality assessment vote that they are non-wild
+  casual_non_wild_only_: Casual (non-wild only)
```
caseyhofford, line 1036: Just to confirm from what I'm seeing in a quick search. Translations keys ending in `_` are in sentence case, rather than lower case? Is this a convention, or does it impact UI and/or machine translation in crowdin?

caseyhofford (approved):
Tested locally and looks great :) thanks @pleary. Added one question for my understanding of translation key naming conventions.

Image

config/locales/en.yml
pleary: That key naming has no special meaning for our i18n plugin, Crowdin, or anything else. It's just a custom convention we've used for some key names, that we're not even consistent with. When the base key not ending in `_` uses one case variant, we have sometimes added a second key ending in `_` that inverts the case of the first letter. Many times the `_:` variant has been sentence-case, but sometimes its the other way around. For this PR, I was just copying what we did with `casual_:` - leaving room for if we did add a lowercase version of the same we could call it `casual_non_wild_only_:`. Not necessary and there's no special consideration for this key naming

PR #4972, opened 2026-07-10
feat: use v2 map tile endpoints, WEB-1061

pleary:
* Prefers v2 mapping endpoints over v1 versions
* Updates the observation uploader to use APIv2

Eventually I'd like to clean up how we reference APIv2. Perhaps by adding a new meta tag that contains the V2 URL instead of having to do a string replace in several places. I'd also like to swap out the React config `testingApiV2` with something else as V2 is no longer a test, but more often than not the default. I'll do that in a later PR - this PR implements using v2 in a way consistent with how its been done up until now

### DIFF
--- a/app/webpack/observations/uploader/webpack-entry.js
+++ b/app/webpack/observations/uploader/webpack-entry.js
@@ -3,11 +3,13 @@ import { render } from "react-dom";
 import { Provider } from "react-redux";
 import _ from "lodash";
 import moment from "moment";
+import inatjs from "inaturalistjs";
 
 import reducers from "./reducers";
 import Uploader from "./containers/uploader";
 import { fetchSavedLocations } from "./ducks/saved_locations";
 import sharedStore from "../../shared/shared_store";
+import { setConfig } from "../../shared/ducks/config";
 
 moment.locale( I18n.locale );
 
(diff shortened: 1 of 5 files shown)


app/webpack/observations/uploader/webpack-entry.js
```diff
@@ -17,6 +19,18 @@ if ( !_.isEmpty( CURRENT_USER ) ) {
   sharedStore.dispatch( fetchSavedLocations( ) );
 }
 
+const element = document.querySelector( "meta[name=\"config:inaturalist_api_url\"]" );
```
caseyhofford, line 22: When do we use `public_node_api_url` vs `node_api_url` that the meta tag can contain in practice? Are different configs used in different environments?

caseyhofford (approved):
Looks good and runs locally. Agreed we should DRY up some of the v1 -> v2 migration code at some point but this matches the existing patterns.

app/webpack/observations/uploader/webpack-entry.js
pleary: It was added at one point for some case hit when running a containerized version of the app. I think though it may not be used anymore as I'm not seeing it referenced in config.docker.yml anymore. I suspect we can remove all references to `public_node_api_url` now

PR #4973, opened 2026-07-13
fix: remove admin gate from ModeratorAction#editable_by?

caseyhofford:
Remove an additional admin gate for timed suspensions, enabling editing of suspensions for non-admins who created the suspension.


caseyhofford:
This change is not needed.

PR #4974, opened 2026-07-13
announcement impression improvements

caseyhofford:
- Counts impressions even if IP address is excluded.
- Fixes support for passing `user_id` rather than `user` to `increment_for_announcement`.


caseyhofford:
Merged in https://github.com/inaturalist/inaturalist/pull/5005

PR #4975, opened 2026-07-13
fix: project maps should not use any as default parameter value, related to WEB-1061

pleary:
API v2 doesn't accept "any" as a parameter default that basically does nothing, e.g. captive=any. The project page maps were still sending a few parameters with "any" as a value, leading to 422 errors rendering map tiles. This PR removes any parameter whose value is "any"


caseyhofford (approved):
Looks good and tested locally!

PR #4977, opened 2026-07-14
web-1009: replace kmz rendering w/ client geojson conversion

caseyhofford:


app/assets/javascripts/taxa/map.gmaps.js
```diff
@@ -1,35 +0,0 @@
-$(document).ready(function() {
```
caseyhofford, line 1: This file appears to be orphaned.

app/views/observations/new.html.erb
```diff
@@ -16,14 +16,15 @@
...
+  <%= javascript_include_tag "webpack/kml-webpack" if @kml_assets.present? %>
   <script type="text/javascript" charset="utf-8">
     var DEFAULT_PHOTO_IDENTITY_URL = <%= @default_photo_identity_url.to_json.html_safe %>,
         PHOTO_IDENTITY_URLS = [<%= @photo_identity_urls.join(',').html_safe %>],
         SOUNDCLOUD_IDENTITY = <%=raw current_user.soundcloud_identity.to_json(:only => [:id]) %>,
         PROJECT = <%= @project.to_json(:only => [:id, :place_id, :latitude, :longitude, :map_type, :zoom_level, :title]).html_safe %>
         PLACE = <%= @place.to_json.html_safe %>,
         SHOW_PLACE_GEOMETRY = <%= @place ? "true" : "false" %>,
-        KML_ASSET_URLS = <%= @kml_assets ? @kml_assets.map{|a| root_url.gsub(/\/$/, "") + a.asset.url}.inspect.html_safe : '[]' %>,
+        KML_ASSET_URLS = <%= @kml_assets ? @kml_assets.map{|a| a.asset.url}.inspect.html_safe : '[]' %>,
```
caseyhofford, line 27: Matches https://github.com/inaturalist/inaturalist/commit/ada34a1f82ad88249bfed9393b6843dafbde9b9a

app/webpack/kml/kml_to_geojson.js
```diff
@@ -0,0 +1,55 @@
+const toGeoJSON = require( "@tmcw/togeojson" );
+const JSZip = require( "jszip" );
+const sanitizeHtml = require( "sanitize-html" );
+
+function kmlTextToGeoJSON( kmlText ) {
+  const dom = new DOMParser( ).parseFromString( kmlText, "application/xml" );
+  if ( dom.querySelector( "parsererror" ) ) {
+    throw new Error( "Could not parse KML" );
+  }
+  return toGeoJSON.kml( dom );
```
pleary: I tested one project, e.g. https://www.inaturalist.org/observations/new?project_id=85, and it has multiple `Folders` in the KML with `Placemarks` that re-use IDs. The Google Maps `KmlLayer` was still seemingly rendering all features OK, but in this PR some of them (maybe of the red trail lines) are not being rendered. Looks like if we reassign the features to have new unique IDs it works better, e.g. 
```diff
diff --git a/app/webpack/kml/kml_to_geojson.js b/app/webpack/kml/kml_to_geojson.js
index b22b001e4..e2cdb5111 100644
--- a/app/webpack/kml/kml_to_geojson.js
+++ b/app/webpack/kml/kml_to_geojson.js
@@ -7,7 +7,17 @@ function kmlTextToGeoJSON( kmlText ) {
   if ( dom.querySelector( "parsererror" ) ) {
     throw new Error( "Could not parse KML" );
   }
-  return toGeoJSON.kml( dom );
+
+  const geojson = toGeoJSON.kml( dom, { skipNullGeometry: true } );
+  geojson.features = geojson.features.reverse( );
+  geojson.features.forEach( ( feature, index ) => {
+    if ( feature.id ) {
+      feature.properties.kml_id = feature.id;
+    }
+    feature.id = `kml-feature-${index}`;
+  } );
+
+  return geojson;
 }

 function kmzToKmlText( arrayBuffer ) {
```

It also seems like the render order changes every time you enable and disable the layer through the overlay menu. The Google Maps `KmlLayer` seemed to render them consistently as I'm seeing the paths above the grid consistently. I don't think it's worth spending any time on that and we can just accept whatever render order we get.

I'll note here one other feature of these KMLs that works in production but not in this PR is the ScreenOverlay Legend, e.g. https://www.inaturalist.org/observations/new?project_id=10. Personally, I think it's OK to drop these. Just about all these project KMLs are older, and make seem to use HTTP not HTTPS when referencing their legend images. Dropping support for displaying these insecure images for a rarely used older feature seems OK to me.

I think it's OK for some aspects of old KML files to lose support, so we don't need to go crazy trying to make sure they all work properly. I think I checked all projects with KML assets (IDs: [3,10,44,51,85,104,107,463,512,605,626,1090,1929,4639]) and they're all looking pretty good. I noticed that in production on the project show page, the KMLs don't seem to be rendering (e.g. https://www.inaturalist.org/projects/2011-bioblitz-at-don-edwards-san-francisco-bay-national-wildlife-refuge), but they are in this PR which I think is an improvement.

pleary (approved):
This is looking good to me. I left one comment about one targeted change we could consider to better handle Placemarks with non-unique IDs, which some project KMLs contain. Otherwise, this looks like a great conversion from `google.maps.KmlLayer` to `google.maps.Data`

app/webpack/kml/kml_to_geojson.js
caseyhofford: Good catch! I've updated `kmlTextToGeoJSON`in 00c210a, added specs for this in 068185f, and added [an issue to add support for ScreenOverlay here](https://linear.app/inaturalist/issue/WEB-1250/support-kml-screenoverlay-legends).

PR #4979, opened 2026-07-17
feat: update taxon browse_photos to use API v2, WEB-1104

pleary:
This PR makes it so the taxon browse_photos page uses API v2. It also fixes some buggy parameter cleaning I added to projects, since I was making similar changes to parameter cleaning for browse_photos. I will later submit a cleanup PR for all this API v2 work to remove the `testingApiV2` config name with something more appropriate


sundisc (approved):
Looks good!

PR #4980, opened 2026-07-17
fix: taxon browse_photos style fixes, WEB-1102

pleary:
* Makes it so taxon names in the browse_photos grid are white, which is what they were until some recent changes
* It also makes it so the entire grid cell is clickable to open the photo modal, which was also the behavior until recently
* Adds a translation for a link label that previously had no matching translation


sundisc (approved):
Awesome, thanks!

PR #4981, opened 2026-07-17
feat: Taxon Details Responsiveness, WEB-984

sundisc:
Makes the taxon detail page (`taxa/show`) responsive behind a gate: unless the user is in the `responsive-taxon-detail` or `responsive-global` test groups, they will continue to see the original page's styling and markup. This was done by:
   - Preserving all original components by adding the `_legacy` filename and dispatching between them at runtime via the new [responsive.ts](https://github.com/inaturalist/inaturalist/pull/4981/files#diff-7da58695767620f1235782602f747e696d2082ecee5b27aee0a088e15411f607) gate (reads the `CURRENT_USER` global, so leaf components need no rewiring) — see [app.tsx](https://github.com/inaturalist/inaturalist/pull/4981/files#diff-64271ad2027aa6403c11df75be4286825ccf7cd948b735cacf5a3a76ec8e65cf) for the dispatch pattern
   - Preserving original styling in [show_legacy.scss](https://github.com/inaturalist/inaturalist/pull/4981/files#diff-c422078353310a1f6e830c6778e671962795ab2106f26115b773785e8796cca7) and switching stylesheets based on test-group inclusion in [show.html.haml](https://github.com/inaturalist/inaturalist/pull/4981/files#diff-412cd5372ff2a90a6cccc2329e33b518f8cfeb5ab2001f0591cff0a3c9bfbc76)
   - Gating on the backend in [taxa_controller.rb](https://github.com/inaturalist/inaturalist/pull/4981/files#diff-11717f28d1c073464a62c291278d951ea9ad737a49da046ccfb61fdd5b960737), where `@responsive_taxon_detail` sets `@responsive` / `@skip_min_width` (replacing the old `responsive-header` + `responsive-taxon-detail` gate) so the `980px` min-width is dropped for testers

  Beyond the gate, the responsive version modernizes the page:
  - **Typescript + React refactor** — the rewritten tab and hero components move from JSX to TSX and drop Bootstrap `Grid/Row/Col` layout in favor of CSS grid/flexbox (hero is now a 2-column grid that stacks below `$md`); UI components are retained
  - **Shared `TaxonPageHeader`** replaces the bespoke header (TaxonCrumbs still rendered inside it)
  - **Mobile nav** — tab bar is hidden below `$md` and replaced by the shared `TabDrawer` drawer UI; tab switching is React-controlled instead of Bootstrap jQuery `data-toggle`
  - **Carousels replaced** — the Map tab's Recent Observations and the Highlights tab drop the Bootstrap jQuery paged carousel for the shared CSS scroll carousel (photoless observations are filtered out; headings render via the carousel's `title` prop)
  - **PhotoPreview** — no longer fixed at 590px; fills the hero grid cell and lays out its thumbnail strip / "View More" cell with flexbox instead of absolute positioning
  - **Identifications tab** — list and map sit side-by-side at desktop (`flex: 2` / `flex: 1`); map is hidden and the list goes full-width on mobile, cards stack, and filters wrap
- **Tab layouts** (About, Taxonomy, Status, Interactions) unified on `.tab-row / .tab-main / .tab-side`; status tables wrapped in `.table-responsive`
- **Similar Species** thumbnails switch to a centered flex-wrap layout
- **`ObservationsGridItem`** extracted to a shared [TSX component](https://github.com/inaturalist/inaturalist/pull/4981/files#diff-445d9030dc33a960acee17f4f78b42c9f5faac9f17e42ee7f9ac6161d444ed55) (with a `_legacy` sibling) since it's also consumed by Identify and Year-in-Review; `splitTaxonOptions` spread replaced with explicit `noParens` / `noInactive` props


app/webpack/taxa/show/components/photo_preview.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+  it( "switches the main photo when a thumbnail is clicked", ( ) => {
+    render(
+      <PhotoPreview
+        taxon={{ id: 1 } as unknown as Taxon}
+        taxonPhotos={entries( 1, [11, 12, 13] )}
+        layout="gallery"
+      />
+    );
+    expect( screen.getByTestId( "main-photo" ) ).toHaveTextContent( "photo-11" );
+    fireEvent.click( screen.getAllByLabelText( "view_photo" )[2] );
```
caseyhofford: I think we're missing the translation key for this in `config/locales/en.yml`.

app/webpack/taxa/show/components/status_tab.tsx
```diff
@@ -0,0 +1,429 @@
...
+                    ) }
+                  </td>
+                  <td>
+                    <div className="media">
+                      <div className="media-body">
+                        { source }
+                      </div>
+                      { status.url && (
+                      <div className="media-right">
+                        <a href={status.url} aria-label={I18n.t( "view_source" )}>
```
caseyhofford, line 243: I think we're missing this translation key as well.

app/webpack/taxa/show/components/app.tsx
```diff
@@ -0,0 +1,83 @@
...
+import TaxonPageHeader from "../../shared/components/taxon_page_header";
+import PhotoPreviewContainer from "../containers/photo_preview_container";
+import ChartsContainer from "../containers/charts_container";
+import Leaders from "./leaders";
+import TaxonPageTabsContainer from "../containers/taxon_page_tabs_container";
+import PhotoModalContainer from "../containers/photo_modal_container";
+import PhotoChooserModalContainer from "../containers/photo_chooser_modal_container";
+import PlaceChooserContainer from "../containers/place_chooser_container";
+import TaxonChangeAlertContainer from "../containers/taxon_change_alert_container";
+
```
caseyhofford: NIT: Is this empty line intentional?

app/webpack/taxa/show/components/app.tsx
```diff
@@ -0,0 +1,83 @@
...
+          <h1>
+            <SplitTaxon taxon={taxon} user={config.currentUser} />
+            { flagsButton }
+          </h1>
+      )}
+        afterSelect={( result: { item: unknown } ) => showNewTaxon( result.item )}
+        showNewTaxon={showNewTaxon}
+        placeChooser={<PlaceChooserContainer container={$( "#app" ).get( 0 )} clearButton />}
+        prefix={<TaxonChangeAlertContainer />}
+        extra={<AkaNamesContainer />}
```
caseyhofford: NIT: I know this isn't really from this PR, but I'd suggest calling this prop `suffix` for consistency with the `prefix` prop.

app/webpack/taxa/show/components/app.tsx
```diff
@@ -0,0 +1,83 @@
...
+interface Props {
+  taxon: Taxon;
+  showNewTaxon: ( taxon: unknown ) => void;
+  config?: Config;
+}
+
+const App = ( { taxon, showNewTaxon, config = {} }: Props ) => {
+  const flagsButton = isCuratorOrAdmin( config.currentUser )
+    && taxon.flag_counts
+    && ( taxon.flag_counts.unresolved ?? 0 ) > 0
```
caseyhofford: I believe this is equivalent logic to the `null`/`undefined` check in `app/webpack/taxa/show/components/app.tsx:62-63` which I would slightly prefer because I think it's easier to quickly parse what it is doing, and separating the line of the null check from the flag_count check could make debugging errors easier.

```suggestion
    && taxon.flag_counts.unresolved
    && taxon.flag_counts.unresolved > 0
```

app/webpack/taxa/show/components/app.tsx
```diff
@@ -0,0 +1,83 @@
...
+        config={config}
+        heading={(
+          <h1>
+            <SplitTaxon taxon={taxon} user={config.currentUser} />
+            { flagsButton }
+          </h1>
+      )}
+        afterSelect={( result: { item: unknown } ) => showNewTaxon( result.item )}
+        showNewTaxon={showNewTaxon}
+        placeChooser={<PlaceChooserContainer container={$( "#app" ).get( 0 )} clearButton />}
```
caseyhofford: Since both uses of `TaxonPageHeader` provide the same props but pass different container references, can we change this prop to be `placeChooserContainer`, and pass the container in? We may even be able to drop this prop entirely and just put the `PlaceChooserContainer` within `TaxonPageHeader`  if `$( "#app" ).get( 0 )` still evaluates to the React app root element from within `TaxonPageHeader`.

app/controllers/taxa_controller.rb
```diff
@@ -230,11 +230,12 @@ def show
         else
           session[:preferred_taxon_page_ancestors_shown]
         end
-        if current_user&.in_test_group?( "responsive-header" ) &&
-            current_user.in_test_group?( "responsive-taxon-detail" )
+        @taxon_detail_responsive = current_user&.in_test_group?( "responsive-taxon-detail" ) ||
+          current_user&.in_test_group?( "responsive-global" )
```
caseyhofford, line 234: Just confirming the change from the test group `"responsive-header"` to `"responsive-global"` is intentional and that the `"responsive-header"` group no longer needs `@skip_min_width = true`.

app/webpack/taxa/show/components/articles_tab.tsx
```diff
@@ -0,0 +1,137 @@
+import React from "react";
+import _ from "lodash";
+import { isCuratorOrAdmin } from "../../shared/util";
+import type { Taxon, CurrentUser } from "../../../shared/types";
+
+type TaxonWithVision = Taxon & { vision?: boolean };
```
caseyhofford: Do we want to add `{ vision?: boolean };` as an optional boolean on the shared `Taxon` type, rather than just here? It looks like this taxon prop is being pulled out of global state, and that `vision` is provided on the get taxa by ID API endpoint: `http://localhost:4000/v2/docs/#/Taxa/get_taxa__id_`.

app/webpack/taxa/show/components/identifications_tab.tsx
```diff
@@ -0,0 +1,948 @@
...
+  photos: { photoUrl: ( size?: string ) => string }[];
+  annotations: {
+    uuid: string;
+    controlled_attribute: AnnotationValue;
+    controlled_value: AnnotationValue;
+  }[];
+  discussion_count: number;
+}
+
+interface Identification {
```
caseyhofford: Some of these interfaces might make sense as shared types?

app/webpack/taxa/show/components/identifications_tab.tsx
```diff
@@ -0,0 +1,948 @@
...
+
+  sortSelect( ) {
+    const {
+      identificationsQuery,
+      setIdentificationsQuery
+    } = this.props;
+    const activeTab = this.activeTab( );
+    const sortOptions = this.state.tabSorts[activeTab];
+    return (
+      <select
```
caseyhofford, line 705: Image

This isn't a blocker at all, but I think there might be some changes needed on the dropdown menu for the identification tips tab of the page. The menu item displacement in the screenshot might be just an artifact of testing this in Chrome DevTools, but the width of the actual dropdown could probably be wider.

caseyhofford (changes requested):
Mostly refactor or style comments, the only blocking comment is adding the two missing translation keys (`view_photo` , `view_source`) to `en.yml` for aria labels.

app/controllers/taxa_controller.rb
sundisc: Yeah this matches how we've already done it with the observation details. It doesn't need `@skip_min_width = true`

app/webpack/taxa/show/components/identifications_tab.tsx
sundisc: Fixed! 
Screenshot 2026-08-12 at 10 52 36 AM

app/webpack/taxa/show/components/identifications_tab.tsx
sundisc: Ah yes, some of them:

- Moved to shared/types.ts
  - Vote, ExemplarIdentification — both were also declared in observations/show/components/activity_item.tsx, now updated to consume the shared ones
  - Identification, IdentificationObservation — inatjs model, so its photos now reuse the shared Photo type instead of redeclaring photoUrl()
  - IdentificationsQuery, SortOption, Bounds — generic enough to be worth sharing now rather than waiting for a second consumer
- Deleted outright
  - AnnotationValue and the inline { controlled_attribute, controlled_value } shapes — the existing shared ControlledAttribute/ControlledValue/TermValue already covered them
- Kept local
  - IdentificationsResponse — loading is injected by the container mid-fetch, so it's a view model
  - IdentificationResult — not a type of its own, just ExemplarIdentification & { identification }; per the convention in types.ts, endpoint-specific fields get intersected at the call site rather than added to the shared type

app/webpack/taxa/show/components/app.tsx
sundisc: Good idea.. can't drop it entirely though because `PlaceChooserContainer` is different for the browse_photos `TaxonPageHeader` vs. the details page `TaxonPageHeader`

app/webpack/taxa/show/components/app.tsx
sundisc: Whoops!

sundisc:
@caseyhofford Thanks for the feedback! Implemented everything https://github.com/inaturalist/inaturalist/pull/4981/changes/8a79322e5cc3b8dee560fabcc2f13de7b3a44f2a

app/webpack/taxa/show/components/app.tsx
caseyhofford: Ah, I missed that. Thanks for making this change; this simplifies things well!

app/webpack/taxa/show/components/identifications_tab.tsx
caseyhofford: Thanks!

caseyhofford (approved):
Thanks for implementing those changes! Looks great.

PR #4983, opened 2026-07-17
feat: INatAPIService better handles v2 queries, update more calls to v2, WEB-1105

pleary:
* Adds the `rison-rb` gem for Rison-encoding `fields` parameters for API V2 calls
* Updates `notification_counts` JS API calls to use V2
* Updates `INatAPIService` to accept a `v2` parameter to switch between v1 and v2 for server-side API requests. It also properly formats the `fields` parameter depending on its type. If a URL is too long for GET requests, it will fallback to making a POST with a header `X-HTTP-Method-Override=GET` that API V2 interprets as a call to the GET endpoint
* Updates all server-side geoip_lookup API calls to use V2 (depends on https://github.com/inaturalist/iNaturalistAPI/pull/651)
* Updates the taxon show server payload API call to use V2


app/assets/javascripts/application.js.erb
```diff
@@ -1027,7 +1027,7 @@ function getMessagesCount() {
 function getHeaderCounts() {
   var apiToken = $( "meta[name='inaturalist-api-token']" ).attr( "content" );
   $.ajax({
-    url: "<%= CONFIG.public_node_api_url || CONFIG.node_api_url %>/users/notification_counts",
+    url: "<%= CONFIG.node_api_url_v2 %>/users/notification_counts",
```
sundisc, line 1030: If this is something that runs on the client browser, do we need to use `public_node_api_url`?

tools/development_tools/taxon_importer/taxon_importer.rb
```diff
@@ -50,22 +51,27 @@ def import( options = {} )
...
+      @results ||= []
+      return @results
     end
+
+    @results ||= INatAPIService.get(
+      "/taxa/#{full_ids.join( ',' )}",
+      { fields: REQUESTED_FIELDS },
+      v2: true,
+      endpoint: PRODUCTION_API_ENDPOINT
+    ).results
```
sundisc: It looks like `INatAPIService.get` could return nil so do we need a nil check before results here and on line 74?

tools/development_tools/taxon_importer/taxon_importer.rb
```diff
@@ -2,6 +2,7 @@
 
 class TaxonImporter
   REQUESTED_FIELDS = %w(id name rank rank_level iconic_taxon_id is_active ancestry).freeze
+  PRODUCTION_API_ENDPOINT = "https://api.inaturalist.org/v2"
```
sundisc, line 5: Should we use INatAPIService::ENDPOINT_V2 instead?

app/models/user.rb
```diff
@@ -1009,29 +1009,28 @@ def set_uri
...
-          lat_lon_acc_admin_level = Place::STATE_LEVEL
-        else
-          # probably just know the country
-          lat_lon_acc_admin_level = Place::COUNTRY_LEVEL
-        end
+    geoip_response = INatAPIService.geoip_lookup( { ip: last_ip, fields: "all" }, v2: true )
+    geoip_result = geoip_response&.results&.first
+    if geoip_result && !geoip_result["country"].blank?
+      ll = geoip_result["ll"]
+      latitude = ll[0]
```
sundisc, line 1023: Is it possible for `ll` to be nil on a `geoip_result`? If so we probably want a guard here.

sundisc (commented):
Looking good, just had a few questions about small things.

app/assets/javascripts/application.js.erb
pleary: `CONFIG.public_node_api_url` is now a legacy reference, and has no current use. Given that, I decided to remove all references to it. If we want something similar going forward, we can implement it in a more central way that doesn't require so many conditionals in views and JS

tools/development_tools/taxon_importer/taxon_importer.rb
pleary: I've added some error catching on blank API responses

tools/development_tools/taxon_importer/taxon_importer.rb
pleary: This particular tool is for importing data from our production API into a development environment. The previous version of the tool also hard-coded a production API URL here, which I am just replicating here. If we used `INatAPIService::ENDPOINT_V2`, that would typically point to a local API V2, thus would not have the production data we want to import. We could update this to accept an API URL, but given this is a likely rarely-used tool for development, I'm going to hold off on adding new features to it for now

app/models/user.rb
pleary: based on [the recent PR that defines](https://github.com/inaturalist/iNaturalistAPI/pull/651/changes#diff-a54a435f7da9a56122a1f0f9acf29bb9c5c7e71553d62a08e569bffe607ab249R18) this endpoint in V2, no it won't be possible for `ll` to be nil. If it is not an array of numbers as the response schema defines it to be, that endpoint will raise an error instead. So we shouldn't need a guard on `ll` here

sundisc (approved):
Thanks for clearing those up.. looks great ✅

PR #4987, opened 2026-07-20
feat: Reapply observation details responsiveness changes (with bugfixes and test-group), WEB-983

sundisc:
Reapplies observation details responsive changes behind a gate: unless the user is in the `responsive-obs-detail` or `responsive-global` test groups, they will continue to see the original page's styling. This was done by:
 - Preserving all original components by adding the `_legacy` filename and using the new [gated_component.tsx](https://github.com/inaturalist/inaturalist/pull/4987/changes#diff-925e6293cc467fa0d61d071b513524e0186d2026e7e243a1147123ed0c13e6db) to switch between them
 - Preserving original styling in [show_legacy.scss](https://github.com/inaturalist/inaturalist/pull/4987/changes#diff-bc4573fad7791d587c67d464d947891998d7b99c1d47684aed2ec0bdc069c694) and switching based on the test group inclusion in [show.html.haml](https://github.com/inaturalist/inaturalist/pull/4987/changes#diff-a07afee5daaaaaa76450c59d17c4a6de5b5538fe5f0c713375dd590faa5b4ec4)
 - Adding responsive versions of existing shared CSS files such as `activity_item.scss` in [‎app/assets/stylesheets/shared/responsive](https://github.com/inaturalist/inaturalist/pull/4987/changes#diff-22384fa6c0bf9780685ffdf097284213374b91094c9cca4b28960dab527cdecb) to prevent responsiveness changes leaking into other pages

Beyond that, additional changes were made to the responsive version's style:
- Fixing DQA sizing bug
- Fixing activity item flex issue
- Showing the compare modal until a smaller breakpoint
- Making the about modal responsive and showing the button at all breakpoints
- Wrapping long observation field values


pleary (approved):
This is all looking great to me. I tested with the global and obs-details test groups, which behaved the same. I noticed a few minor potential styling issues as I was testing. They don't need to be incorporated in this PR, but I'll mention them for consideration for ongoing polishing of the responsive view.

This all seems OK with me to merge

---
Image

When making an ID that is an ancestor of the community taxon, we show a modal asking the user the nature of the disagreement. Very long names in that modal are getting cut off instead of wrapping. Notice I'm using the multiple common names feature here for testing (Account -> Content & Display -> Common Name Lexicons), which can result in some very long name displays.

---
Image

Clicking on the bars in the community ID section open a popup with the referenced taxon's taxonomy. That can get cut off on smaller displays. We might update that to scroll horizontally like has been done in the About Algorithm Scoring section

---

Existing:
Image

New:
Image


One comment we got about the initial release was the location of the action buttons in the activity section were wrapping into new lines. Around width 1000px, before the large breakpoint, the existing design ensures the buttons and taxon names are rendered next to one another, in the new design they wrap. This is one of the only a few visual differences in the responsive view at this page width. It might be nice to preserve the old behavior for visual consistency until the large breakpoint. The name can get a little squished, but it may be preferred by people used to looking for the buttons in that location.

I'll note that with the existing design around this width, the main content margin is a bit narrower, so the Activity panels can be a little wider. That may mean if we try to keep the buttons on the same line, the responsive version will have even less room for the name due to its wider margins, which might make it undesirable to keep them all on the same line. But worth considering as it's one of the only visual differences I'm noticing around this page width  

---

Image

Around that same pre-large breakpoint width, the delete button for annotations can wrap onto new line. If there's room enough to prevent wrapping, that might be preferred. Aside from the intentional changes like the More From User carousel, this and the Activity Items buttons above are the only other visual difference I noticed between the existing page and the responsive page before the large breakpoint

sundisc:
Thanks @pleary! I made fixes for these 4 things.

pleary:
Fixes for these few style issues are looking good to me. Approval still stands - OK with me to merge anytime

PR #4988, opened 2026-07-20
fix: Prevent header from collapsing, WEB-1079

sundisc:
When the user is no in the responsive test group, the search bar was allowed to shrink to 0. Removing the min-width keeps it open.

Before:
https://github.com/user-attachments/assets/20a3808b-2af4-4934-a8ff-4669ca59d393

After:
https://github.com/user-attachments/assets/1b2f1b6d-709c-499e-818d-f22bcb2a0eaa


pleary (approved):
Looks good - thanks for the fix!

PR #4999, opened 2026-07-28
Block requests to throttled endpoints in MetaService#fetch_request_uri

caseyhofford:
I think this might address the underlying cause of persistent throttling by blocking all requests to throttled endpoints within the `fetch_request_uri` method.

### DIFF
--- a/lib/meta_service.rb
+++ b/lib/meta_service.rb
@@ -93,18 +93,7 @@ class MetaService
         return Nokogiri::XML( api_endpoint_cache.response )
       end
 
-      # Providers like Wikipedia throttle by user agent, so being throttled for one request means
-      # every other request to this endpoint will be throttled too. Back off from the endpoint as a
-      # whole rather than sending requests we expect to fail: each failure extends the throttle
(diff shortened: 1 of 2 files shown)


lib/meta_service.rb
```diff
@@ -92,6 +92,21 @@ def self.fetch_request_uri( options = {} )
 
         return Nokogiri::XML( api_endpoint_cache.response )
       end
+
+      # Providers like Wikipedia throttle by user agent, so being throttled for one request means
+      # every other request to this endpoint will be throttled too. Back off from the endpoint as a
+      # whole rather than sending requests we expect to fail: each failure extends the throttle
+      # window and overwrites whatever this URL had cached with the throttling message.
+      if options[:api_endpoint].recently_throttled?
+        Logstasher.write_hash(
```
caseyhofford: Might be an excessive amount of logs, but it would make it easy to check how many requests are being blocked.

sundisc (approved):
Makes sense that we'd want to throttle endpoint-wide here.. looks good to me!

PR #5004, opened 2026-07-29
Display last good cache

caseyhofford:
When the ApiEndpoint is throttled, return the last successful ApiEndpointCache, if it exists.

### DIFF
--- a/lib/meta_service.rb
+++ b/lib/meta_service.rb
@@ -88,14 +88,11 @@ class MetaService
           status_code: 200
         )
 
-        return_response( api_endpoint_cache.response, options )
+        return return_response( api_endpoint_cache.response, options )
       end
 
       if api_endpoint_cache.cached? && !options[:force_update]
-        # A throttled response is not a valid API response; treat it as if there
-        # was no response rather than parsing the throttling message.
-
-        return_response( api_endpoint_cache.response, options )
+        return return_response( api_endpoint_cache.response, options )
       end
     end
     response = nil
@@ -125,7 +122,7 @@ class MetaService
       return
     end
 
-    return_response( response, options )
+    return_response( response.body, options )
   end
 
   def self.fetch_with_redirects( options, attempts = 3 )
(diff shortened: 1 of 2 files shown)


sundisc (approved):
Looks good!

lib/meta_service.rb
```diff
@@ -81,20 +81,24 @@ def self.fetch_request_uri( options = {} )
...
-        # A throttled response is not a valid API response; treat it as if there
-        # was no response rather than parsing the throttling message.
-        return if api_endpoint_cache.throttled?
+      if options[:api_endpoint].recently_throttled?
+        return if options[:force_update]
 
-        if options[:raw_response]
-          return api_endpoint_cache.response
-        end
+        api_endpoint_cache = ApiEndpointCache.find_by(
```
pleary: Quick question here - we already looked up `api_endpoint_cache` above and it will be the record matching the api_endpoint and request_url. We can look at its `status_code` and `success` attributes and see if they are 200 and true. Is this separate query needed, or could we just look at the already fetched `api_endpoint_cache`?

caseyhofford: Good point. Since we use `find_or_create_by` w/ endpoint and URL and just update this record, we can only have one record for each URL in the DB at a time. I was imagining we could have multiple caches for different status_codes, which would allow us to display a good response even when a previous request for this ApiEndpointCache had returned a non-200. This would require some additional changes, though. I will go with your original suggestion and track a ticket for storing ApiEndpointCache per status code. [WEB-1164: Store an ApiEndpointCache per status code](https://linear.app/inaturalist/issue/WEB-1164/store-an-apiendpointcache-per-status-code)

caseyhofford: Done in aac8370a1d2618784dd75c1fa9ec8fe4d1c15ee4 and c5362a14d2e1cd532de3ef1b13c0cb9a3ec2f96e

pleary (approved):
Thanks - looks good to me

PR #5005, opened 2026-07-29
Count impressions with no ip address

caseyhofford:
This is a small change that could increase counted impressions if requests ever come through without IP address, however this seems unlikely, and I did not find any examples in the logs.

It also fixes a bug in passing user_ids correctly to `base_impression_options`.


pleary (approved):
Looks good to me

PR #5006, opened 2026-07-30
feat: use api v2 for lifelists, WEB-1060

pleary:
Depends on the API additions in https://github.com/inaturalist/iNaturalistAPI/pull/657

### DIFF
--- a/app/assets/javascripts/jquery/plugins/inat/place_autocomplete.js.erb
+++ b/app/assets/javascripts/jquery/plugins/inat/place_autocomplete.js.erb
@@ -1,9 +1,14 @@
 /* global _ */
 /* global iNatModels */
+
 $.fn.placeAutocomplete = function ( opts ) {
   var options = opts || { };
   if ( !options.idEl ) { return; }
   var field = this;
+  var apiBaseURL = "<%= CONFIG.node_api_url %>";
+  if ( options.useAPIv2 ) {
+    apiBaseURL = apiBaseURL.replace( "/v1", "/v2" );
+  }
 
   field.template = function ( item ) {
     var wrapperDiv = $( "<div/>" ).addClass( "ac" ).attr( "id", item.id );
@@ -20,22 +25,26 @@ $.fn.placeAutocomplete = function ( opts ) {
   field.genericAutocomplete( _.extend( options, {
     extraClass: "place",
     source: function ( request, response ) {
+      var params = {
+        q: request.term,
+        per_page: 10,
+        sources: "places",
+        fields: options.fields || "all"
+      };
       $.ajax( {
-        url: "<%= CONFIG.node_api_url %>/search",
+        url: apiBaseURL + "/search",
         dataType: "jsonp",
         cache: true,
         jsonpCallback: "placeAutocompleteCallback",
-        data: {
-          q: request.term,
-          per_page: 10,
-          sources: "places"
-        },
+        data: params,
         success: function ( data ) {
           var results = data.results || [];
           response( _.map( results, function ( r ) {
-            return new iNatModels.Place( Object.assign( {}, r.record, {
-              title: r.record.display_name
-            } ) );
+            var place = ( options.useAPIv2 ? r.place : r.record );
+            return new iNatModels.Place( {
+              ...place,
+              title: place.display_name
+            } );
           } ) );
         }
       } );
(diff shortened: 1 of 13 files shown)


app/webpack/lifelists/show/webpack-entry.js
```diff
@@ -32,6 +33,18 @@ if ( !_.isEmpty( PREFERRED_PLACE ) ) {
...
 }
 
+const element = document.querySelector( "meta[name=\"config:inaturalist_api_url\"]" );
+const defaultApiUrl = element && element.getAttribute( "content" );
+if ( defaultApiUrl ) {
+  sharedStore.dispatch( setConfig( {
+    testingApiV2: true
+  } ) );
+  inatjs.setConfig( {
+    apiURL: defaultApiUrl.replace( "/v1", "/v2" ),
```
caseyhofford: Same as above? Can we use `CONFIG.node_api_url_v2` from `config/application.rb:166` here?

app/assets/javascripts/jquery/plugins/inat/place_autocomplete.js.erb
```diff
@@ -1,9 +1,14 @@
 /* global _ */
 /* global iNatModels */
+
 $.fn.placeAutocomplete = function ( opts ) {
   var options = opts || { };
   if ( !options.idEl ) { return; }
   var field = this;
+  var apiBaseURL = "<%= CONFIG.node_api_url %>";
+  if ( options.useAPIv2 ) {
+    apiBaseURL = apiBaseURL.replace( "/v1", "/v2" );
```
caseyhofford: Can we use `CONFIG.node_api_url_v2` from `config/application.rb:166` here to DRY up the `v1`->`v2` replacement logic?

app/webpack/lifelists/show/reducers/lifelist.js
```diff
@@ -347,7 +352,7 @@ export function updateSpeciesPlaceSearch( reload = false ) {
         force: true,
         searchParams: {
           user_id: lifelist.user.id,
-          place_id: lifelist.speciesPlaceFilter.id,
+          place_id: lifelist.speciesPlaceFilter.uuid,
```
caseyhofford, line 356: Does the ES index for taxon support UUID currently? The index at `app/es_indices/taxon_index.rb:115` looks like it currently is expecting an integer ID, but I'm not sure if UUIDs are mapped to these ID fields elsewhere.

app/webpack/lifelists/show/reducers/lifelist.js
caseyhofford, line 125: Do we need this to be UUID?

caseyhofford (commented):
Looking good! I think the UUID for place_id is not supported in the API yet, but could be wrong on that.

app/webpack/lifelists/show/webpack-entry.js
pleary: I was planning on changing these globally in a future commit since this same pattern is used in a lot of webpack-entry files. I added one approach here in this branch to configure the v2 config via the new config `enableApiV2` method. `CONFIG.node_api_url_v2` is only accessible rails-side and is passed to the frontend via a meta tag, with a fallback to this same replacement approach if for some reason the v2 URL isn't in a meta tag. Does that look OK to you? If so I can apply the same to the other instances of this approach in a future commit

app/assets/javascripts/jquery/plugins/inat/place_autocomplete.js.erb
pleary: Thanks, I made that change

app/webpack/lifelists/show/reducers/lifelist.js
pleary: The ES index only indexes place IDs. These parameters will be sent to the API which expects a UUID. The API will look up the place and ultimately search ES via ID. This is working as expected for me

app/webpack/lifelists/show/reducers/lifelist.js
pleary: Thanks - we need the ID here, but we also need the UUID which I added

app/webpack/lifelists/show/webpack-entry.js
caseyhofford: Yeah, that makes sense, thanks for the explanation @pleary!

app/webpack/lifelists/show/reducers/lifelist.js
caseyhofford: Sounds good!

caseyhofford (approved):
Looks great, thanks for making those changes @pleary!

PR #5014, opened 2026-08-03
feat: Responsive dashboard, WEB-1012

sundisc:
Makes the dashboard (`users/dashboard`) and the user subnav responsive behind a gate: unless the user is in the `responsive-global` test group, they will continue to see the original page's styling and markup. This was done by:
   - Gating on the backend in [application_controller.rb](https://github.com/inaturalist/inaturalist/pull/5014/files#diff-766c34fd6533171eaf54300c153f89d6002c35c02cfc9c5b219251f85180ad07), where `responsive-global` membership sets `@responsive` **and** `request.variant = :responsive` (alongside the existing `responsive-header` gate, which only sets `@responsive`)
   - Preserving all original templates untouched and adding variant-suffixed siblings — [dashboard.html+responsive.haml](https://github.com/inaturalist/inaturalist/pull/5014/files#diff-45fe42a0c23a00b41cc54a08d743e9f282c90c12645f487c78fd289d3de2edc9), [_subnav_bootstrap.html+responsive.haml](https://github.com/inaturalist/inaturalist/pull/5014/files#diff-076de4c056d97112fe8d4de05c110fdcfc5371f0aa72c4c07c2d8bd2683f541d), [_dashboard_updates.html+responsive.haml](https://github.com/inaturalist/inaturalist/pull/5014/files#diff-1e7ebbdc5c776721a8472411d4b83f3c5a5e1c4f466c3dabeadc2743d55fdf55), `_by_login_header_bootstrap.html+responsive.haml`, plus the observations/jobs dashboard components. Rails only resolves these when the variant is set, so there is no runtime branch to get wrong and no way for the new markup to reach anyone outside the group
   - Scoping all new styling to [dashboard-responsive.scss](https://github.com/inaturalist/inaturalist/pull/5014/files#diff-89d1f283c7c7aedb5194cdecd0d629b4a7765d3cc7066807336c0bf1aa706b6d), which is only linked from the `+responsive` template

   The only shared code that changed is additive and defaults to the current behavior, so legacy renders are unchanged:
   - `user_image` in [application_helper.rb](https://github.com/inaturalist/inaturalist/pull/5014/files#diff-5f3fc5e3977d242572aa1d08551f5eb557de0ccaff30370838ee9df5386ea0da) previously hardcoded inline `width` / `height` / `max-width`. It now accepts an opt-in `no_inline_dimensions:` flag (off by default) so the responsive stylesheet can size avatars via CSS, and always emits an additive `user_image--<size>` class. `update_image` threads the flag through via `options.slice`
   - `_favorites` / `_followees` take an `item_class ||=` local defaulting to the existing `col-md-3 col-sm-3`, letting the responsive views opt into CSS grid instead of the Bootstrap grid
   - `_update_created_observations` wraps its show more/less text in a `.btn_expand_label` span so the label can be hidden on narrow screens

  Beyond the gate, the responsive version modernizes the page:
  - **Bootstrap grid removed** — the dashboard is a CSS grid/flexbox layout (`.dashboard-columns` / `.dashboard-main`, 2x2 card grid) that stacks on mobile; Bootstrap UI components are retained
  - **Mobile nav** — the user subnav tab bar collapses below `$md` into the shared `TabDrawer`, rendered by a new `users-subnav` webpack entry ([webpack-entry.js](https://github.com/inaturalist/inaturalist/pull/5014/files#diff-95c9fb87590bb781d26eab335b7bd09c2ecc301472b187c7d6d3bfd3a1ff5f01), registered in [webpack.config.js](https://github.com/inaturalist/inaturalist/pull/5014/files#diff-c49a22733230fd83e34ff3bc78014ee352ad46ce10c7c502dd2a7c3982adb0a5)). Subnav items and the active tab are computed once in the partial and passed to both the drawer and the legacy `<ul>`
  - **Timeline item redesign** — Rounded edges better match current mobile designs
 
Screenshot 2026-08-03 at 3 15 49 PM


pleary (changes requested):
This is all looking great to me. I'm hitting the same local development sprockets problem I saw with the `inat-wordmark.webp` image, but this time with `dashboard-responsive.scss` - sprockets is thinking `responsive` is a fingerprint, sees that `dashboard.scss` exists, and see that the fingerprint of `dashboard.scss` is not `responsive`, ultimately rendering a 404. Changing that file name to use an underscore, e.g. `dashboard_responsive.scss` fixes the problem.

I'm setting the review to `Request Changes` for that one small change since it breaks the responsive dashboard styles in a development environment. But otherwise, everything is looking great, and after that file rename this is all OK with me to merge

sundisc:
> This is all looking great to me. I'm hitting the same local development sprockets problem I saw with the `inat-wordmark.webp` image, but this time with `dashboard-responsive.scss` - sprockets is thinking `responsive` is a fingerprint, sees that `dashboard.scss` exists, and see that the fingerprint of `dashboard.scss` is not `responsive`, ultimately rendering a 404. Changing that file name to use an underscore, e.g. `dashboard_responsive.scss` fixes the problem.
> 
> I'm setting the review to `Request Changes` for that one small change since it breaks the responsive dashboard styles in a development environment. But otherwise, everything is looking great, and after that file rename this is all OK with me to merge

Ah, right. I'll try to remember not to use hyphens next time.. should work now @pleary. Thanks for the review!

pleary (approved):
Thanks!

PR #5015, opened 2026-08-04
fix: Use legacy components on Identify, WEB-1183

sundisc:
https://github.com/inaturalist/inaturalist/pull/4987 forgot to update the `ActivityItem` and `Annotations` to import from the legacy file for Identify, which caused some incorrect styles.

**Before**
Screenshot 2026-08-04 at 2 42 20 PM
Screenshot 2026-08-04 at 2 42 28 PM

**After**
Screenshot 2026-08-04 at 2 41 07 PM
Screenshot 2026-08-04 at 2 41 13 PM


pleary (approved):
Fix looks good to me

PR #5019, opened 2026-08-11
fix: Prevent secondary name from escaping parenthesis: WEB-1065

sundisc:
Parenthesis are CSS `::before`/`::after` with `display: contents`. Chrome misplaces those when React re-renders in place. Changed to display: inline to prevent escape.


caseyhofford (approved):
Nice work, that's a weird one!

pleary (approved):
This is working for me. Thanks!

PR #5022, opened 2026-08-13
feat: Add dismiss button to test group join toggle, WEB-1224

sundisc:
- Show dismiss button on test group toggles
- Show message about how to leave test group via settings after dismissing opt-out toggle that disappears after navigating away
- Set localstorage on dismiss item to re-show either opt-in or opt-out test group toggle after 5 days
- Add ability for any user to leave responsiveness test group and leave feedback via settings

Opt-in:
Screenshot 2026-08-13 at 4 45 07 PM

Opt-out:
Screenshot 2026-08-13 at 4 44 31 PM

Message after opt-out dismissal:
Screenshot 2026-08-13 at 4 44 41 PM

User setting:
Screenshot 2026-08-13 at 4 44 23 PM

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -123,6 +123,7 @@ class User < ApplicationRecord
   preference :suggestions_sort, :string
   preference :taxon_page_tab, :string
   preference :taxon_page_ancestors_shown, :boolean, default: false
+  preference :hide_responsive_global_toggle, :boolean, default: false
 
   NOTIFICATION_PREFERENCES = %w(
     comment_email_notification
--- a/app/assets/stylesheets/bootstrap_bundle.scss
+++ b/app/assets/stylesheets/bootstrap_bundle.scss
@@ -402,4 +402,10 @@ body>div>.popover .label {
       line-height: 18px;
     }
   }
+
+  // Flex would make the inline link its own flex item and drop the spaces
+  // around it, so prose banners stay block level.
+  &.TestGroupBanner--text {
+    display: block;
+  }
 }
--- a/app/assets/stylesheets/bootstrap_bundle.scss
+++ b/app/assets/stylesheets/bootstrap_bundle.scss
@@ -353,25 +353,19 @@ body>div>.popover .label {
     position: absolute;
     top: -10px;
     right: -10px;
-    float: none;
-    width: 22px;
-    height: 22px;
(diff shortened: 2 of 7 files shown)


app/models/user.rb
```diff
@@ -123,6 +123,7 @@ def set_uuid
   preference :suggestions_sort, :string
   preference :taxon_page_tab, :string
   preference :taxon_page_ancestors_shown, :boolean, default: false
+  preference :hide_responsive_global_toggle, :boolean, default: false
```
caseyhofford, line 126: Is this preference being used? It looks like the toggle state is being stored in the browser session at https://github.com/inaturalist/inaturalist/pull/5022/changes#diff-4676ac35877e7f94c108b3670a7ebe83e5ad5b511dca52a38f87cc81443dba8fR27, and I don't see any calls to set this preference.

caseyhofford (commented):
Looking good, just want to confirm if the added preference is being used.

app/models/user.rb
pleary: Yes it is. I made this change to the original PR to save the dismissal state in a user preference. `updateSession` [calls UsersController::update_session](https://github.com/inaturalist/inaturalist/blob/main/app/assets/javascripts/application.js.erb#L1270), which in turn updates the active record session, but also checks to see if there is a [preference with the same name](https://github.com/inaturalist/inaturalist/blob/main/app/controllers/users_controller.rb#L866). In this case there is, the newly added preference, and that will ensure we keep the banner dismissed for this user in all sessions on all devices indefinitely.

caseyhofford: Nice! That makes sense, thanks for the explanation.

caseyhofford (approved):
Looks good!

pleary (approved):

pleary:
Noting here that I made a commit after this PR was made which changes the behavior described in this PR description. This no longer uses localstorage at all, and instead stores for dismissal in a user preference. The banner will no longer reappear in 5 days and instead it will remain suppressed indefinitely. The user can still opt in/out via their user settings. We can use announcements or other means to attempt to reengage users that have dismissed the banner if we're not getting the feedback we want, and perhaps to remind people of the setting since the plan is for the responsive mode to be the default and only mode for all users after not too long.

PR #5024, opened 2026-08-14
raise specific rails deprecation warnings in CI

caseyhofford:
This PR will cause CI to fail for specific deprecation warnings that have been fixed and added to the `config.active_support.disallowed_deprecation_warnings` array. We should continue to add to this list if we identify new deprecation warnings that need to be disallowed to unblock rails upgrade. This will prevent us from introducing new breaking changes to the rails upgrade pathway.

This PR also fixes existing deprecations that would otherwise throw errors in CI with these changes.

We also log deprecation warnings from production to capture all deprecation warnings in the disallowed list, not just those with test coverage.

### DIFF
--- a/app/controllers/places_controller.rb
+++ b/app/controllers/places_controller.rb
@@ -450,7 +450,7 @@ class PlacesController < ApplicationController
       format.html { redirect_to @place }
       format.json do
         @taxa.map! do | taxon |
-          taxon.html = render_to_string( partial: "taxa/taxon.html.erb",
+          taxon.html = render_to_string( partial: "taxa/taxon", formats: [:html],
             object: taxon, locals: {
               image_options: { size: "small" },
               link_image: true,
--- a/app/controllers/projects_controller.rb
+++ b/app/controllers/projects_controller.rb
@@ -541,7 +541,7 @@ class ProjectsController < ApplicationController
         render :partial => "projects/observed_taxa_count_widget"
       end
       format.widget do
-        render :js => render_to_string(:partial => "projects/observed_taxa_count_widget.js.erb")
+        render :js => render_to_string(:partial => "projects/observed_taxa_count_widget", :formats => [:js])
       end
     end
   end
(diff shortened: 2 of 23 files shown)


app/controllers/shared/lists_module.rb
pleary, line 344: This endpoint (`/lists/:id/taxa`) isn't functional and we may as well remove it. That doesn't have to happen in this branch/PR, but if it doesn't it might be nice to make an issue for methods that can be removed. I'm noticing a few testing some of these changes

app/controllers/places_controller.rb
pleary, line 432: This is another endpoint that isn't functional due to invalid parameters being sent to `paginate` - it could be removed with its associated route

app/controllers/projects_controller.rb
```diff
@@ -611,7 +611,7 @@ def list
 
     # Load tips HTML
     @taxa.map! do |taxon|
```
pleary, line 613: This endpoint doesn't appear to even be associated with a route, so I'm not sure its accessible. If it were, it raises an error here on `map!`. This is another endpoint that could be removed

pleary (approved):
Looks good to me. I left some comments about touched endpoints that could instead be removed entirely. Everything else appears to be working well.

caseyhofford:
Thanks for the review @pleary! I've filed a ticket to remove the deprecated controller actions/routes here: https://linear.app/inaturalist/issue/WEB-1238/remove-deprecated-endpoints, will do in a quick follow up PR.

app/controllers/projects_controller.rb
caseyhofford: https://linear.app/inaturalist/issue/WEB-1238/remove-deprecated-endpoints

app/controllers/places_controller.rb
caseyhofford: https://linear.app/inaturalist/issue/WEB-1238/remove-deprecated-endpoints

app/controllers/shared/lists_module.rb
caseyhofford: https://linear.app/inaturalist/issue/WEB-1238/remove-deprecated-endpoints

PR #5026, opened 2026-08-18
Allow indexing of TaxonPhoto embeddings into Qdrant, WEB-1078

pleary:
* Adds an ActsAsQdrantModel that was inspired by parts of [elasticsearch-model](https://github.com/elastic/elasticsearch-rails/tree/main/elasticsearch-model) and our own ActsAsElasticsearch wrapper
  * Adds helper methods to manage Qdrant collections
  * Adds helper methods to manage indexing Active Record instances as points in Qdrant collections
  * Models can enable lifecycle callbacks to automatically index or remove from the Qdrant index as models are created, updated, and destroyed
  * If Qdrant is not defined in the config all ActsAsQdrantModel methods will not throw errors, rather they just don't do anything
  * Adds support for testing including collection recreation and converting "after_commit" callbacks to work well with transactional fixtures, like we already have for Elasticsearch 
* Updates rspec to fix a bug where methods with keyword arguments were not compatible with `and_call_original`


app/qdrant_collections/taxon_photo_collection.rb
```diff
@@ -0,0 +1,86 @@
...
+      next unless payload["photo_id"] == taxon_photo.photo_id
+
+      indexed_file_updated_at = payload["photo_file_updated_at"]
+      if indexed_file_updated_at.present?
+        indexed_file_updated_at = Time.parse( indexed_file_updated_at )
+      end
+      next unless indexed_file_updated_at&.floor == taxon_photo.photo&.file_updated_at&.floor
+
+      true
+    end
```
caseyhofford: Let's indent this whole block, this still passes rubocop and looks easier to read to me.
 
```suggestion
    batch.select {| taxon_photo | taxon_photo&.taxon&.is_active? }.
      reject do | taxon_photo |
        indexed_doc = existing_indexed_documents[taxon_photo.id]
        # if there is an existing indexed document with the same taxon ancestors,
        # and the same photo and version, no need to reindex
        next unless indexed_doc

        payload = indexed_doc["payload"]
        next unless payload["ancestor_ids"] && taxon_photo&.taxon &&
          payload["ancestor_ids"].sort == taxon_photo.taxon.self_and_ancestor_ids&.sort
        next unless payload["photo_id"] == taxon_photo.photo_id

        indexed_file_updated_at = payload["photo_file_updated_at"]
        if indexed_file_updated_at.present?
          indexed_file_updated_at = Time.parse( indexed_file_updated_at )
        end
        next unless indexed_file_updated_at&.floor == taxon_photo.photo&.file_updated_at&.floor

        true
      end
```

app/qdrant_collections/taxon_photo_collection.rb
```diff
@@ -0,0 +1,86 @@
...
+      # and the same photo and version, no need to reindex
+      next unless indexed_doc
+
+      payload = indexed_doc["payload"]
+      next unless payload["ancestor_ids"] && taxon_photo&.taxon &&
+        payload["ancestor_ids"].sort == taxon_photo.taxon.self_and_ancestor_ids&.sort
+      next unless payload["photo_id"] == taxon_photo.photo_id
+
+      indexed_file_updated_at = payload["photo_file_updated_at"]
+      if indexed_file_updated_at.present?
```
caseyhofford: Might want safe navigation on this in case "photo_file_updated_at" is missing or nil.
```suggestion
      if indexed_file_updated_at&.present?
```

app/qdrant_collections/taxon_photo_collection.rb
```diff
@@ -0,0 +1,86 @@
+# frozen_string_literal: true
+
+class TaxonPhoto < ApplicationRecord
+  scope :load_for_qdrant_index, -> { load_for_index }
```
caseyhofford, line 4: NIT: Will these scopes always be just aliases of `load_for_index` from `app/es_indices/taxon_photo_index.rb`? If they will always be the same, I'd argue we shouldn't add a new scope and maybe move the scope to the shared `app/models/taxon_photo.rb,` but if they are sometimes different, maybe it's worth duplicating the scope definition here to be more explicit and make it easier to quickly find out what this scope represents.

spec/models/user_spec.rb
```diff
@@ -1500,15 +1500,30 @@
     end
 
     it "should queue moving photos if needed" do
```
caseyhofford, line 1502: Are these changes intended to be in this PR? They don't seem closely related to me.

caseyhofford (approved):
Looks great @pleary, a few small nitpicks, but nothing blocking.

app/qdrant_collections/taxon_photo_collection.rb
pleary: No - a model's qdrant collections will not always have the same data loading requirements as their Elasticsearch index if it exists. In this case the TaxonPhoto Qdrant collection is pretty much a straight up replacement for the TaxonPhoto Elasticsearch index, needing pretty much the same data, so this works. Eventually we'll remove the TaxonPhoto Elasticsearch index and we'll need to move the scope over to this file. I'd prefer to leave them separate so the ActsAsQdrant code can check for the Qdrant-specific scope, even if it sometimes happens to be the same as the ActsAsElasticsearch-specific scope

spec/models/user_spec.rb
pleary: Yes it was intended. This was an indirect side-effect of updating rspec, which was needed to run I believe this spec assertion `expect( TaxonPhoto.__qdrant__.client.collections ).to receive( :create ).and_call_original`. The previous version of rspec could not use `and_call_original` with methods that accept keyword parameters. After updating rspec, these user specs were failing (and maybe should have been failing before) due to some wonky use of `spy`. So - not directly related to the code at hand, but the rspec update caused them to be necessary in this PR

app/qdrant_collections/taxon_photo_collection.rb
pleary: I added indenting in 56e8d8ce955114a2c18438eeff00f6e14091e8dd

app/qdrant_collections/taxon_photo_collection.rb
pleary: Added in 56e8d8ce955114a2c18438eeff00f6e14091e8dd

PR #5027, opened 2026-08-18
fix: ensure DwC-A EML uses proper license case for search URL, WEB-1100

pleary:


caseyhofford (approved):
Looks good! Thank you.

PR #5030, opened 2026-08-24
Web 1228 upgrade devise from 480 to 494 mediumhigh risk

caseyhofford:
Devise gem upgrade in preparation for rails upgrade.
- Upgrades devise from `4.8.0` to `4.9.4`.
- Adds a lot of specs around devise behavior, confirming that there are no changes to response codes.
- Explicitly sets defaults for response codes to match existing behavior. Technically unnecessary, but provides more certainty that status codes will not change.

## Testing

This is currently up for testing on staging4.

- [ ] Test web login
- [ ] Test generating a JWT
- [ ] Test authenticated API request


pleary (approved):
Looks good to me

PR #5033, opened 2026-08-25
fix: upgrade billboard.js to 3.18.0 (CVE-2026-1513)

anupamme:
## Summary
Upgrade billboard.js from 3.15.1 to 3.18.0 to fix CVE-2026-1513.

## Vulnerability
| Field | Value |
|-------|-------|
| **ID** | CVE-2026-1513 |
| **Severity** | HIGH |
| **Scanner** | trivy |
| **Rule** | `CVE-2026-1513` |
| **File** | `package-lock.json` (dependency: `billboard.js`) |
| **Assessment** | Likely exploitable |

**Description**: billboard.js is vulnerable to XSS during chart option binding

## Evidence

**Scanner confirmation**: trivy rule `CVE-2026-1513` flagged this pattern.

## Changes
- `package.json`
- `package-lock.json`

## Behavior Preservation
The change is scoped to 2 files on the vulnerable path; it only tightens handling of untrusted input and leaves valid inputs unaffected.

---
*This change addresses a pattern flagged by static analysis. The code path handles user-influenced input and the fix reduces the attack surface against both manual and automated exploitation.*

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/app/webpack/taxa/show/components/charts.test.jsx
+++ b/app/webpack/taxa/show/components/charts.test.jsx
@@ -1,0 +1,126 @@
+import React from "react";
+import $ from "jquery";
+import _ from "lodash";
+import { render } from "@testing-library/react";
+import Charts from "./charts";
+
+global.$ = $;
+global.jQuery = $;
+
+// jsdom doesn't implement SVG layout, which billboard.js relies on for text/tick sizing.
+window.SVGElement.prototype.getBBox = ( ) => ( {
+  x: 0, y: 0, width: 0, height: 0, toJSON( ) { return this; }
+} );
+window.SVGElement.prototype.getComputedTextLength = ( ) => 0;
+// jsdom's getBoundingClientRect doesn't implement DOMRect's toJSON either.
+window.Element.prototype.getBoundingClientRect = ( ) => ( {
+  x: 0, y: 0, top: 0, left: 0, right: 0, bottom: 0, width: 0, height: 0, toJSON( ) { return this; }
+} );
+// jsdom doesn't implement SVGAnimatedLength, which d3-zoom's defaultExtent() needs
+// to compute a zoomable SVG's extent.
+Object.defineProperty( window.SVGSVGElement.prototype, "width", {
+  configurable: true,
+  get( ) { return { baseVal: { value: 0 } }; }
+} );
+Object.defineProperty( window.SVGSVGElement.prototype, "height", {
+  configurable: true,
+  get( ) { return { baseVal: { value: 0 } }; }
+} );
+
+const seasonalityKeys = [0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11];
+const seasonalityColumns = [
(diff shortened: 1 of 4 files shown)


pleary:
This version of billboard breaks our taxon page taxon page charts, e.g. the seasonality chart. See https://github.com/inaturalist/inaturalist/pull/4779. Feel free to further investigate necessary changes needed to make this work. If you aren't up for that, we can close this PR and leave discussion about updating billboard to the Dependabot PR

anupamme:
Thanks for flagging this and linking #4779 — I dug into it rather than just closing the PR.

## What I tried

I couldn't find a way to reproduce a taxon-page chart crash with billboard.js 3.18.0 against the current `charts.jsx`, across three independent, increasingly faithful checks:

1. **Jest, aligned to production module resolution.** Jest's default CJS resolution actually lands on billboard.js's `dist/billboard.pkgd.js` build, which has a real packaging bug — it doesn't re-export `areaSpline`/`spline`/`zoom` at the top level, so a naive Jest test throws `spline is not a function`. But that's not what our webpack build uses in the browser (no custom `resolve` overrides in `config/webpack.config.js`, so it resolves via the "import" condition → `dist-esm/billboard.js`, which exports everything correctly — confirmed via a clean `webpack` build with no missing-export warnings). Once I pointed Jest at the same ESM build and stubbed the SVG APIs jsdom is missing (`getBBox`, `getBoundingClientRect().toJSON`, `SVGSVGElement.width/height` — all real, known jsdom gaps, unrelated to billboard's version), the seasonality chart, history chart (including the ≥10-year decade-zoom path), field-value charts, and a real tooltip-hover call (exercising the new sanitizer on our custom `tooltip.contents` HTML) all rendered cleanly. Added this as `app/webpack/taxa/show/components/charts.test.jsx` — pushed in `d06804d28`, along with the two `jest.config.js` fixes needed to get this component under test at all.
2. **Real Chromium, single bundle.** Same fixture data/interactions, no jsdom stubs needed — same result, no errors.
3. **Real Chromium, exact production chunk topology** — specifically to test the "conflict with... how we are including [D3] on taxon pages" theory. I rebuilt the actual webpack config with an extra entry sharing `taxa-show`'s exact `dependOn: ["react-main", "d3", "user-text"]`, then loaded `runtime-webpack.js` → `react-main-webpack.js` → `d3-webpack.js` → `user-text-webpack.js` → the entry, in the same order `application.html.erb` + `taxa/show.html.haml` use. Still no crash.

## What I couldn't rule out

My fixtures use a simplified `I18n.t` stub rather than the real translation data, and synthetic (not live-API) observation histograms. If there's something specific to real i18n arrays or real API response shapes, I wouldn't have caught it this way.

Given I can't reproduce it, could you share the exact error/stack trace (and browser) from when you tested this on #4779? That would let us target the actual fix instead of guessing further. Happy to also try standing up a full local Rails env against real data if that's a better use of time than trading guesses.

PR #5037, opened 2026-09-01
feat: Make People index and leaderboard layouts responsive, WEB-1209

sundisc:
Adds responsiveness in the same fashion as the dashboard. Most additions are fairly duplicative of existing non-responsive versions but with changes applied for responsiveness (i.e. I haven't rewritten much erb / haml).

Also updates the test group banner to be displayed on pages that don't use Bootstrap (the People page uses Blueprint).


app/views/users/index.html+responsive.erb
```diff
@@ -0,0 +1,177 @@
...
+            <% end %>
+            <abbr class="timeago" title="<%=t :n_ago, :n => time_ago_in_words(update.created_at.iso8601) %>"><%=t :n_ago, :n => time_ago_in_words(update.created_at) %></abbr>
+          </div>
+        </div>
+      <% end %>
+    </div>
+  <% end -%>
+</div>
+
+<% cache @leaderboard_key, :expires_in => 1.day do %>
```
pleary, line 108: This is going to fragment cache the generated HTML with the key `@leaderboard_key`. We'll want to make sure we use different keys for responsive views compared to non-responsive. Otherwise we'll be showing the HTML for whatever happened to be cached first, potentially showing responsive-related HTML to non-responsive viewers and vice versa. We'll want to ensure the proper key is set early in the controllers before they perform additional logic if the cached fragments do not exist

app/views/users/index.html+responsive.erb
```diff
@@ -0,0 +1,177 @@
...
+        :sub => Proc.new {|user, count| link_to( t(:x_identifications_html, count: count ), identifications_by_login_path( user.login, on: Time.now.strftime( "%Y" ) ) )} %>
+    </div>
+    <div class="leaderboard-readmore">
+      <%- prev_date = Date.today - 1.year -%>
+      <%= link_to t(:view_leaderboard_from_last_year), leaderboard_for_people_path(:year => prev_date.year) %>
+    </div>
+  </div>
+<% end -%>
+
+<% cache @curators_key, :expires_in => 1.day do %>
```
pleary, line 151: Another cache key to consider

app/views/users/leaderboard.html+responsive.haml
```diff
@@ -0,0 +1,45 @@
+= content_for :title do
+  = @title = t(:leaderboard_for_x, :x => (@time_unit == 'month' ? l(@date, :format => :month_year) : @year))
+= content_for :extracss do
+  = stylesheet_link_tag 'users_responsive'
+- cache @leaderboard_key, :expires_in => 1.day do
```
pleary, line 5: Another cache key to consider

app/views/users/_leaderboard_column.html+responsive.haml
```diff
@@ -0,0 +1,32 @@
...
+                  = sub.call(user, count)
+                - else
+                  - if noun == 'species'
+                    - url = taxa_observations_path(:user_id => user.login, :on => time_unit == 'month' ? date.strftime("%Y-%m") : date.year, :hrank => "species")
+                  - else
+                    - url = observations_path( user_id: user.login, on: time_unit == "month" ? date.strftime( "%Y-%m" ) : date.year )
+                  = link_to url do
+                    =t "x_#{noun.camelize.parameterize.pluralize}_html", :count => count.to_i
+    - else
+      .empty=t :no_results_found
```
pleary, line 30: Very trivial and perhaps intentional - this text on the non-responsive view is grey and center aligned, vs black and left justified in the responsive view

app/views/users/index.html+responsive.erb
```diff
@@ -0,0 +1,177 @@
...
+      <%= text_field_tag :q, @q, :placeholder => t(:search_by_name_or_email) %>
+      <%= submit_tag t(:search) %>
+    <% end %>
+  <% end -%>
+</div>
+
+<div id="recent">
+  <div class="section-header">
+    <h3><%= t :recently_active %></h3>
+    <div class="updated">
```
pleary, line 28: Another small and maybe intentional change - these updated labels on a wider display get very small - an effective 9.6px vs 14.4px. A lot of the text on this page is already pretty small compared to other pages, but these updated labels in the responsive view are particularly tiny

app/views/users/index.html+responsive.erb
```diff
@@ -0,0 +1,177 @@
...
+      <%= render 'leaderboard_column', :noun => 'species', :noun_plural => 'species', :time_unit => 'month', :data => @most_species %>
+      <%= render 'leaderboard_column',
+        :noun => 'identification',
+        :noun_plural => 'identifications',
+        :time_unit => 'month',
+        :data => @most_identifications,
+        :sub => Proc.new {|user, count| link_to( t(:x_identifications_html, count: count ), identifications_by_login_path( user.login, on: Time.now.strftime( "%Y-%m" ) ) )} %>
+    </div>
+    <div class="leaderboard-readmore">
+      <div class="leaderboard-note"><%=raw t :observations_observed_this_month %></div>
```
pleary, line 129: Another trivial one - text used to be grey and now it's black. Again - OK with me if this is intentional, I'm just calling it out since I noticed it while reviewing

app/views/users/index.html+responsive.erb
```diff
@@ -0,0 +1,177 @@
...
+<% cache @curators_key, :expires_in => 1.day do %>
+  <div id="curators">
+    <h3><%= t :noble_curators, :site => @site.site_name_short %></h3>
+    <p><%= t :these_folks_kindly_volunteer_to %></p>
+    <% if !@curators.blank? %>
+      <div class="CardGrid">
+        <% @curators.each do |user| %>
+          <div class="UserWithIcon">
+            <div class="icon">
+              <%= link_to_user( user ) do %><%= user_image user, size: "thumb", no_inline_dimensions: true %><% end %>
```
pleary, line 160: Another smaller one - username links used to be blue, now they are black

pleary (changes requested):
I'm also seeing that on the Leaderboard pages, the header section styles seem off on wider views - the header text is off to the right of the page until very narrow views:

Image


If I haven't opted into the responsive test, and I dismiss the banner, there is some extra top padding on the banner-dismissed alert. It is not present when dismissing the banner when already opted-in:

Image


Overall this is looking great. The fragment caching is probably the only release blocker, but would be great to address these two UI issues above as they seem the most obviously unintentional

app/views/users/index.html+responsive.erb
sundisc: Ahh yes, thanks for catching this. Added a cache suffix based on the variant on the key before the additional logic happens.

app/views/users/_leaderboard_column.html+responsive.haml
sundisc: Not intentional. Fixed, thanks!

app/views/users/index.html+responsive.erb
sundisc: Ah true, fixed it at 14px.

app/views/users/index.html+responsive.erb
sundisc: I glazed over this, thanks! Fixed it.

sundisc:
@pleary Implemented changes for the caching and styling to match prod better.

pleary (approved):
Thanks for those fixes. This is looking good to me

PR #5047, opened 2026-09-03
web-1266: add carto key param

caseyhofford:
Adds an API key for carto due to a change in their requirements for free tiles. Key in 1password engineering vault as "Carto API Key".

### DIFF
--- a/app/views/layouts/bootstrap.html.erb
+++ b/app/views/layouts/bootstrap.html.erb
@@ -215,7 +215,8 @@
       };
       var CONTROLLER_ACTION = "<%= controller.class.to_s %>::<%= action_name %>";
       var CONFIG = {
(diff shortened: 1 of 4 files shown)


sundisc (approved):

pleary (approved):
Looks good. Could you also [add it here](https://github.com/inaturalist/inaturalist/blob/main/app/webpack/stats/year/components/global_map.jsx#L28)?

caseyhofford:
> Looks good. Could you also [add it here](https://github.com/inaturalist/inaturalist/blob/main/app/webpack/stats/year/components/global_map.jsx#L28)?

Good catch @pleary; added in 062e28bf2e3f82a27209044b9a4aa67236bccb6e.

PR #5050, opened 2026-09-03
Flipper MVP

caseyhofford:
Adds FeatureFlagging using Flipper gem.

To test, add a feature flag `client_demo_banner` at `/admin/feature_flags/features`.

Add an actor for your logged in user to the created feature (e.g. `User;<USER_ID>`).

Load `/home` and confirm you see the `Feature flag demo` pill in the footer.

I've created a flipper user guide in a linear document here https://linear.app/inaturalist/document/feature-flags-creating-and-managing-flags-eba43086a147.


app/controllers/application_controller.rb
```diff
@@ -931,6 +931,7 @@ def allow_cors
   def append_info_to_payload(payload)
     super
     payload.merge!(Logstasher.payload_from_request( request ))
+    payload.merge!( FeatureFlagging::Telemetry.payload )
```
pleary: Before deploy any new attributes to the logstash payload, we need to coordinate with Ops to make sure the new attributes are already defined in the receiving DB with proper types. I'm not sure we want to be sending these 100% of the time, especially since initially there won't be any requests using Flipper. I think it would be great to be able to isolate which requests actually might involve Flipper and only include these additional payload fields for those requests

app/views/layouts/admin.html.haml
```diff
@@ -40,6 +40,8 @@
           %li= link_to "Redirect Links", redirect_links_path
           %li= link_to "Blocked IPs", admin_blocked_ips_path
           %li= link_to t( :username_reserved_words ), admin_username_reserved_words_path
+          - if current_user&.is_admin?
```
pleary, line 43: This template is only rendered for admins, and we'll have already confirmed that in the controller. This is redundant and can be removed

app/views/layouts/application.html.erb
```diff
@@ -63,7 +63,8 @@
       <% end -%>
       var CONTROLLER_ACTION = "<%= controller.class.to_s %>::<%= action_name %>";
       var CONFIG = {
-        content_freeze_enabled: <%= !!CONFIG.content_freeze_enabled %>
+        content_freeze_enabled: <%= !!CONFIG.content_freeze_enabled %>,
+        feature_flags: <%= feature_flags_json %>
```
pleary: I don't think we should be including these in the layouts. This will trigger DB queries for Flipper on every request when initially nothing will actually be using it, and eventually I suspect we'll be using it in a targeted way. Ideally only the views that actually need these flags query for and insert them

app/views/layouts/basic.html.haml
```diff
@@ -42,7 +42,8 @@
       ).html_safe };                                    |
       var CONTROLLER_ACTION = "#{controller.class.to_s}::#{action_name}";
       var CONFIG = { |
-        content_freeze_enabled: #{ !!CONFIG.content_freeze_enabled } |
+        content_freeze_enabled: #{ !!CONFIG.content_freeze_enabled }, |
+        feature_flags: #{ feature_flags_json } |
```
pleary: Again, my preference would be to remove these from layouts

app/views/layouts/bootstrap.html.erb
```diff
@@ -216,7 +216,8 @@
       var CONTROLLER_ACTION = "<%= controller.class.to_s %>::<%= action_name %>";
       var CONFIG = {
         content_freeze_enabled: <%= !!CONFIG.content_freeze_enabled %>,
-        carto_api_key: <%= CONFIG.carto&.api_key.to_json.html_safe %>
+        carto_api_key: <%= CONFIG.carto&.api_key.to_json.html_safe %>,
+        feature_flags: <%= feature_flags_json %>
```
pleary: Again, my preference would be to remove these from layouts

app/views/shared/_footer.html.haml
```diff
@@ -67,6 +67,13 @@
             &nbsp;
           .footer-link
             =t :documentation_for_developers_html
+          - if FeatureFlagging.enabled?( :client_demo_banner, current_user )
```
pleary: It is useful to see how we implement Flipper flagging into a view, but since this is just a "demo" and we don't intend to actually use it, I think this should be removed. This also causes Flipper to be queried for every single page load, something I'd like to avoid. Maybe this can be added to documentation, but left out of the codebase?

app/webpack/observations/show/components/app_legacy.jsx
```diff
@@ -47,6 +47,7 @@ import TagsContainer from "../containers/tags_container";
 import ModeratorActionModalContainer from "../containers/moderator_action_modal_container";
 import ObservationModalContainer from "../containers/observation_modal_container";
 import RtlTestGroupToggle from "../../../shared/components/rtl_test_group_toggle";
+import FeatureFlagDemoBanner from "../../../shared/components/feature_flag_demo_banner";
```
pleary: Similarly here, this React implementation is really helpful to know about, but since we don't intent to actually use it, I don't think it should be committed to the codebase

app/webpack/shared/feature_flags.js
```diff
@@ -0,0 +1,9 @@
+// Read CONFIG on every call, not at import, to avoid bundle ordering issues.
+export default function featureFlagEnabled( key ) {
```
pleary, line 2: We have a shared [CurrentUser model](https://github.com/inaturalist/inaturalist/blob/492c4ff11622b4b079c1586a488e8c57bfa6ebc0/app/webpack/shared/models/current_user.js), created and managed through the [config reducer](https://github.com/inaturalist/inaturalist/blob/492c4ff11622b4b079c1586a488e8c57bfa6ebc0/app/webpack/shared/ducks/config.js), and that seems to me like a great place to incorporate Flipper feature flagging. We could have a method like `config?.currentUser?.featureFlagEnabled( flag )`. Not a big deal, but could be helpful to consolidate some of the same types of methods, e.g. can the current user do this thing

config/initializers/flipper.rb
```diff
@@ -0,0 +1,52 @@
...
+  config.adapter { FeatureFlagging.build_adapter }
+end
+
+flipper_config = Rails.application.config.flipper
+
+# One gate read per flag per request instead of one per check.
+flipper_config.memoize = true
+
+# One get_all per request (safe via FailClosedAdapter); the feature set it loads is also the flag registry.
+flipper_config.preload = true
```
pleary: My preference would be to set this to false. Given we won't have any use of Flipper at all at launch, preloading will just cause a lot of queries to be run that we'll never use. Even at launch, I'd suggest we might want to leave this off forever - it depends on how many requests the features actually affect and how many feature flags we have

app/views/admin/index.html.erb
```diff
@@ -1 +1,41 @@
 <h2><%=t :admin_tools %></h2>
+
+<% feature_keys = FeatureFlagging.feature_keys %>
```
pleary, line 3: This basically changes this view from be a jumping off point to anything, to being all about feature flags. I think my preference would have been to put this template into a separate action like all the other admin tools, e.g. have this be what renders at `/admin/feature_flags`, leaving `/admin/feature_flags/features` for the more detailed view, and leaving `/admin` to be as it is now a generic jumping off point to any admin tool

pleary (changes requested):
I left a few comments and suggestions for now

app/views/layouts/application.html.erb
caseyhofford: Good call, I agree it makes sense to be way more targeted with querying for flags. Fixed in f08b852f47c11ea76fcf91c06d624ece94c962f2

app/views/layouts/basic.html.haml
caseyhofford: Fixed f08b852f47c11ea76fcf91c06d624ece94c962f2

app/views/layouts/bootstrap.html.erb
caseyhofford: Fixed f08b852f47c11ea76fcf91c06d624ece94c962f2

app/views/shared/_footer.html.haml
caseyhofford: Yeah, that's a good call, how to use the flags is already documented in linear, which is a better place for that then in the code. https://linear.app/inaturalist/document/feature-flags-creating-and-managing-flags-eba43086a147#heading-for-developers-9dd3b519

d8acabb6d5bb460761ad5885b4a8cff6a14f68eb

app/webpack/observations/show/components/app_legacy.jsx
caseyhofford: Agreed! d8acabb6d5bb460761ad5885b4a8cff6a14f68eb

config/initializers/flipper.rb
caseyhofford: Yeah, I think that makes sense, because we will probably always have more requests that do not need FF than do. The default is enabling preloading, which flipper recommends but that assumes few feature flags which are needed on most requests, in the future we can define more complex rules for when preloading should happen: (https://www.flippercloud.io/docs/optimization#preloading).  Disabled in 7385903cac12338d58477a53e9b7e4cb1e026d1b

app/controllers/application_controller.rb
caseyhofford: Good to know, and agreed that we don't need flags on every request. Maybe we can move telemetry into `FeatureFlagging#flags_for` so we get tracking only when flags are evaluated. Removed in 89bac60e3c8261e5e19a35d07f402a23a8f2eef7

caseyhofford:
I didn’t quite get everything working before PTO so moving this back to draft. No need to merge until after I’m back @pleary .

PR #5051, opened 2026-09-04
fix: Prevent header from overflowing, WEB-1261

sundisc:
Adds more to keep the header from overflowing on narrow screens (adds a crowded class at runtime, responsive nav/leaderboard breakpoints).

Also migrates the header JS from application.js.erb into the typed shared/header webpack module, which is more convenient for linting, testing, editor highlighting, webpack processes like tree-shaking, avoiding globals as much as possible, etc...

Initial updates/messages counts now flow through data-* attributes on #header (read by initHeaderCounts) instead of a per-layout inline <script>, collapsing the duplicated init in both layouts into the shared header partial. The window global is trimmed to just setUpdatesCount/setMessagesCount (still called by the legacy dropdown handlers, so we can remove if we migrate those eventually). getUpdatesCount/getMessagesCount removed.

In the future, it would be nice to continue to refactor application.js.erb into TS.

Additionally simplified some data structure around breakpoints for testing and broke out the overflow assertion from the `expectNoHorizontalOverflow` test so that it can be used separately.

### DIFF
--- a/app/assets/javascripts/application.js.erb
+++ b/app/assets/javascripts/application.js.erb
@@ -457,6 +457,8 @@ $(document).ready(function() {
   $('.dna').dnaHighlight()
 
   if ( $( "#header" ).length > 0 ) {
+    fitHeader( );
+    $( window ).on( "resize", _.debounce( fitHeader, 100 ) );
     $( "#header .search input" ).universalAutocomplete( );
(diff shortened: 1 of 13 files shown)


caseyhofford (approved):
Looks good!

pleary (approved):

PR #5052, opened 2026-09-04
fix: Use "Show More" button with style that hides text when viewport is small, WEB-1252

sundisc:
Also:
- add padding between the last row of always-shown observations and the first row of hidden observations in multi-observation dashboard items.
- use translated strings

Before:
Screenshot 2026-09-04 at 5 01 36 PM

After:
Screenshot 2026-09-04 at 5 01 59 PM


app/views/users/_update_created_post.html.haml
```diff
@@ -14,7 +14,8 @@
     %button{:type => "button", :class => 'btn btn-sm btn-default btn_expand collapsed pull-right', "data-toggle" => "collapse", "data-target" => "#expand#{target_id}"}
       %span.expand_inactive
         %i.glyphicon.glyphicon-collapse-down
-        Show More...
+        = succeed "..." do
+          = t(:show_more)
```
caseyhofford, line 18: Do we want to add `class="btn_expand_label"` to these strings as well?

caseyhofford (approved):
Looks good, one question about adding the `btn_expand_label` class in one more place.

pleary (approved):
Looks good. Maybe this is feature creep for a targeted fix, but seems like there's a few more dashboard partials that would benefit from the same fix of wrapping the label with span.btn_expand_label, e.g.:
* https://github.com/inaturalist/inaturalist/blob/main/app/views/comments/_listing_for_dashboard.html%2Bresponsive.erb#L104
* https://github.com/inaturalist/inaturalist/blob/main/app/views/comments/_listing_for_dashboard.html.erb#L107 (maybe not if we don't want to touch the not explicitly responsive partials)
* https://github.com/inaturalist/inaturalist/blob/main/app/views/projects/_update_created_post.html.erb#L22 (looks like the label was moved and changed to use i18n, but without a span.btn_expand_label)
* https://github.com/inaturalist/inaturalist/blob/main/app/views/taxa/_update_new_observations.html.erb#L26
* https://github.com/inaturalist/inaturalist/blob/main/app/views/users/_dashboard_updates.html%2Bresponsive.haml#L54
* https://github.com/inaturalist/inaturalist/blob/main/app/views/users/_dashboard_updates.html.haml#L64 (maybe not if we don't want to touch the not explicitly responsive partials)
* https://github.com/inaturalist/inaturalist/blob/main/app/views/users/_update_created_post.html.haml#L16 (mentioned by @caseyhofford)

sundisc:
@pleary @caseyhofford Thanks for the reviews! I considered adding the style to these as well, but since they only have two buttons, the button never gets squeezed enough for the label to overflow.

pleary:
OK sounds good. It would need to take a combination of a narrow device plus a language that uses a lot of characters for the buttons, and if we don't have a combo like that then this all sounds good to me

sundisc:
> OK sounds good. It would need to take a combination of a narrow device plus a language that uses a lot of characters for the buttons, and if we don't have a combo like that then this all sounds good to me

Oh @pleary you're totally right. I did a search for the longest language strings in this case and some do overflow with just two buttons. I've added the button shrinking to those that you suggested except for the legacy ones. Also implemented it in a way that preserves the text for screen-readers.

app/views/users/_dashboard_updates.html+responsive.haml
```diff
@@ -51,11 +51,12 @@
                 %button.btn.btn-sm.btn-default.btn_expand.collapsed{ "aria-pressed" => "false", "data-target" => ".expand#{target_id}", "data-toggle" => "collapse", type: "button" }
```
pleary: This is really minor, but I noticed that on some dashboard updates the arrow wasn't changing from a down to up arrow after expanding the section. It seems like this is an existing bug and the problem is related to using a class selector here instead of an ID selector. I think we might want this to be `"#expand#{target_id}"` and update lines 41 and 45 here to use `:id => "expand#{target_id}"` instead of `:class => "expand#{target_id}"`

pleary (approved):
This is all looking good to me. I left one other minor comment. OK with me to merge this whenever

app/views/users/_dashboard_updates.html+responsive.haml
sundisc: Fixed https://github.com/inaturalist/inaturalist/pull/5052/commits/376cd627577a867bb369128f7163f18f96f43caf

PR #5055, opened 2026-09-09
fix: Prevent autocomplete overflows, WEB-1265

sundisc:
- Prevent autozoom on iPhone, which was causing the autocomplete to render offscreen.
IMG_1788

- Suggest an Identification -> Identify to fit small screens
Screenshot 2026-09-09 at 11 40 05 AM

- Fix observation fields autocomplete overflowing screen when really long fields are present
Screenshot 2026-09-09 at 11 40 34 AM

### DIFF
--- a/app/assets/stylesheets/_breakpoints.scss
+++ b/app/assets/stylesheets/_breakpoints.scss
@@ -1,12 +1,14 @@
 $mobile-max:  767px;
 $desktop-min: 768px;
 
+$xs: 420px;
 $sm:  576px;
 $md:  768px;
 $lg:  992px;
 $xl:  1200px;
 $xxl: 1400px;
 
+$xs-max:  419.98px;
 $sm-max:  575.98px;
 $md-max:  767.98px;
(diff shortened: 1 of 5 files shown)


caseyhofford (approved):
Looks good!

pleary (approved):
Pretty minor, but when testing from an iPhone or using the browser to simulate one, the `Species name` text input is slightly taller than the input group addon. Seems like it comes from the `height: auto` on `.ac-chooser input.form-control`. Other than that, this looks good - not seeing zooming and for taxon and observation field search I'm not seeing overflows

Image

sundisc:
> Pretty minor, but when testing from an iPhone or using the browser to simulate one, the `Species name` text input is slightly taller than the input group addon. Seems like it comes from the `height: auto` on `.ac-chooser input.form-control`. Other than that, this looks good - not seeing zooming and for taxon and observation field search I'm not seeing overflows
> Image

Ah yeah, thanks @pleary. Made the thumb also `height: auto` to match.

PR #5056, opened 2026-09-09
fix: Make project observation fields modal responsive, WEB-1263

sundisc:
Before:
Screenshot 2026-09-09 at 2 12 21 PM

After:
Screenshot 2026-09-09 at 2 12 10 PM

### DIFF
--- a/app/assets/stylesheets/shared/observation_projects.scss
+++ b/app/assets/stylesheets/shared/observation_projects.scss
@@ -1,4 +1,5 @@
 @import "../colors";
+@import "../breakpoints";
 
 .Projects {
   .project,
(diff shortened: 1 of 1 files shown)


caseyhofford (approved):
Looks good! Tested locally on mobile screen widths and desktop widths.

pleary (approved):
The responsive styles look good to me. When I tested this on my iPhone directly, when clicking on the project autocomplete field there was a slight autozoom. That caused the right side of the modal to be obscured. But zooming out slightly showed the whole thing styled correctly for the narrow screen. I did not see the autozoom when simulating the same iPhone in a browser

Image

PR #5057, opened 2026-09-09
chore: Allow mDNS hostname in development config

sundisc:
Allows us to connect easily on LAN.


caseyhofford (approved):
Nice! Tested this and it worked to allow me to connect to the rails server from other local devices using my machine name rather than my IP. Thanks for getting that working!

pleary (approved):

PR #5060, opened 2026-09-10
fix: Prevent dashboard taxon link from overlapping agree button, WEB-1257

sundisc:
Before:
Screenshot 2026-09-10 at 12 04 57 PM

After:
Screenshot 2026-09-10 at 12 05 07 PM

### DIFF
--- a/app/assets/stylesheets/dashboard.scss
+++ b/app/assets/stylesheets/dashboard.scss
@@ -663,6 +663,7 @@ h3.dashboard_id{
 h3.dashboard_id > .taxon > .comname,
 h3.dashboard_id > .taxon > a > .sciname {
   display: block;
+  width: fit-content;
 }
 
 .glyphicon-cog {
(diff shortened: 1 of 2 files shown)


pleary (commented):
I can confirm this is causing the name span width to restrict to the name text, but the `a` link still appears to span the entire H3. It seems to work OK on my iPhone clicking on the Agree button, but in a browser it's possible to have the taxon link override the Agree button link. It depends on how the cursor enters the Agree button. From the left or top, it uses the name link, and from the right or bottom it uses the Agree button link. Can you replicate that? Should a similar fix (display: block, width: fit-content) be applied to the `a` link as well?

https://github.com/user-attachments/assets/867f094a-0c07-4aac-af96-9d44124262d5

sundisc:
Ah good catch @pleary .. sorry about that, my testing was a little short-sighted. These all can be fixed by modernizing to `flex` so I added an override to the responsive version.

pleary (approved):
The flex fix appears to be working for me

PR #5063, opened 2026-09-11
fix: remove long transaction from BulkObservationFile, OPS-424

pleary:
Removes a transaction wrapping the entire processing of a BulkObservationFile. This was leading to table locking issues related to delayed jobs since these transactions could potentially remain open without committing for hours

### DIFF
--- a/lib/bulk_observation_file.rb
+++ b/lib/bulk_observation_file.rb
@@ -117,43 +117,41 @@ class BulkObservationFile
   def import_file
     row_count = 1
     observations = []
-    ActiveRecord::Base.transaction do
-      CSV.foreach( @observation_file, encoding: "iso-8859-1:utf-8", headers: true ) do | row |
-        next if skip_row?( row )
+    CSV.foreach( @observation_file, encoding: "iso-8859-1:utf-8", headers: true ) do | row |
+      next if skip_row?( row )
 
-        # Add the observation file name as a tag for identification purposes.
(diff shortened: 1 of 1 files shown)


lib/bulk_observation_file.rb
```diff
@@ -117,44 +117,42 @@ def validate_file
...
+        observations << obs
 
-        # Manually update counter caches.
-        ProjectUser.update_observations_counter_cache_from_project_and_user( project.id, user.id )
-        ProjectUser.update_taxa_counter_cache_from_project_and_user( project.id, user.id )
-        Project.update_observed_taxa_count( project.id )
+        # Increment the row count so we can tell them where any errors are.
+        row_count += 1
+      rescue ActiveRecord::RecordInvalid
+        raise BulkObservationException.new( "Invalid record encountered", row_count )
```
caseyhofford, line 141: Should we log any more details from the failed row here, such as the filename to make it easier to confidently identify the row that failed.

caseyhofford (approved):
Looks good! One comment about adding more details to the exception.

lib/bulk_observation_file.rb
pleary: The only change I made to this file is removing a transaction, which changed the whitespace making it look like I touched these other lines. To me, changing anything else is out of scope of this issue so I'll leave this for now

PR #5064, opened 2026-09-11
feat: Add pagination and scroll to top on dashboard, WEB-1258 + WEB-1254

sundisc:
- Scrolls to top after pressing the *new* Next / Back buttons
  - This change bleeds over to the non-responsive page as well, but it seems like a good all-around change
- Adds scroll to top button under `$md` breakpoint
- Adds a more paginated way of navigating with better accessibility
- Adds `aria-live` and `aria-busy` to indicate loading for screen reader users
- Adds Playwright tests for `dashboard.js` coverage
- Removes unused stuff in `dashboard.js`
- Lints `dashboard.js` and adds it to the CI linting step

https://github.com/user-attachments/assets/88f75a65-47a8-46f1-aed2-632c07518ee2

### DIFF
--- a/app/assets/javascripts/users/dashboard.js
+++ b/app/assets/javascripts/users/dashboard.js
@@ -1,11 +1,12 @@
-/* eslint-disable */
+/* global DASHBOARD_FROM, DASHBOARD_TAB, I18n, updateSession */
 
 var DASHBOARD = {
-  fromIDs: { }
+  fromIDs: { },
+  fromPage: { }
 };
 
 // hide the flash message after 5 seconds
-setTimeout( function( ) {
+setTimeout( function ( ) {
   $( "#flash" ).fadeOut( 1000 );
 }, 5000 );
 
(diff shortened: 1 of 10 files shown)


config/locales/en.yml
```diff
@@ -2335,6 +2335,7 @@ en:
   find_your_current_location: Find your current location
   finished: Finished
   finished_with_page: Finished With Page
+  first: First
```
pleary, line 2339: Can we add translation context for these 3 new translations?

app/views/users/_dashboard_updates.html+responsive.haml
```diff
@@ -61,6 +61,12 @@
...
-  .pagination
-    - unless @pagination_updates.blank? || @pagination_updates.next_page.blank?
-      = link_to t(:more), url_for_params( from: @pagination_updates.last.created_at.to_i ), data: { from: @pagination_updates.last.created_at.to_i }, id:"more_pagination", class: "btn btn-sm btn-default"
+  - if @pagination_updates.respond_to?( :previous_page )
+    - previous_page = @pagination_updates.previous_page
+    - next_page = @pagination_updates.next_page if @pagination_updates.current_page < UsersController::DASHBOARD_MAX_PAGE
+    %nav.dashboard-pagination{ "aria-label" => t(:pagination) }
+      - { t(:first) => ( 1 if previous_page ), t(:back) => previous_page, t(:next) => next_page }.each do | label, page |
+        - if page
+          = link_to label, url_for_params( page: page ), data: { page: page }, class: "btn btn-sm btn-default page_button"
```
pleary, line 70: The change to using page here make sense if we want to make easy page links. I'll point out it could change behavior a little in that using the date previously meant hitting "More" would load all new content created after the last item displayed. If we just increment page, it's possible new content was added while looking at a page, and the items at the top of the "Next" page will be the same items at the bottom of the previous page. Classic edge-case pagination problem and one we know we have elsewhere on the site.

app/assets/javascripts/users/dashboard.js
```diff
@@ -62,133 +70,149 @@ DASHBOARD.tabSettings = function( type ) {
...
+    window.history.pushState( state, "", "/home?" + $.param( dashboardParams ) );
   }
 };
 
-DASHBOARD.loadingPanel = function( selector ) {
-  $( "body" ).scrollTop( 0 );
-  $( selector ).html( "<div class='loading status'>" + I18n.t( "loading" ) + "</div>" );
+DASHBOARD.startPanelLoading = function ( selector ) {
+  var target = $( selector );
+  window.scrollTo( 0, 0 );
```
pleary: Might we want this to go to the top of the dashboard updates and not the entire page, e.g. `$( window ).scrollTop( $( ".dashboard_tab_row" ).offset( ).top );`? I think the added scroll-to-top button should go to  the top of the page. But for dashboard updates it's possible there's a lot of stuff at the top of the page (e.g. announcements), and really we just want to jump up to the top of the updates list

pleary (changes requested):
Looking good to me. I'm only requesting the updated context for translations and the rest is OK with me to merge, though I left another comment about the scrolling behavior we might want to consider

app/views/users/_dashboard_updates.html+responsive.haml
sundisc: Ah yeah, I suppose probably not that big of a deal in the short run as a user navigates through their updates, but I didn't think about the case where a user may want to save the link to a page of updates for later. Do you think that's a common thing? I sort of doubt it, but could be some possible feedback we should look out for.

app/assets/javascripts/users/dashboard.js
sundisc: If we jump to the top of the updates list it seems a little hard to get a grasp of whether you're at the top of the page or in the middle of the updates somewhere, since only updates are visible on the screen. That's why I was scrolling all the way to the top, but I wasn't thinking about the announcements; good call. Instead, how about we scroll to the top of the user header? Added here: https://github.com/inaturalist/inaturalist/pull/5064/commits/0c922f2bc3dc7c5f3e33f29d4c271f3f4a6b2928

Screenshot 2026-09-17 at 11 58 45 AM

config/locales/en.yml
sundisc: Definitely: https://github.com/inaturalist/inaturalist/pull/5064/commits/acd80fb28c45464a08afca452ceebc4407f6785a

app/views/users/_dashboard_updates.html+responsive.haml
pleary: I wouldn't think bookmarking dashboard pages is common, but I'm not sure. I think we're more likely (though also probably quite rarely) to seem people report seeing the same updates on multiple pages during say a high contribution event. But I don't think that warrants any changes here given how much of an edge case that will likely be

pleary (approved):

PR #5067, opened 2026-09-15
feat: add segment codes to some donation links, WEB-1237

pleary:
Add segment codes to several donation links. Also updates any absolute URLs to be relative, and removing explicit values of `utm_source` to a fixed domain as the donate page will populate that as necessary with a redirect


sundisc (approved):
Looks good!

PR #5069, opened 2026-09-15
chore: upgrade acts-as-taggable-on to 10.0.0, WEB-1125

pleary:
Updates `acts-as-taggable-on` to 10.0.0 which is the latest version before Rails 6.1. compatibility is dropped. This version does not support Rails 7.2. We'll need to upgrade to 11.0.0 to get Rails 7.2 support, at which time we'll lose 6.1 support


sundisc (approved):
Looks good! Think we need to create a ticket to remind ourselves to upgrade to 11 once we're on Rails 7?

pleary:
> Looks good! Think we need to create a ticket to remind ourselves to upgrade to 11 once we're on Rails 7?

Thanks - I made an issue for the 7.1 phase

PR #5071, opened 2026-09-16
feat: Improve autocomplete functionality for mobile users, WEB-1262

sundisc:
Update so that ([on mobile only!](https://github.com/inaturalist/inaturalist/pull/5071/changes#diff-c821f7c32b1fefc62837a1d0df415d9f816f47bf74f545239dcc634783bcaa81R164)) the suggestions stay open when leaving the focus of the "Suggest an identification" taxon autocomplete. It's not enabled for desktop because we do want the suggestion list to disappear when tabbing out of that textbox, e.g. when a user is using keyboard only and tabbing through page inputs.

This is available on staging3 for testing.

### DIFF
--- a/app/webpack/observations/show/components/activity_create_panel.jsx
+++ b/app/webpack/observations/show/components/activity_create_panel.jsx
@@ -163,8 +163,18 @@ class ActivityCreatePanel extends React.Component {
             config={config}
             onKeyDown={e => {
               const key = e.keyCode || e.which;
-              if ( key === 13 ) {
+              if ( key !== 13 ) { return; }
+              const ac = $( ".id_tab input[name='taxon_name']" ).data( "uiAutocomplete" );
+              if ( ac?.selectedItem ) {
                 this.postIdentification( );
+                return;
+              }
+              // return key blurs the mobile keyboard; keep the suggestion list
+              // open and the term in place instead of clearing on blur
+              e.preventDefault( );
+              if ( ac ) {
+                ac.keepOpen = true;
+                setTimeout( ( ) => { ac.keepOpen = false; }, 300 );
               }
             }}
           />


pleary (commented):
I don't think this is working properly for me on my iPhone. Though I'm also not 100% clear how this should work so maybe I'm missing something. If I open the taxon name autocomplete on an observation details page, I can tap outside it to close it. The close event actually seems to fire earlier as it gets closed on touchstart now with the `touchstart.acKeepMenuOpen` event, whereas in production it doesn't fire until touchend. What should the proper behavior be in this PR? When and how should the menu be closed on mobile devices differently from how it works in production?

One minor style suggestion - the `glyphicon-search` icon in `ac-select-thumb` has a `top: 3px` custom style, with with the flex styling added the top value causes the search icon to appear off centered a little too low, so maybe it should be removed. 

One request: could you please add `square_url: true,` to `default_photo` and `representative_photo` in https://github.com/inaturalist/inaturalist/blob/6195f4e5d773b766eecaf523923e511fbeb19849/app/webpack/observations/uploader/components/taxon_autocomplete.jsx#L14? That's a bug that I caused that is resulting in the taxon thumbnails not showing when selecting most search results. It'd be great if you could fix my mistake while touching some similar files.

sundisc:
> I don't think this is working properly for me on my iPhone. Though I'm also not 100% clear how this should work so maybe I'm missing something. If I open the taxon name autocomplete on an observation details page, I can tap outside it to close it. The close event actually seems to fire earlier as it gets closed on touchstart now with the `touchstart.acKeepMenuOpen` event, whereas in production it doesn't fire until touchend. What should the proper behavior be in this PR? When and how should the menu be closed on mobile devices differently from how it works in production?

Ah yeah I could have explained this better. The blur this is solving for is when the user closes the mobile keyboard (on iPhone, they press the checkmark at the top right). This allows the user to get the keyboard out of the way and still be able to select an item from the autocomplete menu.

Tapping outside behavior shouldn't have really changed, though I do see what you mean about it closing on touchstart now as opposed to touchend. I don't think that's a big deal though, if somebody touchstarts the touch will eventually touchend.

So on mobile:
- If the user closes the keyboard, the menu should still be visible
- If the user touches outside of the menu or keyboard, the menu should close
- If the user selects an item, the menu should close
- After the user selects an item, the menu should stay closed

And I just fixed a case where the menu returns when the keyboard is closed _then_ a menu item is selected. https://github.com/inaturalist/inaturalist/pull/5071/commits/2587a80d1fb7beb28fdbf1d4b850f6b28c21d1e5

> One minor style suggestion - the `glyphicon-search` icon in `ac-select-thumb` has a `top: 3px` custom style, with with the flex styling added the top value causes the search icon to appear off centered a little too low, so maybe it should be removed.
> 
> One request: could you please add `square_url: true,` to `default_photo` and `representative_photo` in
> 
> https://github.com/inaturalist/inaturalist/blob/6195f4e5d773b766eecaf523923e511fbeb19849/app/webpack/observations/uploader/components/taxon_autocomplete.jsx#L14
> ? That's a bug that I caused that is resulting in the taxon thumbnails not showing when selecting most search results. It'd be great if you could fix my mistake while touching some similar files.

Sure! Here: https://github.com/inaturalist/inaturalist/pull/5071/commits/469fd485e75d2d62bfc58aaa3152c84ce72e49bf

sundisc:
@pleary FYI I just deployed this to staging4 because I'm using 3 for something else.. should be up in a bit for testing.

pleary (approved):
Thanks for the explanation. Yes this does appear to be working correctly for me. Thanks for making those other changes as well

